PR #6, opened 2017-07-03
fix-symlinks

dradetsky:
from a fresh checkout & sync, a bunch of the links were broken. I don't know if other people are doing weird custom things such that the old links work, but these work with what happens by default.

there are still several more broken links. I didn't fix them because I wasn't totally sure what was correct. 

- `source/layouts/commands-workspace.erb`

I think this is the same thing as `commands-env.erb`. There seems to be some kind of not-entirely-successful renaming effort in progress I don't entirely understand.

- `source/layouts/terraform-enterprise.erb`

not totally sure where this is supposed to go.

- `source/docs/providers/do`

I assume this is short for digital ocean, in which case it's missing from `.gitmodules` too. Even if I'm right, I'm not totally sure how git modules work, except for having a general impression that they're hard to use correctly, so I'm disinclined to just go ahead and add a similar entry to `.gitmodules`.

- `source/docs/providers/terraform-enterprise`

Maybe this is supposed to point to "atlas" instead of "terraform-enterprise." Either way, it's missing from `.gitmodules`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -270,3 +270,11 @@
 	path = ext/providers/vsphere
 	url = https://github.com/terraform-providers/terraform-provider-vsphere
 	branch = stable-website
+[submodule "ext/providers/do"]
+	path = ext/providers/do
+	url = https://github.com/terraform-providers/terraform-provider-digitalocean
+	branch = stable-website
+[submodule "ext/providers/atlas"]
+	path = ext/providers/atlas
+	url = https://github.com/terraform-providers/terraform-provider-atlas
+	branch = stable-website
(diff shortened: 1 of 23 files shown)


grubernaut:
Hey @dradetsky, thanks for this PR! I've just submitted, https://github.com/hashicorp/terraform-website/pull/7, to add the missing submodules that you found. This should clear up the `do` and `terraform-enterprise` link issues you've found above. 

I'm not 100% sure about the `commands-workspace` vs `commands-env.erb` issue, but will clarify internally. 

Thanks again!

dradetsky:
@grubernaut cool, thanks.

PR #12, opened 2017-07-31
Add some assets for dev provider program

catsby:
Will add link to PR when I can. The repo is kind of split brain at the moment, with assets here but content in hashicorp/terraform 🤔


catsby:
https://github.com/hashicorp/terraform/pull/15677

PR #14, opened 2017-08-04
Give bucket owner full control when uploading to S3

radeksimko:
This is (AFAIK) necessary for cross-account uploads in this context.


catsby:
Hey @jvoorhis can you weigh in on if this is still necessary?

catsby:
Or @radeksimko , do you know if this is still necessary? Website deploys seem to be working 🤔

radeksimko:
I don't mind closing this one - as long as we keep using credentials which belong to the same account as bucket owner - not sure if that's the long-term plan @jvoorhis ?

PR #19, opened 2017-09-05
Remove inaccurate warning about MFA usage

roooms:
Following a conversation with Kevin about the wording, the decision was made to remove the warning altogether.


skierkowski:
@roooms we should have taken care of this one earlier. But for now I'm going to close without merging because we'll be making changes to these pages soon.

PR #22, opened 2017-09-14
remove arukas provider from document

nasa9084:
since the service has been closed (2017-07-31).


yamamoto-febc:
I commented for the same issue on [hashicorp/terraform#16092](https://github.com/hashicorp/terraform/pull/16092).

nasa9084:
hashicorp/terraform#16092 is closed.

PR #31, opened 2017-09-26
DON'T MERGE: Use jQuery to activate sidebar nav sections

nfagerlund:
The Terraform website uses a collapsing sidebar navigation, where all nested sections should be collapsed except for:

- The link corresponding to the current page.
- Every section that is a parent or grandparent of that link. 

This is all governed by CSS via the `active` class. But right now, we set that class on elements at build time by doing double-bookkeeping on an extra piece of page metadata called `sidebar_current`. 

This system is hard to work with, and it's also redundant; the hierarchy of sections already tells us everything we need to figure out the active parent sections.

There should be an easy and correct way to make sidebars figure themselves out, but I'm pretty sure that way would be CSS parent selectors, which do not exist. (Yet? CSS4? Maybe?) So instead, here are three OK possibilities: 

1. Use JS to do it in the browser. This is super easy with jQuery, which we already use. This PR has a working POC. 
    * Downside: Nav doesn't work without JS. But if users with JS disabled are a concern (?!), we could default sidebars to expanded and set `inactive` classes with JS. 
2. Use Nokogiri to do it at build time. Probably easy enough to code, Nokogiri's interface isn't too bad.
    * Downside: Pulling in Nokogiri as a dependency is, uh... extreme. 
3. Stop representing sidebar navs as HTML (plus templating code), and instead represent them as data, writing a new helper method to generate HTML (with appropriate `active` classes) at build time. 
    * Downside: More work than either 1 or 2... but it's interesting work, and might let us re-use other data Middleman already has in some fun ways. 

So, let's discuss? I'd really like to stop doing this `sidebar_current` double-bookkeeping asap.

### DIFF
--- a/content/source/assets/javascripts/application.js
+++ b/content/source/assets/javascripts/application.js
@@ -4,7 +4,7 @@
 //= require hashicorp/mega-nav
 //= require hashicorp/sidebar
 
-$( document ).ready( function() {
+document.addEventListener("turbolinks:load", function() {
     "use strict";
     var docsSidebar = $("#docs-sidebar ul.nav.docs-sidenav");
     var flexibleLocation = new RegExp(
--- a/content/source/assets/javascripts/application.js
+++ b/content/source/assets/javascripts/application.js
@@ -3,3 +3,19 @@
 
 //= require hashicorp/mega-nav
 //= require hashicorp/sidebar
+
+$( document ).ready( function() {
+    "use strict";
+    var docsSidebar = $("#docs-sidebar ul.nav.docs-sidenav");
+    var flexibleLocation = new RegExp(
+        location.pathname.replace(/(\/|\/index.html)$/, "/(index.html)?") + '$'
+    );
+    var activeLinks = docsSidebar.find("a").filter(
+        function(index, element) {
+            return flexibleLocation.test( $(element).prop("href") );
+        }
+    );
+    var activeListItems = docsSidebar.find("li").has(activeLinks);
+    console.log(activeListItems);
+    activeListItems.addClass("active");
+});
--- a/content/source/assets/javascripts/application.js
+++ b/content/source/assets/javascripts/application.js
@@ -8,7 +8,7 @@ document.addEventListener("turbolinks:load", function() {
     "use strict";
     var docsSidebar = $("#docs-sidebar ul.nav.docs-sidenav");
     var flexibleLocation = new RegExp(
-        location.pathname.replace(/(\/|\/index.html)$/, "/(index.html)?") + '$'
+        location.pathname.replace(/(\/|\/index\.html)$/, "/(index.html)?") + '$'
     );
     var activeLinks = docsSidebar.find("a").filter(
         function(index, element) {
@@ -16,6 +16,5 @@ document.addEventListener("turbolinks:load", function() {
         }
     );
     var activeListItems = docsSidebar.find("li").has(activeLinks);
-    console.log(activeListItems);
     activeListItems.addClass("active");
 });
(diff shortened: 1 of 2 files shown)


nfagerlund:
Another problem with `sidebar_current` is fake parents. I was trying to make sibling pages called `...-vcs-github` and `...-vcs-github-enterprise`, but then I realized it would mark `github` as a parent of `github-enterprise`. So I renamed it to `ghenterprise`.

content/source/assets/javascripts/application.js
```diff
@@ -3,3 +3,18 @@
 
 //= require hashicorp/mega-nav
 //= require hashicorp/sidebar
+
+document.addEventListener("turbolinks:load", function() {
```
sethvargo: I think we also need to add an event listener for "body:load" too, since this will only activate on nav to the page, but I don't think it will activate on direct visit.

sethvargo:
I personally would like to auto-generate the sidebar items.

content/source/assets/javascripts/application.js
nfagerlund: Nah, [it fires on both](https://github.com/turbolinks/turbolinks#running-javascript-when-a-page-loads) (I tested to make sure). Luckily they took the 99%-of-the-time use case into account when designing their event API. 😅

nfagerlund:
@sethvargo Could you say a bit more about that? Auto-generate them from what, and auto-generate how much of that content?

sethvargo:
Hey @nfagerlund 

I think we can auto-generate them from the filenames and by parsing the frontmatter for the sidebar name. Basically just iterate over all files in a directory recursively and build the `<li>`s from it.

nfagerlund:
@sethvargo I do like the idea of pushing names into the source files, but the ordering of pages within a hierarchical section matters quite a lot, and I don't think pushing the ordering into the filesystem makes sense. 

I'd be in favor of keeping the record of the hierarchy closer to the files it governs (like, a TOC file in each dir), but I don't think making it wholly implicit works.

sethvargo:
> I do like the idea of pushing names into the source files, but the ordering of pages within a hierarchical section matters quite a lot, and I don't think pushing the ordering into the filesystem makes sense.

I totally agree, but I think they _should_ match. The URL structure (which is actually the local file structure) should match the content nesting. So if A is a child of B is a child of C, that URL should be A/B/C and the sidebar should probably drill down as A -> B -> C. Are you thinking otherwise?

nfagerlund:
Well, we're talking about two related things when we talk about the sidebar: 

* Hierarchy/parentage
* Ordering of siblings

I don't mind enforcing hierarchy in the filesystem; there's benefits and tradeoffs, but it's generally fine.

Enforcing sibling ordering in the filesystem seems silly, and using an effectively random order would be worse than silly.

sethvargo:
> Enforcing sibling ordering in the filesystem seems silly

I think I'm not following. Could you give me an example of "sibling ordering"? I think we might be talking about different things :smile:.

nfagerlund:
https://www.terraform.io/docs/enterprise-beta/users-teams-organizations/index.html

The elements in this section go in the order:

* Users
* Teams
* Organizations
* Permissions

If you were to just auto-generate the TOC based on the filesystem, it would probably be alphabetical, so it would instead be ordered:

* Organizations 
* Permissions
* Teams
* Users

...Which is wrong. (Similarly, if you zoom out and look at the TOC as whole, auto-generating the nav would put "API" above "Getting Started.") 

To enforce sibling ordering in the filesystem, you could rename the files nonsensically so that alphabetical ordering would put them in the desired order:

* 01users.html.md
* 02teams.html.md
* 03organizations.html.md
* 04permissions.html.md

...which I think is deranged. 

Anyway, since ordering of pages within a section (and ordering of sections within a complete set of documents) is meaningful and important, I don't think it's a good idea to try to intuit a complete TOC from the filesystem alone.

PR #32, opened 2017-09-28
TFE API doc updates Sept 28

skierkowski:
- Update to use `/runs` instead of `/workspaces/:workspace_id/runs` as it is not yet supported
- Adds DELET on workspaces


content/source/docs/enterprise-beta/api/workspaces.html.md
```diff
@@ -284,3 +284,26 @@ $ curl \
 }
 ```
 
+
+## Delete a workspace
+
+This endpoint deletes a workspace
```
nfagerlund: add a "."

PR #34, opened 2017-09-29
Fix "installing software" instructions

KFishner:
it should reference Terraform not Packer

cc @ryanuber @nfagerlund @skierkowski @apparentlymart


nfagerlund:
This seems reasonable... but I don't have the context to make the call. (I've heard TFE can do Packer runs as well.) I defer to @ryanuber ?

ryanuber:
At present TFE definitely still supports both tools. There are talks of removing Packer, so not sure if we want to document that also. But if we are going for TF-only in docs, this is 👌

KFishner:
this page is all about Terraform so I figured to just reference Terraform. Also fixed the YAML header based on @ryanuber's comment. This should be good to go

PR #56, opened 2017-11-15
Fix TFE Team Access API

skierkowski:


content/source/docs/enterprise-beta/api/team-access.html.md
```diff
@@ -29,35 +27,83 @@ $ curl \
   --header "Authorization: Bearer $ATLAS_TOKEN" \
   --header "Content-Type: application/vnd.api+json" \
   --request GET \
-  https://atlas.hashicorp.com/api/v2/team_workspaces?filter[organization][username]=hashicorp&filter[workspace][name]=example&filter[team][id]=1
+  https://atlas.hashicorp.com/api/v2/team-workspaces?filter%5Bworkspace%5D%5Bid%5D=ws-5vBKrazjYR36gcYX
```
nfagerlund: Ugh! 😩 I was about to suggest `--get --data-urlencode field=content`, but it turns out that only urlencodes the `content` part of `field=content` and the problem is all in the `field`! Well....... oh well, I guess. 

Is this structure inherited from JSON API or some foundational thingy we're using, or did we do it ourselves? If it's the latter, I wonder if it'd be easier to work with if the parameter name was just `filter` and it took a filter-specifying object like `{"workspace":{"id":"<ID>"}}`. 

That's probably wack in other ways I'm not thinking of? Anyway, I've gotten off track so I'll leave this out of the review.

content/source/docs/enterprise-beta/api/team-access.html.md
```diff
@@ -102,52 +173,7 @@ $ curl \
...
-{
-  "data": {
-    "type": "team-workspaces",
-    "id": "1",
-    "attributes": { "permission": "read" }
-  }
-}
-```
-
-## Update Team Access to a Workspace
```
nfagerlund: Is the proper way to do this by removing, then adding?

content/source/docs/enterprise-beta/api/team-access.html.md
skierkowski: The `filter` field and the `[]` type are both specified in JSON-API. While it is a little strange to use on the command line with `curl`, it is actually relatively well supported in other clients. I use [rest-client](https://github.com/rest-client/rest-client) which makes this very easy.

content/source/docs/enterprise-beta/api/team-access.html.md
skierkowski: I used the UI as a guide for this, in the UI you remove then add.

PR #69, opened 2017-12-01
Adds TFE API docs for the module registry

skierkowski:
- Adds a menu item for the new page
- Adds API docs for listing, creating and deleting modules from the private module registry


content/source/docs/enterprise-beta/api/modules.html.md
```diff
@@ -0,0 +1,189 @@
...
+sidebar_current: "docs-enterprise2-api-modules"
+---
+
+# Registry Modules API
+
+-> **Note**: These API endpoints are in beta and are subject to change.
+
+## List modules
+
+Lists all the modules for a given organization. **Note** that all other endpoints have the subpath `/api/v2`, this endpoint does not use that subpath.
```
banks: Hmm this ends up being kinda janky.

The real reason is that this is actually part of the Public API although we didn't actually document then "browse a single namespace" option yet. It's hard to know how best to explaina that.

I'd be happy to document this in public API and then explain here that this is part of public API and that TFE maps `namespace` to `organisation` name.

What do you think?

content/source/docs/enterprise-beta/api/modules.html.md
```diff
@@ -0,0 +1,189 @@
...
+```
+
+### Sample Response
+
+```json
+{
+  "meta": {
+    "limit": 15,
+    "current_offset": 0
+  },
```
banks: Should we cross-link to Public API docs here since this pagination format and behaviour is not the same as other TFE JSON API endpoints but is documented in public API?

content/source/docs/enterprise-beta/api/modules.html.md
```diff
@@ -0,0 +1,189 @@
...
+      "downloads": 0,
+      "verified": false
+    }
+  ]
+}
+```
+
+## Publish a Module
+
+This endpoint can be used to publish a new module to the registry. The publishing process will ingress the configuration from the repository and parse the various fields (e.g. input variables, output variables, readme) from the repository.
```
banks: Suggestion:

The publishing process will fetch all tags in the source repository that look like SemVer versions with optional 'v' prefix. For each version, the tag is cloned the config parsed to populate module details (input and output variables, readme, submodules, etc.).

content/source/docs/enterprise-beta/api/modules.html.md
```diff
@@ -0,0 +1,189 @@
...
+
+| Method | Path           |
+| :----- | :------------- |
+| POST | /registry-modules |
+
+### Parameters
+
+- `linkable-repo-id` (`string: <required>`) - Specifies the repository to be used to ingress the configuration, this is in the format `<PROJECT_KEY>/<REPO>`
```
banks: Should we make that more VCS agnostic? For example when we have GitHub support `<PROJECT_KEY>` is a bit confusing since it's presumably just an organisation name in GitHub.

Fair enough for now if the plan is to update it later when other VCS are supported. If we use that terminology globally elsewhere then I guess fine too. It's probably obvious anyway to most devs.

content/source/docs/enterprise-beta/api/modules.html.md
```diff
@@ -0,0 +1,189 @@
...
+- `o-auth-token-id` (`string: <requires>`) - Specifies the VCS Connection (OAuth Conection + Token) to use as identified. This ID can be obtained from the [o-auth-tokens](./o-auth-tokens.html) endpoint.
+
+
+### Sample Payload
+They payload must include `ingress-trigger-attributes`; however, those fields **MUST** be set to the default values provided.
```
banks: > the default values provided

Do you mean the values in the example below? I guess so but not 100% sure, if so I suggest just adding "in the example below" to make it crystal clear.

content/source/docs/enterprise-beta/api/modules.html.md
```diff
@@ -0,0 +1,189 @@
...
+    "links": {
+      "self": "/api/v2/registry-modules/show/skierkowski-v2/instance/aws"
+    }
+  }
+}
+```
+
+## Delete a module
+
+This set of endpoint can be used to delete a single version, all versions for a given module and provider, or all versions for all providers for a given module.
```
banks: We should add that deletes cascade probably. I've tried explaining this in RFCs and internal docs abotu 3 times now and not found a non-clumsy way to put it so you can try and do better but the technical behaviour is:

 1. If you delete a specific provider, and that provider was the only provider for that module, the whole module is deleted.
 2. If you delete a specific version, and that version was the only version for that provider, the whole provider record is deleted, and if the provider was the only provider for that module the module is deleted too.

content/source/docs/enterprise-beta/api/modules.html.md
banks: We have this: https://www.terraform.io/docs/registry/api.html#list-latest-version-of-module-for-all-providers but basically need to add the version without the module name too then cross link this to there and explain that it's the same and response format/pagination is documented there etc.

nfagerlund:
For now I'll defer to @banks on this; I don't have much additional review to add.

content/source/docs/enterprise-beta/api/modules.html.md
nfagerlund: How about this:

> These endpoints can delete a single version for a provider, a single provider (and all its versions) for a module, or a whole module. If the requested deletion would leave a provider with no versions or a module with no providers, the empty items will be automatically deleted as well.

banks: Sounds great :)

content/source/docs/enterprise-beta/api/modules.html.md
banks: @mitchellh and I agreed we should just document this endpoint in the public docs and then link there like with the other public endpoints.

I started looking at that and realise it's actually a little bit hairy since right now we don't list _any_ of the browse or search endpoints in public - it was a TODO item. Im going to fix that up so we can cross link but now also hitting issues with the way we specified Discovery documents - they currently include trailing slash in the "prefix" however the actual API for listing modules 301 redirects that back to no slash which is pretty gross to try an document.

I'll think about it while I'm travelling next week and either test out and change the discovery documents to not include trailing (we think TF CLI will handle that already), or I'll leave that but change the public API a little and update the registry and TFE so they aren't relying on 301 redirects.

Kicking myself for not spotting this silly little one character that makes this all messy earlier.

**tl;dr** you should probably not bother documenting this here and link to Public API (https://www.terraform.io/docs/registry/api.html) instead for all read operations with a note about what the path prefix is on TFE maybe?

content/source/docs/enterprise-beta/api/modules.html.md
mitchellh: I also recommend linking to the guidelines for publishing a module here (standard module structure, maybe the public docs too) so its clearer what is to be expected.

content/source/docs/enterprise-beta/api/modules.html.md
mitchellh: I'd cut it off at the comma and then make sentence 2:

> For BitBucket server, the format is `<PROJECT_KEY>/<REPO>`. This is the only supported VCS right now.

content/source/docs/enterprise-beta/api/modules.html.md
mitchellh: Ping @ryanuber if we can get these defaulted before publishing these docs so that we don't need to specify them at all.

skierkowski: Just checked with Ryan and this should be ready soon so I'm going to remove `ingress-trigger-attributes` altogether.

content/source/docs/enterprise-beta/api/modules.html.md
skierkowski: Yeah, good call. I'll add more context around it and link to those docs.

content/source/docs/enterprise-beta/api/modules.html.md
skierkowski: I just gutted this section and replaced it with the reference to the Public API docs with details on the deltas and an example.

content/source/docs/enterprise-beta/api/modules.html.md
```diff
@@ -0,0 +1,145 @@
...
+
+# Registry Modules API
+
+-> **Note**: These API endpoints are in beta and are subject to change.
+
+## Listing and reading modules, providers and versions
+
+The Terraform Enterprise Module Registry provides the same API endpoints as the [Terraform Registry HTTP API](../../registry/api.html). Terraform Enterprise extends this API to add [authentication](./index.html#authentication). Refer to the [Module Registry HTTP API](../../registry/api.html) to perform the following:
+
+- List available versions for a specific module
```
banks: I have a PR ready to go to add listing modules and search modules in the public API docs. I'll post back here if this is still open or if not I'll try make sure this is updated once that lands.

content/source/docs/enterprise-beta/api/modules.html.md
```diff
@@ -0,0 +1,145 @@
...
+sidebar_current: "docs-enterprise2-api-modules"
+---
+
+# Registry Modules API
+
+-> **Note**: These API endpoints are in beta and are subject to change.
+
+## Listing and reading modules, providers and versions
+
+The Terraform Enterprise Module Registry provides the same API endpoints as the [Terraform Registry HTTP API](../../registry/api.html). Terraform Enterprise extends this API to add [authentication](./index.html#authentication). Refer to the [Module Registry HTTP API](../../registry/api.html) to perform the following:
```
banks: Maybe:

> The Terraform Enterprise Module Registry implements the [Registry standard API](../../registry/api.html) for consuming modules. Refer to the [Module Registry HTTP API](../../registry/api.html) ...

content/source/docs/enterprise-beta/api/modules.html.md
```diff
@@ -0,0 +1,145 @@
...
+- Download source code for a specific module version
+- List latest version of a module for all providers
+- Get the latest version for a specific module provider
+- Get a specific module
+- Download the latest version of a module
+
+The TFE Module Registry endpoints differs from the Module Registry endpoints in the following ways:
+
+- The `:namespace` parameter should be replaced with the organization name.
+- All other TFE endpoints are on the `/api/v2` subpath; the module registry endpoint has the subpath `/api/registry`.
```
banks: It's `/api/v1/modules`. I suggest changing this to:

> The module registry endpoints have the path prefix provided in the [discovery document](../../registry/api.html#service-discovery) which is currently`/api/v1/modules`.

content/source/docs/enterprise-beta/api/modules.html.md
```diff
@@ -0,0 +1,145 @@
...
+```shell
+$ curl \
+  --header "Authorization: Bearer $ATLAS_TOKEN" \
+  https://atlas.hashicorp.com/api/registry/v1/modules/my-tfe-org/consul/aws/versions
+```
+
+
+## Publish a Module
+
+This endpoint can be used to publish a new module to the registry. The publishing process will fetch all tags in the source repository that look like SemVer versions with optional 'v' prefix. For each version, the tag is cloned and the config parsed to populate module details (input and output variables, readme, submodules, etc.). The [Module Registry Requirements](../../registry/modules/publish.html#requirements) defines additional requirements on naming, standard module structure and tags for releases.
```
banks: typo: "Requirements ... defines" should be "define"?

content/source/docs/enterprise-beta/api/modules.html.md
skierkowski: The discovery endpoint is `/api/registry/v1` the CUD endpoints are on `/api/v2/registry-modules`; or are you saying that this is changing?

PR #72, opened 2017-12-05
Force setting acl-public for images

catsby:
Add a post-put step that sets `--acl-public` for `gif` and `png` images


catsby:
Should address our occasional issue with images not getting the right acl

PR #93, opened 2017-12-30
Update part3.2.html.md

charleswhchan:
Combine links for "Terraform" & "modules". Prevents a space break when mouse over hyperlink text.


nfagerlund:
Ha, that looks like some html &lt;> md round trip jank. Thank you for spotting it!

PR #94, opened 2017-12-31
Update part3.2.html.md

charleswhchan:
Suggest edit to clarify the rational for using configuration management tool.

Also: it may be useful to add include this link to clarify the differences:
https://www.thoughtworks.com/insights/blog/why-configuration-management-and-provisioning-are-different


nfagerlund:
Sure, why not.

PR #96, opened 2018-01-09
Clean up presentation of run states, and add policy check states

nfagerlund:
Showing this as a flat list of states was confusing, and didn't clarify the
progression of a run through its lifecycle. Splitting the states into
groups/stages hopefully makes this more legible.

### DIFF
--- a/content/source/docs/enterprise/workspaces/run-basics.html.md
+++ b/content/source/docs/enterprise/workspaces/run-basics.html.md
@@ -26,17 +26,79 @@ When you request a run, TFE locks the run to the current Terraform code (usually
 
 ## Run States
 
-A given Terraform run can be in the following states:
-
-- **Pending:** A pending run hasn't been planned or applied. It's waiting for other runs to finish.
-- **Planning:** TFE is currently running `terraform plan`.
-- **Planned:** TFE has run `terraform plan`, and is waiting for a user to approve a `terraform apply`.
-- **Applying:** TFE is currently running `terraform apply`.
-- **Applied:** TFE has finished applying. The next run can now proceed.
-- **Plan Errored:** The `terraform plan` command failed, and you won't be able to apply this run. This usually means you need to fix variables or code. The next run can now proceed.
-- **Discarded:** The `terraform plan` command succeeded, but a user chose not to apply it. The next run can now proceed.
+Each run passes through several stages of action, and TFE shows the progress through those stages as run states.
 
 In the list of workspaces on TFE's main page, each workspace shows the state of its most recently queued run. So if any runs are pending, the workspace is in "pending," regardless of what it's currently running or waiting for.
+
+### 1. The Pending Stage
+
+_States in this stage:_
+
+- **Pending:** TFE hasn't started action on a run yet. TFE processes each workspace's runs in the order they were queued, and a run remains pending until every run before it has completed.
+
+_Leaving this stage:_
+
+- Proceeds automatically to the plan stage (**Planning** state) when it becomes the first run in the queue.
+- Can skip to completion (**Discarded** state) if a user discards it before it starts.
+
+### 2. The Plan Stage
+
+_States in this stage:_
+
+- **Planning:** TFE is currently running `terraform plan`.
+- **Planned:** `terraform plan` has finished. TFE sometimes pauses in this state, depending on the workspace and organization settings.
+
+_Leaving this stage:_
+
+- If the `terraform plan` command failed, TFE skips to completion (**Plan Errored** state).
+- If the plan succeeded:
+    - If [Sentinel policies][] are enabled, TFE proceeds automatically to the policy check stage.
+    - If there are no Sentinel policies and auto-apply is enabled on the workspace, TFE proceeds automatically to the apply stage.
+    - If there are no Sentinel policies and auto-apply is disabled, TFE pauses in the **Planned** state until a user with write access to the workspace takes action. The run proceeds to the apply stage if they approve the apply, or skips to completion (**Discarded** state) if they reject the apply.
+
+### 3. The Policy Check Stage
+
+This stage only occurs if [Sentinel policies][] are enabled. After a successful `terraform plan`, TFE checks whether the plan obeys policy to determine whether it can be applied.
+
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/workspaces/run-basics.html.md
```diff
@@ -26,17 +26,79 @@ When you request a run, TFE locks the run to the current Terraform code (usually
 
 ## Run States
 
-A given Terraform run can be in the following states:
+Each run passes through several stages of action, and TFE shows the progress through those stages as run states.
+
+In the list of workspaces on TFE's main page, each workspace shows the state of its most recently queued run. So if any runs are pending, the workspace is in "pending," regardless of what it's currently running or waiting for.
```
beekus: This is wrong. We are using current run now.

nfagerlund: Ah, thank you. I'll fix.

content/source/docs/enterprise/workspaces/run-basics.html.md
```diff
@@ -26,17 +26,79 @@ When you request a run, TFE locks the run to the current Terraform code (usually
 
 ## Run States
 
-A given Terraform run can be in the following states:
+Each run passes through several stages of action (pending, plan, policy check, apply, and completion), and TFE shows the progress through those stages as run states.
+
+In the list of workspaces on TFE's main page, each workspace shows the state of the run it's currently processing. (Or, if no run is in progress, the state of the most recent completed run.)
```
beekus: 👍

content/source/docs/enterprise/workspaces/run-basics.html.md
```diff
@@ -26,17 +26,79 @@ When you request a run, TFE locks the run to the current Terraform code (usually
...
+_States in this stage:_
 
-- **Pending:** A pending run hasn't been planned or applied. It's waiting for other runs to finish.
 - **Planning:** TFE is currently running `terraform plan`.
-- **Planned:** TFE has run `terraform plan`, and is waiting for a user to approve a `terraform apply`.
+- **Planned:** `terraform plan` has finished. TFE sometimes pauses in this state, depending on the workspace and organization settings.
+
+_Leaving this stage:_
+
+- If the `terraform plan` command failed, TFE skips to completion (**Plan Errored** state).
```
beekus: **Plan Errored** you have a few places. 

It may be better captured as just **Errored**, since Plan kind of refers to two things:
* The stage of the run: *Your terraform run is in the plan stage*
* Sort of refers to the configuration????: *Your terraform plan has errored in the applying stage*

I don't remember exactly what it says in the UI.

content/source/docs/enterprise/workspaces/run-basics.html.md
```diff
@@ -26,17 +26,79 @@ When you request a run, TFE locks the run to the current Terraform code (usually
...
+- If the plan succeeded:
+    - If [Sentinel policies][] are enabled, TFE proceeds automatically to the policy check stage.
+    - If there are no Sentinel policies and auto-apply is enabled on the workspace, TFE proceeds automatically to the apply stage.
+    - If there are no Sentinel policies and auto-apply is disabled, TFE pauses in the **Planned** state until a user with write access to the workspace takes action. The run proceeds to the apply stage if they approve the apply, or skips to completion (**Discarded** state) if they reject the apply.
```
beekus: This appears in the UI as **Needs Confirmation** I believe.

content/source/docs/enterprise/workspaces/run-basics.html.md
```diff
@@ -26,17 +26,79 @@ When you request a run, TFE locks the run to the current Terraform code (usually
...
+- When the apply finishes, TFE proceeds automatically to completion (**Applied** state).
+
+### 5. Completion
+
+A run is considered completed if it finishes applying, if the plan or policy check fails, or if a user chooses not to continue it. Once a run is completed, the next run in the queue can enter the plan stage.
+
+_States in this stage:_
+
+- **Applied:** TFE has finished applying.
+- **Plan Errored:** The `terraform plan` command failed (usually requiring fixes to variables or code), or a hard-mandatory Sentinel policy failed. The run cannot be applied.
```
beekus: `terraform apply` command could have failed as well.

content/source/docs/enterprise/workspaces/run-basics.html.md
nfagerlund: It says "Plan Errored" in the UI, so I've been matching that in the docs. But I agree that plain old "Errored" seems like a better UI label for that! I'm in favor of changing in both places, but the UI should probably change first.

content/source/docs/enterprise/workspaces/run-basics.html.md
nfagerlund: ah! I'd totally forgotten about that! Thank you.

content/source/docs/enterprise/workspaces/run-basics.html.md
nfagerlund: About that! How can I learn which state (in the UI) it goes to if that happens? I don't know enough about Terraform to craft a configuration where the plan will succeed but the apply will fail, so I haven't been able to observe that yet.

nfagerlund: (...it's probably Plan Errored.)

beekus: hmmmmm. someone with more terraform experience might be better suited to answer this...

oooo, you could have a plan that creates more resources than you are allowed in AWS. Like, try to create 10 VPCs in a region (I think the default limit is 5 VPCs). I would guess the plan would pass, but the apply would fail?

nfagerlund: image

Indeed, there's an Apply Errored state! And I found a way to do it myself: apply a configuration with resources from (some provider), then delete all those resources (and any provider config block) and apply the config that would delete those resources. TFE will fail at apply until you add a provider config block for that provider.

nfagerlund:
K, I've got @beekus's approval from earlier and have followed up and double-checked the lingering questions and adjustments. Gonna merge!

PR #98, opened 2018-01-10
Fix versions typo in example for create

T-Wong:


nfagerlund:
Thanks for catching this!

PR #107, opened 2018-01-23
Add UI/concepts documentation for variables

nfagerlund:
For review: 

- I think @ryanuber is the one to ask about the "how TFE uses variables" section at the bottom. Did I get this right, and are there any extra nuances a curious user should know about? The bit about Archivist is what I'm least sure about, I think.
- @alisdair, I think you were the last person to change the UI behavior. Are the UI descriptions right? 
- @skierkowski, I went ahead and just left out anything about personal variables, like we talked about.


content/source/docs/enterprise/workspaces/variables.html.md
```diff
@@ -0,0 +1,94 @@
...
+- You cannot commit a file named `terraform.tfvars` to version control, since TFE will overwrite it. (Note that you shouldn't check in `terraform.tfvars` even when running Terraform solely on the command line.)
+- In the event that HCL syntax changes slightly between Terraform versions, all values in a given workspace will be handled consistently, since the variables file is interpreted by the same Terraform version that loads the rest of the code.
+
+### Environment Variables
+
+TFE performs Terraform runs on disposable Linux worker VMs using a POSIX-compatible shell. Before running Terraform, TFE populates the shell with environment variables using the `export` command.
+
+### Secure Storage of Variables
+
+TFE's backend storage system encrypts variables on disk using AES-128, and uses Vault to protect the encryption key.
```
ryanuber: Was this written assuming that Archivist was used for variable storage? It would probably be better to write this like:

> TFE encrypts all variable values securely using Vault's transit backend prior to saving them. This ensures that no out-of-band party is able to read these values without proper authorization.

nfagerlund: Yeah, the legacy TFE docs said "variables are stored in vault," and after looking into it for a while I landed on the Archivist README. 

I just read up briefly on the transit backend, but it's new to me so I want to make sure I understand the implications: This means the data is encrypted in its at-rest storage, right? Archivist doesn't necessarily understand how to decrypt it, but that's fine because access goes through Vault, so Vault handles the decryption on access? ...Did I get that right?

nfagerlund:
@alisdair Huh, it doesn't look like that anymore! Either it got fixed after I made the screenshots, or it was a result of weird CSS caching, or I had extra newlines that I later deleted, for some reason. I'll update the screenshots when I follow up on Ryan's edit.

content/source/docs/enterprise/workspaces/variables.html.md
skierkowski: Ryan's suggested statement is exactly what I'd like to read. It calls out the use of Vault transit backend and why it's used.

I don't think we need to call out how the unseal keys are managed or where the data gets stored, as that level of detail is only necessary for PTFE (not SaaS) and it'll be configured in the install docs for PTFE. (Though I'm guessing you want to know this for your own benefit, not necessarily for the docs.)

nfagerlund:
Sweet, that sounds like approval to merge. Thanks all!

PR #120, opened 2018-02-01
Clarify ssh submodules

armchairlinguist:
I had a long conversation with @beekus about this as a result of customer confusion (and my own!) and hope these edits will make it clearer which SSH key is/keys are for what and when.

### DIFF
--- a/content/source/docs/enterprise/workspaces/settings.html.md
+++ b/content/source/docs/enterprise/workspaces/settings.html.md
@@ -12,7 +12,7 @@ In Terraform Enterprise, the workspace's associated repository can only be confi
 
 The directory where Terraform will execute. This defaults to the root of the repository.
 
-If the repo contains multiple environments, set this to the subdirectory matching the workspace's environment. More at [Repo Structure](./repo-structure.html)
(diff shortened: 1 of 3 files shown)


content/source/docs/enterprise/workspaces/settings.html.md
```diff
@@ -25,5 +24,12 @@ Most users don't need external files to be deleted, and should use the Terraform
 
 ## VCS branch
 
-The branch to import new versions from. This defaults to your
-version control's default branch for the repository.
+The branch to import new versions from. This defaults to the version control system's default branch for the repository.
+
+## SSH Key
+
+Set an SSH key here if the configuration sources Terraform modules from other repos. The SSH key must have read access to the module repos. See also [Assigning Keys to Workspaces](./ssh-keys#assigning-keys-to-workspaces). If no SSH keys are listed here, visit "Manage SSH Keys" in the Organization settings or review [Adding and Deleting Keys](./ssh-keys.html#adding-and-deleting-keys).
```
catsby: There's a broken link here, `./ssh-keys#` should be `./ssh-keys.html#`.

Fixed in https://github.com/hashicorp/terraform-website/pull/121

PR #121, opened 2018-02-02
Fix broken link

catsby:
Fixes broken link on https://www.terraform.io/docs/enterprise/workspaces/settings.html that was accidentally introduced in https://github.com/hashicorp/terraform-website/pull/120


armchairlinguist:
thanks @catsby !

PR #123, opened 2018-02-05
Fix tar to change into config directory

roooms:
The generated archive must not package the config directory but rather the files inside the config directory. The output of `tar tvzf my-configuration.tar.gz` should be something like this:

```
drwxr-xr-x  0 dan    staff       0  5 Feb 16:50 ./
-rw-r--r--  0 dan    staff      76 29 Jan 21:12 ./.gitignore
-rw-r--r--  0 dan    staff   16725 29 Jan 21:12 ./LICENSE
-rw-r--r--  0 dan    staff      32 30 Jan 08:35 ./main.tf
-rw-r--r--  0 dan    staff      98 30 Jan 09:59 ./outputs.tf
-rw-r--r--  0 dan    staff      12 29 Jan 21:12 ./README.md
-rw-r--r--  0 dan    staff       1 30 Jan 09:03 ./variables.tf
```

not this:

```
drwxr-xr-x  0 dan    staff       0  5 Feb 16:50 my-configuration/
-rw-r--r--  0 dan    staff      76 29 Jan 21:12 my-configuration/.gitignore
-rw-r--r--  0 dan    staff   16725 29 Jan 21:12 my-configuration/LICENSE
-rw-r--r--  0 dan    staff      32 30 Jan 08:35 my-configuration/main.tf
-rw-r--r--  0 dan    staff      98 30 Jan 09:59 my-configuration/outputs.tf
-rw-r--r--  0 dan    staff      12 29 Jan 21:12 my-configuration/README.md
-rw-r--r--  0 dan    staff       1 30 Jan 09:03 my-configuration/variables.tf
```

Thanks,
Dan

### DIFF
--- a/content/source/docs/enterprise/workspaces/run-api.html.md
+++ b/content/source/docs/enterprise/workspaces/run-api.html.md
@@ -53,6 +53,8 @@ WORKSPACE_NAME="$(cut -d'/' -f2 <<<"$2")"
 
 The [configuration version API](../api/configuration-versions.html) expects a tar.gz file to be uploaded, so the configuration directory must be packaged into a tar.gz.
 
+~> **Important:** The configuration directory must be the root of the tar file, with no intermediate directories. In other words, when the tar file is extracted the result must be paths like `./main.tf` rather than `./terraform-appserver/main.tf`.
+
 ```bash
 UPLOAD_FILE_NAME="./content-$(date +%s).tar.gz"
 tar cvzf $UPLOAD_FILE_NAME -C $CONTENT_DIRECTORY .
--- a/content/source/docs/enterprise/workspaces/run-api.html.md
+++ b/content/source/docs/enterprise/workspaces/run-api.html.md
@@ -55,7 +55,7 @@ The [configuration version API](../api/configuration-versions.html) expects a ta
 
 ```bash
 UPLOAD_FILE_NAME="./content-$(date +%s).tar.gz"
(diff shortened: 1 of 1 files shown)


nfagerlund:
Does it mess things up if there's an intervening directory? If so, I think we should add a more explicit warning in addition to changing the example, because that's an easy thing to mess up.

Or better yet, we should make it so TFE does the right thing if the right thing is obvious (e.g. in a tarball with just one directory, that's probably the directory you wanted to be the root module).

roooms:
The `tar -C` command has this guidance:

```
-C directory
             In c and r mode, this changes the directory before adding the following files.  In x mode,
             change directories after opening the archive but before extracting entries from the archive.
```

I haven't tested multiple examples, but this looks like it should allow for changing directory into the root module regardless of the folder structure, as long as `$CONTENT_DIRECTORY` is specified correctly.

nfagerlund:
Bah, sorry, that was unclear of me. What I meant was: does TFE fail to run the configuration if you tar up the directory? If so, we should be explicit about that in this page, because I think a lot of readers will be like, "ah, a tar command, I know how to run tar," and won't really _see_ the `-C` option unless they understand why you're doing it.

roooms:
Took a while until I could test this again to be certain. If TFE receives a packaged directory rather than just the directory contents, it finds no terraform configuration to plan against as it is not at the root of the package. There is not an error as such, but TFE shows there are no changes to make.

nfagerlund:
I went ahead and added it.

PR #136, opened 2018-02-16
Remove documentation on vcs-root-path

davidcelis:
This is being removed.


nfagerlund:
Sorry I didn't catch this when it initially went in! I found one more reference to that setting.

PR #139, opened 2018-02-22
Fixes formatting for airgapped instructions

petems:
* Before, the numbering wasn’t formatted
* Adding this newline fixes it


petems:
Before:
Before

After:
After

nfagerlund:
Hmm, that sub-item looks like maybe it's not supposed to be numbered. That smells like a bit of a google docs translation glitch to me. But I can look into that later, let's just merge this.

petems:
@nfagerlund Oh yeah, that should probably be a sub-bullet rather than a number... and the URL might be nicer as `https://[hostname or ip of your instance]:8800` but I guess that's personal taste?

PR #142, opened 2018-02-27
Clarify the usage of self-signed TLS certs

evanphx:
Self-signed certs are rejected by default by VCS providers when sending webhooks, so we should not recommend their usage.

### DIFF
--- a/content/source/docs/enterprise/private/install-installer.html.md
+++ b/content/source/docs/enterprise/private/install-installer.html.md
@@ -22,7 +22,15 @@ instance ready.
 
 ### Data Files
 
-* TLS private key and certificate _Optional: Self-signed materials can be used_
+* TLS private key and certificate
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/private/install-installer.html.md
```diff
@@ -22,7 +22,15 @@ instance ready.
 
 ### Data Files
 
-* TLS private key and certificate _Optional: Self-signed materials can be used_
+* TLS private key and certificate
+  * The installer allows for using a self-signed certificate but HashiCorp does
+    _not_ recommended this. Your VCS provider will likely reject that certificate
+    when sending webhooks. If you do use the self-signed certificate, you must configured
+    each webhook to ignore SSL errors within your VCS provider.
```
phinze: s/configured/configure/

PR #144, opened 2018-02-28
TFE API update workspace Terraform version

dylanegan:
By passing through the `terraform_version` parameter, it is possible to set the version of Terraform for the given workspace.

This work is most likely superseded by upcoming API documentation changes, but I wanted to get it in so that the `terraform_version` attribute would be tracked. /cc @davidcelis 

This could be merged in to master and deployed before those updates or simply closed if the updates include this attribute.


dylanegan:
@nfagerlund I rebased against master and I also added documentation for creating because that is also available, although it is just documented under the VCS connected creation section.

PR #145, opened 2018-02-28
Update Workspaces API documentation to account for removal of compound workspaces

davidcelis:
A deployment changed/removed these API endpoints earlier today, so we should update the docs.

### DIFF
--- a/content/source/docs/enterprise/api/workspaces.html.md
+++ b/content/source/docs/enterprise/api/workspaces.html.md
@@ -21,10 +21,15 @@ This endpoint is used to create a new workspace either with or without a VCS con
 ### Parameters
 
 - `:organization_username` (`string: <required>`) - Specifies the username or organization name under which to create the workspace. The organization must already exist in the system, and the user must have permissions to create new workspaces. This parameter is specified in the URL path.
+- `name`
 
 ### Create a Workspace Without a VCS Repository
 
-If you supply nothing but a name, you can create a Workspace without configuring it against a VCS Repository.
+If you supply nothing but a name, you can create a workspace without configuring it against a VCS repository.
+
+#### Parameters
+
+- `name` (`string: <required>`) - Specifies the name of the workspace, which can only include letters, numbers, `-`, and `_`. This will be used as an identifier and must be unique in the organization.
 
 #### Sample Payload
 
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/api/workspaces.html.md
```diff
@@ -10,61 +10,48 @@ sidebar_current: "docs-enterprise2-api-workspaces"
...
-- `oauth-token-id` (`string: <optional>`) - Specifies the VCS Connection (OAuth Conection + Token) to use as identified. This ID can be obtained from the [oauth-tokens](./oauth-tokens.html) endpoint.
-- `default-branch` (`boolean: true`) - specifies if the default branch should be used.
-- `ingress-submodules` (`boolean: false`) - Specifies whether submodules should be fetched when cloning the VCS repository.
-- `linkable-repo-id` (`string: <optional>`) - This is the reference to your VCS repository in the format :org/:repo where :org and :repo refer to the organization and repository in your VCS provider.
-- `name` (`string: <required>`) - Specifies the name of the workspace, which can only include letters, numbers, `-`, and `_`. This will be used as an identifier and must be unique in the organization.
-- `working-directory` (`string:''`) - Specifies the directory that Terraform will execute within. This defaults to the root of your repository and is typically set to a subdirectory matching the environment when multiple environments exist within the same repository.
+If you supply nothing but a name, you can create a Workspace without configuring it against a VCS Repository.
```
nfagerlund: In sentence text, objects like "workspace" and "repository" shouldn't be capitalized as proper nouns.

nfagerlund: There should be a parameters list here that includes `name`, since it looks like that's required.

content/source/docs/enterprise/api/workspaces.html.md
```diff
@@ -116,35 +98,40 @@ $ curl \
...
 
+By supplying the necessary attributes under a `vcs-repository` object, you can create a Workspace that is configured against a VCS Repository.
 
+#### Parameters
 
-### Parameters
-
-- `:organization_username` (`string: <required>`) - Specifies the username or organization name under which to create the workspace. The organization must already exist in the system, and the user must have permissions to create new workspaces. This parameter is specified in the URL path.
+- `name` (`string: <required>`) - Specifies the name of the workspace, which can only include letters, numbers, `-`, and `_`. This will be used as an identifier and must be unique in the organization.
+- `working-directory` (`string: ''`) - Specifies the directory that Terraform will execute within. This defaults to the root of your repository and is typically set to a subdirectory matching the environment when multiple environments exist within the same repository.
```
nfagerlund: I realize this is carried over verbatim, but 1. could use some clarification about what the empty string means, and 2. could stand to specify that this is a relative path.

davidcelis: I'm guessing that these all take the form of `type: default_value|<optional>` so my guess is that it's just that we default it to an empty string which doesn't do anything. I'll update the relative path bit though!

PR #148, opened 2018-02-28
Clarify scope of tokens, especially re: private modules

nfagerlund:
After talking with Ryan, I decided to NOT raise the possibility of overwriting ATLAS_TOKEN in the context of private modules.

### DIFF
--- a/content/source/docs/enterprise/registry/publish.html.md
+++ b/content/source/docs/enterprise/registry/publish.html.md
@@ -94,3 +94,9 @@ To delete a module or version:
 ~> **Note:** If a deletion would leave a module with no versions, the module will be automatically deleted.
 
 ![TFE screenshot: the deletion dialog](./images/publish-delete.png)
+
+## Sharing Modules Across Organizations
+
+In normal operation, TFE doesn't allow one organization's workspaces to use private modules from a different organization. (When TFE runs Terraform, it provides temporary credentials that are only valid for the workspace's organization, and uses those credentials to access modules.) And although it's possible to mix modules from multiple organizations when running Terraform on the command line, we strongly recommend against it.
+
+However, you can easily share modules across organizations by sharing the underlying VCS repository. Grant each organization access to the module's repo, then add the module to each organization's registry. When you push tags to publish new module versions, both organizations will update appropriately.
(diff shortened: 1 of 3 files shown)


rberlind:
The docs seem ok, but the only reference to using a module from a different organization is:

"When you authenticate with a user token, you can access modules from any organization you are a member of. (A user is a member of an organization if they belong to any team in that organization.)"

I feel that we could say more here such as: "If you reference a module in one organization from a workspace in a second organization, this will work as long as you are a member of both organizations."

nfagerlund:
@rberlind But it _won't_ work when TFE is doing the run, right? Unless you overwrite ATLAS_TOKEN, which we're discouraging for this purpose.

skierkowski:
you are right, it will NOT work if you are running it in TFE, but it will work if running locally. In other words, when running locally you are probably using your personal ATLAS_TOKEN, and when running TFE you are probably using the auto-generated token.

My general feedback is this:
- We must document if personal ATLAS_TOKENS and auto generated tokens have cross-org access.
- We should discourage the use of a personal ATLAS_TOKEN in a workspace, either by not documenting or explicitly calling out that it is discouraged.
- We may document the use of the use of personal ATLAS_TOKEN in a workspace, but only as an exception and specific to use cases (e.g. manually migration), other than that, it shouldn't be called out.
- If there are some exceptions, support team may reference these docs on access levels to explain the access levels, but the use of the personal ATLAS_TOKEN should be called out in 1-1 communications with a customer.

As a side note, the one use case for the private module registry and cross-org access came from a customer who has since then found an alternative design which no longer requires this.

rberlind:
Apologies, @nfagerlund.  I was confused because runs in TFE UI against module in different org used to work in older version, but then we tightened security to require ATLAS_TOKEN as Maciej described. I had forgotten about that change.

nfagerlund:
Cool cool. In that case, shall I go ahead and merge?

rberlind:
I have no objection to merge.

skierkowski:
> I think this PR stays within your guidelines by not mentioning personal tokens as an option;

@nfagerlund 👍 for me

PR #149, opened 2018-03-01
TFE API workspace SSH key association/dissociation

dylanegan:
A simple update to document an endpoint that allows you to associate and dissociate an SSH key to and from a workspace.

I used `an SSH key` as per documented elsewhere. I also included the difference responses for when you associate and dissociate an SSH key. It'd be great to draw attention to the parts of the response that matter, but unsure how best to do that.

### DIFF
--- a/content/source/docs/enterprise/api/ssh-keys.html.md
+++ b/content/source/docs/enterprise/api/ssh-keys.html.md
@@ -1,0 +1,217 @@
+---
+layout: enterprise2
+page_title: "SSH Keys - API Docs - Terraform Enterprise"
+sidebar_current: "docs-enterprise2-api-ssh-keys"
+---
+
+# SSH Keys
+
+-> **Note**: These API endpoints are in beta and are subject to change.
+
+The `ssh-key` object represents an SSH key which includes a name and the SSH private key. This object is used to clone Terraform modules at the organization level, and allows multiple keys to be added for the organization. You can also easily use them in any [workspace](./workspaces.html#assign-an-ssh-key-to-a-workspace) that clones modules from a Git server.
+
+## List SSH Keys
+
+List all the SSH Keys for a given organization
+
+| Method | Path           |
+| :----- | :------------- |
+| GET | /organizations/:organization_name/ssh-keys |
+
+### Parameters
+
+- `:organization_name` (`string: <required>`) - specifies the organization name where the SSH Keys are defined
+
+### Sample Request
+
+```shell
+curl \
+  --header "Authorization: Bearer $ATLAS_TOKEN" \
+  https://app.terraform.io/api/v2/organizations/my-organization/ssh-keys
+```
+
+### Sample Response
+
+```json
+{
+  "data": [
+    {
+      "attributes": {
+        "name": "SSH Key"
+      },
+      "id": "sshkey-GxrePWre1Ezug7aM",
+      "links": {
+        "self": "/api/v2/ssh-keys/sshkey-GxrePWre1Ezug7aM"
+      },
+      "type": "ssh-keys"
+    }
+  ]
+}
+```
+
+
+## Get an SSH Key
+
+Get an SSH Key
+
+| Method | Path           |
+| :----- | :------------- |
+| GET | /ssh-keys/:ssh_key_id |
+
+### Parameters
+
+- `:ssh_key_id` (`string: <required>`) - specifies the SSH key ID to get
+
+### Sample Request
+
+```shell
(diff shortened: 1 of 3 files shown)


content/source/docs/enterprise/api/workspaces.html.md
```diff
@@ -553,3 +553,172 @@ $ curl \
   }
 }
 ```
+
+
+## Associate an SSH key to a workspace
```
nfagerlund: This wording is odd, though I'm a little hard-pressed to say exactly why. Anyway: in [the related UI docs](https://www.terraform.io/docs/enterprise/workspaces/ssh-keys.html), I used "assign [an SSH key] to," but I think "associate [...] with" would also work; I'm fine with either. 

(And "dissociate [...] from" doesn't have the same problem. ENGLISH.)

content/source/docs/enterprise/api/workspaces.html.md
```diff
@@ -553,3 +553,172 @@ $ curl \
...
+This endpoint associates an SSH key to a workspace.
+
+| Method | Path           |
+| :----- | :------------- |
+| PATCH | /workspaces/:workspace_id/relationships/ssh-key |
+
+### Parameters
+
+* `:workspace_id` (`string: <required>`) - Specifies the workspace ID to associate the SSH key to.
+* `id` (`string: <required>`) - Specifies the SSH key ID to associate.
```
nfagerlund: Could use a quick mention of which endpoint I should use to obtain an SSH key ID.

nfagerlund: ...Actually, is it an undocumented endpoint? I'm not finding anything about SSH keys in a quick search.

dylanegan: @nfagerlund yeah, it is undocumented. Shall I document it as part of this PR or handle it separately?

dylanegan: @nfagerlund looks like the API is really required here I guess, as there's no way in the UI to gather this information, unless we told them to inspect the source or added the ID to the UI on the manage page.

content/source/docs/enterprise/api/workspaces.html.md
dylanegan: I'll go with `associate .... with` in this context since `dissociate` is being used. I notice in the UI it talks about `provide` rather than `assign`. I'd be happy to make changes across the board so that the terminology is the same if you'd prefer that.

content/source/docs/enterprise/api/workspaces.html.md
nfagerlund: In that case, I think add the documentation in this PR. We'll need a new page and a new item in the nav sidebar. IDK if you've done the sidebar business before, but the file is at content/source/layouts/enterprise2.erb. The only really squirrelly part is the `sidebar_current` business, but you can probably figure it out from the existing examples.

dylanegan: Sounds good. Thanks for the pointers.

dylanegan:
@nfagerlund I decided to go with assign/unassign in the end because it works better, although `unassign` itself is a bit of a weird one itself. It at least brings it inline with the other documentation for now.

I also add `SSH Keys` and updated the sidebar. Feel free to recommend wording changes to the opening paragraph. I linked directly to the API for `Workspaces` where you assign a key because I thought that's a nice jump to get to.

I also linked back to `SSH Keys` from the `Workspaces` document about how to obtain an `SSH Key ID`.

content/source/docs/enterprise/api/ssh-keys.html.md
```diff
@@ -0,0 +1,217 @@
...
+        "self": "/api/v2/ssh-keys/sshkey-GxrePWre1Ezug7aM"
+      },
+      "type": "ssh-keys"
+    }
+  ]
+}
+```
+
+
+## Get an SSH Key
```
nfagerlund: ...Is there any purpose to ever doing this action? I guess if you have an ID and need to map it to a name.

This page could use some clarification that we never allow you to obtain the actual private key text, but I won't hold up the review on that and am OK with merging without it for now.

dylanegan: @nfagerlund I just went ahead and documented it, but yeah, it ain't entirely useful since you're just getting a name. A hint about not returning the private key text would be great. If you have a bit of text you would prefer I can tack that on separately.

PR #154, opened 2018-03-02
Update the installer instructions to make mounted disc the recommended Postgres option

KFishner:
Changes in this branch:
- Update the documentation to make the Postgres options clear. Primarily, remove the misunderstanding that Postgres is a dependency. Instead of making a separate "Data Storage" page, I updated installer instructions to make mounted disc the recommended Postgres option ([Asana](https://app.asana.com/0/525586892110393/552253409846418))
- Fixed the logging page to not be a directory (@davidcelis was there a reason you made it a directory?)

Educational changes with sales/SEs:
- Postgres is not a dependency. It actually was never called out as a dependency anywhere in the docs, this is just a tribal misinformation


content/source/docs/enterprise/private/logging.html.md
```diff
@@ -1,7 +1,7 @@
 ---
 layout: "enterprise2"
 page_title: "Terraform Enterprise Logging"
-sidebar_current: "docs-enterprise2-home"
+sidebar_current: "docs-enterprise2-private-installer"
```
nfagerlund: This should be `docs-enterprise2-logging`, right? (Man, I really want to get rid of this sidebar_current business.)

davidcelis:
> Fixed the logging page to not be a directory (@davidcelis was there a reason you made it a directory?)

Weird, I don't know why that happened. Definitely a mistake!

PR #155, opened 2018-03-02
Constrict supported versions further

evanphx:
For the time being, restrict the versions of the OSs we support to the most recent versions.

### DIFF
--- a/content/source/docs/enterprise/private/install-installer.html.md
+++ b/content/source/docs/enterprise/private/install-installer.html.md
@@ -42,12 +42,12 @@ The Private Terraform Enterprise Installer currently supports the following
 operating systems:
 
 * Debian 7.7+
-* Ubuntu 14.04 / 15.10 / 16.04
+* Ubuntu 14.04 / 16.04
 * Fedora 21 / 22
-* Red Hat Enterprise Linux 6.5+
-* CentOS 6+
-* Amazon Linux 2014.03 / 2014.09 / 2015.03 / 2015.09 / 2016.03 / 2016.09 / 2017.03
-* Oracle Linux 6.5+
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/private/install-installer.html.md
```diff
@@ -42,12 +42,12 @@ The Private Terraform Enterprise Installer currently supports the following
 operating systems:
 
 * Debian 7.7+
-* Ubuntu 14.04 / 15.10 / 16.04
+* Ubuntu 14.04 / 16.04
```
blalor: Why leave 14.04?

evanphx: My thinking was it's LTS and still in the support period. When 18.04 comes out in a few months, we'll drop support for 14.04.

content/source/docs/enterprise/private/install-installer.html.md
```diff
@@ -42,12 +42,12 @@ The Private Terraform Enterprise Installer currently supports the following
...
 * Fedora 21 / 22
-* Red Hat Enterprise Linux 6.5+
-* CentOS 6+
-* Amazon Linux 2014.03 / 2014.09 / 2015.03 / 2015.09 / 2016.03 / 2016.09 / 2017.03
-* Oracle Linux 6.5+
+* Red Hat Enterprise Linux 7.2+
```
blalor: Is there a reason not to say 7.4, or 7.4+

evanphx: We could, again I was trying to give a little bit of wiggle room. I know we hav 7.3 users and I don't know that we need to exclude them atm.

PR #157, opened 2018-03-05
A few broken links came up during build

dylanegan:
I went to deploy today and a few broken links appeared during the build process. I'm unsure if we can easily check these links locally as we make changes, but that would be neat.

Anyone else familiar enough with `Middleman` to know if we can do that?

I'm just opening this PR up for reference, but will merge and deploy immediately.


KFishner:
thanks for catching these @dylanegan !

PR #166, opened 2018-03-08
Add documentation for SAML SSO API token expiration

zhssh:


content/source/docs/enterprise/saml/login.html.md
```diff
@@ -9,3 +9,7 @@ sidebar_current: "docs-enterprise2-saml-login"
 Once SAML is configured users can visit `https://<YOUR_TERRAFORM_ENTERPRISE_DOMAIN>/session` to login.
 
 They can follow the link to complete the SAML login process with the identity provider. If the user is logging in for the first time, an account will be created from them in Terraform Enterprise. Their username will be autogenerated from their email address using the text before the `@`. The username will only contain alphanumeric characters, `-`, or `_`. All invalid characters will be converted to `_`.
+
+## API token expiration
+
+When a user's web session expires, their API tokens will be temporarily disabled until they reauthenticate at `https://<YOUR_TERRAFORM_ENTERPRISE_DOMAIN>/session`.
```
nfagerlund: Whoa, that's intense, given that these tokens are probably being used for something automated. As a reader, I have questions. 

- How often does a web session expire? Is it configurable? 
- Do we recommend that you only use team-based or org-based tokens for automation if SSO is enabled?

skierkowski: >How often does a web session expire? Is it configurable?

Sessions expiration periods are configurable in the organization settings.

> Do we recommend that you only use team-based or org-based tokens for automation if SSO is enabled?

Users should use team-based or org-based tokens for automation, but for personal use (e.g. from command line) users should use the user-tokens. This is true regardless if SSO is used or not.

> Also... just why? My guess is that it's because under SSO, a login attempt is TFE's only way to know whether a user has been deleted from the directory or not. But I think we should say as much here, to get readers on board with this being a reasonable limitation.

Yeah, your guess is exactly right. @zhssh I think it is worth calling out that this forces the user to go through SAML to refresh and thus update their permissions based on team-mapping. Probably should link to those docs too.

@zhssh it might be good to call out that this only impacts the user tokens and that the org and team-based tokens are not impacted.

PR #167, opened 2018-03-08
Add instructions about ATLAS_TOKEN to the upgrade docs

nfagerlund:


content/source/docs/enterprise/upgrade/index.html.md
```diff
@@ -68,23 +68,44 @@ terraform {
...
 ```bash
 terraform init
 ```
 
-## 4. Disable VCS integration in the legacy environment
+## 4. Set Variables in the New Workspace
+
+Open two side-by-side browser windows. In one of them, navigate to the new workspace you created and click on the "Variables" tab. In the other, navigate to the legacy environment you're migrating and view its variables.
+
+Copy the old environment's variables and values to the new workspace, ensuring that all of the values are identical.
```
skierkowski: Might be worth calling out that they'll need to re-enter the sensitive variables, as those are not readable from the old workspace.

content/source/docs/enterprise/upgrade/index.html.md
```diff
@@ -68,23 +68,44 @@ terraform {
...
+
+### 4.a. Set `ATLAS_TOKEN` (Optional)
+
+~> **Important:** This step is **only** for configurations that access state from a legacy TFE environment using a `terraform_remote_state` data source. If this environment doesn't do that, skip to the next step.
+
+TFE usually makes it easy to share state data between workspaces by automatically handling authentication. However, this only works _within_ a single organization, and your legacy environments are in a separate organization from your new workspaces. If your Terraform configuration accesses remote state from a legacy TFE environment, you must manually provide credentials until all of the relevant environments have been migrated to the new TFE.
+
+To do this, create a new environment variable in the new workspace called `ATLAS_TOKEN` and enter a valid TFE user token that can access the environment in question. (Make sure you check the "Sensitive" checkbox to protect the token.)
```
skierkowski: The generated token should actually have access to both organizations, the current (V2) and the previous (V1).

content/source/docs/enterprise/upgrade/index.html.md
nfagerlund: ah good call

content/source/docs/enterprise/upgrade/index.html.md
nfagerlund: I was wondering about that. What all does it need access to, and what level of access? Write perms in the specific workspace where it's being used? More? Less?

nfagerlund: read access to the affected v1 env(s) and write access to the v2 workspace. Because this is the atlas_token that gets used for the remote state operations for just performing the current tf run.

nfagerlund:
Got verbal 👍 from Maciej earlier today in our sync.

PR #170, opened 2018-03-08
Fix issues related to `?` and note that both filters are required if one is provided

davidcelis:
When URLs contain query params, most shells have a bad time. Let's quote 'em.

### DIFF
--- a/content/source/docs/enterprise/api/variables.html.md
+++ b/content/source/docs/enterprise/api/variables.html.md
@@ -103,8 +103,8 @@ curl \
 
 ### Parameters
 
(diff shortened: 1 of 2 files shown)


content/source/docs/enterprise/api/variables.html.md
```diff
@@ -103,16 +103,16 @@ curl \
 
 ### Parameters
 
-- `?filter[organization][name]` (`optional`) - Optionally filter the list to an organization given the organization name
-- `?filter[workspace][name]` (`optional`) - Optionally filter the list to a workspace given the workspace name
+- `?filter[organization][name]` (`optional`) - Optionally filter the list to an organization given the organization name. If this parameter is provided, `filter[workspace][name]` must also be provided.
```
dylanegan: You can remove the `?` here too.

PR #171, opened 2018-03-08
Fixed workspace examples to use name instead of id

rberlind:
Fixed workspace examples to use name instead of id when required.
Also documented Show workspace API

### DIFF
--- a/content/source/docs/enterprise/api/workspaces.html.md
+++ b/content/source/docs/enterprise/api/workspaces.html.md
@@ -213,7 +213,7 @@ Update the workspace settings
 
 ### Parameters
 
-Note that only the first four of these parameters can be used when updating a workspace without a VCS repository.
+Note that workspaces without an associated VCS repository only use the organization name, workspace name, terraform version, and working directory.
 
 - `:organization_name` (`string: <required>`) - Specifies the name of the organization the workspace should belong to. The organization must already exist in the system, and the user must have permissions to create new workspaces. This parameter is specified in the URL path.
 - `name` (`string: <required>`) - Specifies the name of the workspace to update, which can only include letters, numbers, `-`, and `_`. This will be used as an identifier and must be unique in the organization.
--- a/content/source/docs/enterprise/api/workspaces.html.md
+++ b/content/source/docs/enterprise/api/workspaces.html.md
@@ -213,10 +213,12 @@ Update the workspace settings
 
 ### Parameters
 
+Note that only the first four of these parameters can be used when updating a workspace without a VCS repository.
+
 - `:organization_name` (`string: <required>`) - Specifies the name of the organization the workspace should belong to. The organization must already exist in the system, and the user must have permissions to create new workspaces. This parameter is specified in the URL path.
 - `name` (`string: <required>`) - Specifies the name of the workspace to update, which can only include letters, numbers, `-`, and `_`. This will be used as an identifier and must be unique in the organization.
 - `terraform_version` (`string: <optional>`) - Specifices the version of Terraform to use for this workspace.
-- `working-directory` (`string: ''`) - Specifies the directory that Terraform will execute within. This defaults to the root of your repository and is typically set to a subdirectory matching the environment when multiple environments exist within the same repository.
+- `working-directory` (`string: ''`) - Specifies the directory that Terraform will execute within. This defaults to the root of your configuration and is typically set to a subdirectory matching the environment when multiple environments exist within the same configuration.
 - `vcs-repo.branch` (`string: ''`) - Specifies the repository branch that Terraform will execute from. If left null or as an empty string, this defaults to the repository's default branch (e.g. `master`) .
 - `vcs-repo.oauth-token-id` (`string: <optional>`) - Specifies the VCS Connection (OAuth Conection + Token) to use as identified. This ID can be obtained from the [oauth-tokens](./oauth-tokens.html) endpoint.
 - `vcs-repo.ingress-submodules` (`boolean: false`) - Specifies whether submodules should be fetched when cloning the VCS repository.
--- a/content/source/docs/enterprise/api/workspaces.html.md
+++ b/content/source/docs/enterprise/api/workspaces.html.md
@@ -97,7 +97,7 @@ $ curl \
       }
     },
     "links": {
-      "self": "/api/v2/organizations/my-organization/workspaces/ws-YnyXLq9fy38afEeb"
+      "self": "/api/v2/organizations/my-organization/workspaces/workspace-demo"
     }
   }
 }
@@ -196,7 +196,7 @@ $ curl \
       }
     },
     "links": {
-      "self": "/api/v2/organizations/my-organization/workspaces/ws-SihZTyXKfNXUWuUa"
+      "self": "/api/v2/organizations/my-organization/workspaces/workspace-demo"
     }
   }
 }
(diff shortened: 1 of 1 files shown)


nfagerlund:
OK, so, I tested this out and it looks good for the GET method! So that's great! But I can't get the PATCH method to work at all, either with the old or the new addressing? 

So, I guess I'm 👍 , since you're not making it MORE wrong, but I could definitely use some help verifying how the update API is supposed to work.

rberlind:
I did get the update workspace to work with PATCH.  I did this for workspaces with and without VCS repos attached.  It definitely only worked with name, not ID.

Here is the process I suggested a customer user who had asked about updating a workspace with VCS attached.

1. In a command shell, first export the API token of the owners team of your organization.  For instance, `export ATLAS_TOKEN=<token>`.
2. Then run `curl --header "Authorization: Bearer $ATLAS_TOKEN" --header "Content-Type: application/vnd.api+json" https://app.terraform.io/api/v2/organizations/<your_org>/workspaces/<your_workspace_name>` to see information about the workspace.  (If you have jq installed, pipe the results of that to jq (adding "| jq") to get nicer formatting.)
3. Copy the outputted JSON document to a file such as workspace.json and edit it to change the vcs settings.
4. Then run `curl --header "Authorization: Bearer $ATLAS_TOKEN" --header "Content-Type: application/vnd.api+json" --request PATCH --data @workspace.json https://app.terraform.io/api/v2/organizations/<your_org>/workspaces/<your_workspace_name>` to update the workspace.

I just tested again twice.  One time, I changed the VCS branch.  The second time, I changed the name of the workspace.

I was also able to test like this for a workspace without VCS, but had to change my process by deleting almost everything form the workspace.json file except the type and name.  So, effectively, I was only able to change the name. Here is what my stripped down workspace.json looked like:

```
{
  "data": {
    "type": "workspaces",
    "attributes": {
      "name": "workspace-from-api"
    }
  }
}
```
Note that as I stripped out content from workspace.json, I went from getting a JSON error message like {"errors":[{"status":"422","title":"unknown attribute","description":"unknown attribute 'locked' for Workspace."}]}  with HTTP 422 error to getting nothing, but adding -v to the curl command showed I was getting HTTP 400 errors.

I think the problem is that the only allowed fields for workspace without VCS are type and name.

content/source/docs/enterprise/api/workspaces.html.md
```diff
@@ -400,18 +399,90 @@ $ curl \
...
 | Method | Path           |
 | :----- | :------------- |
-| DELETE | /organizations/:organization/workspaces/:workspace |
+| DELETE | /organizations/:organization/workspaces/:name |
 
 ### Parameters
 
-- `:workspace` (`string: <required>`) - Specifies the workspace ID to delete. This parameter is specified in the URL path.
+- `:name` (`string: <required>`) - Specifies the name of the workspace to delete. This parameter is specified in the URL path.
 - `:organization` (`string: <required>`) - Specifies the organization name where the workspace belongs. This parameter is specified in the URL path.
```
armchairlinguist: I'd probably say "the name of the organization the workspace belongs to", but since this isn't actually changing, either way is fine.

content/source/docs/enterprise/api/workspaces.html.md
```diff
@@ -252,7 +251,7 @@ $ curl \
   --header "Content-Type: application/vnd.api+json" \
   --request PATCH \
   --data @payload.json \
-  https://app.terraform.io/api/v2/compound-workspaces/ws-erEAnPmgtm5mJr77
+  https://app.terraform.io/api/v2/organizations/my-organization/workspaces/my-workspace-2
```
armchairlinguist: Is there a reason this doesn't use the same demo name as above?

content/source/docs/enterprise/api/workspaces.html.md
```diff
@@ -400,18 +399,90 @@ $ curl \
...
+- `:organization` (`string: <required>`) - Specifies the organization name under which to list the workspaces. This is specified in the URL path.
+- `:name` (`string: <required>`) - Specifies the name of the workspace to show details for. This is specified in the URL path.
+
+### Sample Request
+
+```shell
+$ curl \
+  --header "Authorization: Bearer $ATLAS_TOKEN" \
+  --header "Content-Type: application/vnd.api+json" \
+  https://app.terraform.io/api/v2/organizations/my-organization/workspaces/my-workspace-3
```
armchairlinguist: I'd probably use the same workspace name here as well, unless there's a reason not to?

armchairlinguist:
@rberlind Did you happen to test whether the terraform version is settable on a non-VCS workspace? Seems like it should be, but I don't know if it is.

content/source/docs/enterprise/api/workspaces.html.md
rberlind: Agree on this and will do new commit with the change.

content/source/docs/enterprise/api/workspaces.html.md
rberlind: There are several workspace names used including my-workspace, my-workspace-1, my-workspace-2, and workspace-demo.  I did not want to make more changes than necessary.  I did make sure names were consistent within single examples.  Note that we need more than one name for the List Workspaces API.

content/source/docs/enterprise/api/workspaces.html.md
rberlind: Changing to my-workspace-1.

content/source/docs/enterprise/api/workspaces.html.md
rberlind: I suspect workspace-demo is valid as separate name because it is for workspace without VCS and should be distinct from others.

rberlind: I'm going to reduce to workspace-1 and workspace-2 with first for without VCS and second with.

rberlind:
@armchairlinguist :  I had not checked changing terraform-version for non-VCS workspace, but I just did test and was able to do it.  Note that when looking at workspace Settings tab in UI, I had to refresh the page to see.

I used workspace.json with

```
{
  "data":
  {
    "attributes": {
      "name":"workspace-from-api",
      "terraform-version":"0.11.1"
    },
    "type":"workspaces"
  }
}
```

Note, however, that I could not set terraform-version when creating the workspace without VCS.  It seems the only thing we can set when creating the workspace without VCS is the name.

armchairlinguist:
Okay, so the update workspace endpoint should probably make clear what is currently updatable for each type of workspace, then. Basically the first three params can be used without VCS, but the rest cannot.

rberlind:
I'll add a note on that. I actually just tested with workspace.json including working-directory for workspace without VCS.  And it did work.  But this is strange because the working directory is actually shows up in the UI on the Integrations tab as the Terraform Working Directory field under VCS settings. I'm not sure what real meaning there is in setting a working directory for a workspace without VCS. Perhaps, it means that you could upload a configuration for which your main.tf is not in the top-level directory?

content/source/docs/enterprise/api/workspaces.html.md
```diff
@@ -213,10 +213,12 @@ Update the workspace settings
 
 ### Parameters
 
+Note that only the first four of these parameters can be used when updating a workspace without a VCS repository.
```
armchairlinguist: I'd probably make this something like "Workspaces without an associated VCS repository only use the organization name, workspace name, terraform version, and working directory."

rberlind:
OK.  I made that change, @armchairlinguist

rberlind:
I changed, @armchairlinguist , but did not test.

armchairlinguist:
It's okay, I did 🎉 

@nfagerlund what do you think about trying to get this out soon? Lots of improvements, will make this page much more accurate.

nfagerlund:
Yeah let's do it!

rberlind:
Awesome!  Thanks.

PR #174, opened 2018-03-09
A few small fixes

nfagerlund:
- Update the navigation info in getting-started/workspaces, since we don't show the product switcher to people anymore. 
- Document supported versions of Bitbucket Server.


nfagerlund:
I went with "4.9.1 and newer" because we'd treat it as a bug if a 5.x release stopped working... right?

PR #182, opened 2018-03-13
Fix broken rhel link, fix incorrect sidebar_current highlighting

catsby:
Fixes broken link and loss of sidebar highlighting and expansion when on RHEL page


catsby:
Checked links locally 👍

PR #184, opened 2018-03-15
First cut of the installer beta logs info

evanphx:


rberlind:
So, I see that `docker logs <container>` could be used to retrieve logs from any of our containers.  But do we actually do anything with the logs other than storing them in the containers? Are we already pulling them from the containers to the instanced PTFE runs on? Maybe it would be helpful to show customers our actual docker logging configuration if anything other than standard.  It might also be good to list all the container names.

content/source/docs/enterprise/private/logging.html.md
```diff
@@ -38,6 +38,32 @@ tfe.mycompany.io-terraform-state-parser
...
+The container uses the docker API internally to find the containers and ingress the logs, at which point it then sends
+them `logs.mycompany.com` over TCP on port 55555 using syslog over TLS.
+
+~> **NOTE:** Private Terraform Enterprise installations only supports docker `log-driver` configured to either `json-file` or `journald`.
+   All other log drivers prevent the support bundle functionality from gathering logs, making it
+   impossible to provide product support. **DO NOT** change the log driver of an installation to anything other than `json-file` or `journald`.
+
 ---
 
 ## Audit Logs
```
KFishner: > Audit log entries are written to the application logs. To distinguish audit Log entries from other log entries, the JSON is prefixed with [Audit Log]. These are written out to CloudWatch logs just like all other application-level logs.

where are audit logs in the installer-based version?

evanphx:
@rberlind We don't save them onto the instance currently, no. They're only available as container logs.

content/source/docs/enterprise/private/logging.html.md
evanphx: They are mixed into the atlas logs. I'll specify the container that emits them.

content/source/docs/enterprise/private/logging.html.md
```diff
@@ -38,13 +38,45 @@ tfe.mycompany.io-terraform-state-parser
...
 
----
+## System-level Logs
+
+All other system-level logs can be found in the standard locations for an Ubuntu 16.04 system.
+
+# Application Logs - Installer Beta
+
+The installer-based version runs the application in a set of docker containers. As such, any tooling that can interact with docker logs
+can read the logs. This includes running the command `docker logs` as well as access the [Docker API](https://docs.docker.com/engine/api/v1.36/#operation/ContainerLogs).
```
blalor: > This includes running the command `docker logs`, as well as accessing the Docker API.

Massaged 👆 a little.

content/source/docs/enterprise/private/logging.html.md
```diff
@@ -38,13 +38,45 @@ tfe.mycompany.io-terraform-state-parser
...
+
+```shell
+$ docker run --name="logspout" \
+	--volume=/var/run/docker.sock:/var/run/docker.sock \
+	gliderlabs/logspout \
+	syslog+tls://logs.mycompany.com:55555
+```
+
+The container uses the docker API internally to find the containers and ingress the logs, at which point it then sends
+them `logs.mycompany.com` over TCP on port 55555 using syslog over TLS.
```
blalor: "… sends them _to_ `logs…` over TCP …"

content/source/docs/enterprise/private/logging.html.md
```diff
@@ -38,13 +38,45 @@ tfe.mycompany.io-terraform-state-parser
...
+
+~> **NOTE:** Private Terraform Enterprise installations only supports docker `log-driver` configured to either `json-file` or `journald`.
```
blalor: s/supports/support/

Also, this this sentence needs some context.  It's not clear that `log-driver` refers to the daemon-wide configuration, as no earlier reference is made to changing that config.

content/source/docs/enterprise/private/logging.html.md
```diff
@@ -38,13 +38,45 @@ tfe.mycompany.io-terraform-state-parser
...
+   All other log drivers prevent the support bundle functionality from gathering logs, making it
+   impossible to provide product support. **DO NOT** change the log driver of an installation to anything other than `json-file` or `journald`.
+
+# Audit Logs
 
-## Audit Logs
+Audit log entries are written to the application logs. To distinguish audit Log entries from other log entries, the JSON is prefixed with `[Audit Log]`.
```
blalor: "audit Log" → "audit log"

blalor: Also, I think it would be useful to show a couple of example lines.  Are we talking 
```
[Audit Log] { … some json … }
```
or
```
{ "msg": "[Audit Log] …", … }
```
?

evanphx: Ah, ok sure.

PR #190, opened 2018-03-21
TFE API for 2FA, introducing Account

dylanegan:
CHANGELOG: TFE API for 2FA, manage 2FA for your account

I made use of `_template.md`, really really nice.

A couple of notes;

1) I used `null` when value is optional, but default is empty
2) I think we should provide status codes somewhere as `two-factor-resend-verification-code` should return a 202 rather than a 200 and it'd be handy to know that

### DIFF
--- a/content/source/docs/enterprise/api/account.md
+++ b/content/source/docs/enterprise/api/account.md
@@ -1,0 +1,221 @@
+---
+layout: enterprise2
+page_title: "Account - API Docs - Terraform Enterprise"
+sidebar_current: "docs-enterprise2-api-account"
+---
+
+# Account API
+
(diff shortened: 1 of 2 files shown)


thrashr888:
Another note is that this page is missing the `/account/details` and password change actions.

Are we going to leave those out on purpose? If not, I suppose it's fine to add those separately.

dylanegan:
@thrashr888 they can probably be added in, but wanted to keep focus on 2FA here. I will open up an Asana task to get the others completed and do it separately.

It'd be great to figure out the status codes in this though because I feel like that'd be important for responses with no content.

PR #197, opened 2018-03-26
TFE API ConfigurationVersions spring cleaning

dylanegan:
CHANGELOG: Spring cleaning for TFE API ConfigurationVersion endpoint

While doing some work for the ConfigurationVersion API I noticed that the documentation was a bit lacking. This PR will be used to track ongoing updates.

### DIFF
--- a/content/source/docs/enterprise/api/configuration-versions.html.md
+++ b/content/source/docs/enterprise/api/configuration-versions.html.md
@@ -151,7 +151,7 @@ curl \
       "source": "tfe-api",
       "status": "pending",
       "status-timestamps": {},
-      "upload-url": "https://ARCHIVIST/v1/object/9224c6b3-2e14-4cd7-adff-ed484d7294c2"
+      "upload-url": "https://archivist.terraform.io/v1/object/9224c6b3-2e14-4cd7-adff-ed484d7294c2"
     },
     "relationships": {
       "ingress-attributes": {
@@ -171,7 +171,7 @@ curl \
 
 -> **Note**: Uploading a configuration file automatically creates a run and associates it with this configuration-version. Therefore it is unnecessary to [create a run on the workspace](./run.html#create-a-run) if a new file is uploaded.
 
-`POST https://ARCHIVIST/v1/object/4c44d964-eba7-4dd5-ad29-1ece7b99e8da`
+`POST https://archivist.terraform.io.terraform.io/v1/object/4c44d964-eba7-4dd5-ad29-1ece7b99e8da`
 
 **The URL is provided in the `upload-url` attribute in the `configuration-versions` resource.**
 
@@ -191,7 +191,7 @@ Properties without a default value are required.
 curl \
   --request PUT \
   -F 'data=@config.tar.gz' \
-  https://ARCHIVIST/v1/object/4c44d964-eba7-4dd5-ad29-1ece7b99e8da
+  https://archivist.terraform.io.terraform.io/v1/object/4c44d964-eba7-4dd5-ad29-1ece7b99e8da
 ```
 
 ## Available Related Resources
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/api/configuration-versions.html.md
```diff
@@ -60,19 +72,119 @@ curl \
 
 -> **Note**: Uploading a configuration file automatically creates a run and associates it with this configuration-version. Therefore it is unnecessary to [create a run on the workspace](./run.html#create-a-run) if a new file is uploaded.
 
-| Method | Path           |
-| :----- | :------------- |
-| POST | The upload URL is provided in the `upload-url` attribute in the `configuration-versions` resource |
+`POST https://archivist.terraform.io/v1/object/4c44d964-eba7-4dd5-ad29-1ece7b99e8da`
```
ryanuber: This would be a PUT for Archivist.

content/source/docs/enterprise/api/configuration-versions.html.md
```diff
@@ -60,19 +72,119 @@ curl \
...
+
+**The URL is provided in the `upload-url` attribute in the `configuration-versions` resource.**
+
+### Request Body
+
+This POST endpoint requires the following properties as a request payload.
```
ryanuber: `PUT`!

content/source/docs/enterprise/api/configuration-versions.html.md
```diff
@@ -60,19 +72,119 @@ curl \
...
+
+Properties without a default value are required.
+
+| Key path | Type | Default | Description                                                                                                                                                                                                         |
+| -------- | ---- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `data`   | file |         | A local .tar.gz file containing the folder of the terraform configuration files. This file can be created by running `tar -zcvf ./config.tar.gz .` from within the terraform configuration directory. |
```
ryanuber: The file should actually be the raw PUT body. There is no encoding / no form parsing, etc. on the server side, so this has to be raw bytes. To that end, the content-type also needs to be application/octet-stream (not sure if that is covered somewhere already).

content/source/docs/enterprise/api/configuration-versions.html.md
```diff
@@ -60,19 +72,119 @@ curl \
...
+
+### Sample Request
+
+```shell
+curl \
+  --request PUT \
+  -F 'data=@config.tar.gz' \
```
ryanuber: Recommend `--data-binary @config.tar.gz`. `-F` will send the payload as form data and fail.

dylanegan:
All good! Thanks filling in on the Archivist API. Would it make sense to separate this from the `ConfigurationVersions` API documentation? I'm unsure about how much Archivist does.

content/source/docs/enterprise/api/configuration-versions.html.md
dylanegan: It is not. I'll update accordingly.

ryanuber:
I think just specifying that you need to PUT raw bytes to the URL should be good. The type of thing you are uploading varies on a case-by-case basis (some places we use tar's, other places we have flat files or streams), so maybe it makes the most sense to keep this near the API docs which actually give you the URL in the first place. Archivist doesn't do much - it gives you an API to store and retrieve things, not much else (from the consumer side).

dylanegan:
@ryanuber great, thanks for the quick follow up! 💯

content/source/docs/enterprise/api/configuration-versions.html.md
nfagerlund: Also, we should decorate this to emphasize that it's not copy-pasteable from the docs! I suggest replacing the real object ID here with a placeholder. Like this: 

`PUT https://archivist.terraform.io/v1/object/<UNIQUE OBJECT ID>`

The different hostname for the archivist service might also be confusing for readers on PTFE, but I'm not sure? It's probably fine.

PR #207, opened 2018-03-28
provide more clarity on initial runs and plan-only runs

armchairlinguist:
Customers tend to be a bit confused sometimes by having to manually queue an initial run, and how plan-only runs behave, so I wanted to add a few more notes on that.


nfagerlund:
Oh! Apparently I was confused about that as well. Thank you for catching this.

PR #228, opened 2018-04-09
Sidebar: wrapping longer resource names

tombuildsstuff:
Currently resource names which are too long are truncated (technically hidden, I guess) by the main page body, as shown with `azurerm_sql_active_directory_administrator` here:

screen shot 2018-04-09 at 08 17 01

.. instead this PR attempts to fix this by wrapping the resource name onto multiple lines; as shown here:

screen shot 2018-04-09 at 08 16 47

as shown in the image above this PR isn't flawless (IMO the `n` in `azurerm_postgresql_configuration` above shouldn't be on a separate line) - but it helps make longer resource names readable/differentiable.

### DIFF
--- a/content/source/assets/stylesheets/_docs.scss
+++ b/content/source/assets/stylesheets/_docs.scss
@@ -20,6 +20,7 @@
       a {
         color: $sidebar-link-color;
         font-size: $sidebar-font-size;
+        overflow-wrap: break-word;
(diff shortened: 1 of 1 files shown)


paultyng:
@jescalan is this the proper way to fix this for now?

nfagerlund:
👍 I think "ugly" is way better than "busted."

PR #231, opened 2018-04-09
API template updates

nfagerlund:
- Add status codes and return values.
- Add a spot for related resources.
- Add example of `data[].x`.

For @caseylang: I've associated the related resources with each endpoint here. Sacrificing brevity for clarity explicitness. y/n?

For @dylanegan: How you feelin' about that format for the status/return values? Tried to keep it basic and terse, even though the nature of a `vars` object, for example, is a little opaque at the moment... still thinking about how to handle that.

### DIFF
--- a/content/source/docs/enterprise/api/_template.md
+++ b/content/source/docs/enterprise/api/_template.md
@@ -8,11 +8,6 @@ Follow this template to format each API method. There are usually multiple secti
 
 <!-- ^ The method and path are styled as a single code span, with global prefix (`/api/v2`) omitted and the method capitalized. -->
 
-- On success: HTTP 200 and a JSON API document (`type: "somethings"`).
-- On failure: HTTP 409 and a non-JSON error message.
-
-<!-- ^ Include status codes even if they're plain 200/404. If a JSON API document is returned, specify the `type`. -->
-
 Parameter            | Description
 ---------------------|------------
 `:organization_name` | The name of the organization to create the something in. The organization must already exist in the system, and the user must have permissions to create new somethings.
@@ -20,6 +15,13 @@ Parameter            | Description
 <!-- ^ The list of URL path parameters goes directly below the method and path, without a header of its own. They're simpler than other parameters because they're always strings and they're always mandatory, so this table only has two columns. Prefix URL path parameter names with a colon.
 
 If further explanation of this method is needed beyond its title, write it here, after the parameter list. -->
+
+Result  | Status and response
+--------|------------------------
+Success | HTTP 200 and a JSON API document (`type: "somethings"`).
+Failure | HTTP 409 and a non-JSON error message.
+
+<!-- ^ Include status codes even if they're plain 200/404. If a JSON API document is returned, specify the `type`. If there are multiple success or failure scenarios, put something descriptive in the "Result" column to distinguish them. -->
 
 ### Query Parameters
 
--- a/content/source/docs/enterprise/api/_template.md
+++ b/content/source/docs/enterprise/api/_template.md
@@ -41,6 +41,7 @@ Properties without a default value are required.
 Key path                    | Type   | Default | Description
 ----------------------------|--------|---------|------------
 `data.type`                 | string |         | Must be `"somethings"`.
+`data[].type`               | string |         | ... <!-- use data[].x when data is an array of objects. -->
 `data.attributes.category`  | string |         | Whether this is a blue or red something. Valid values are `"blue"` or `"red"`.
 `data.attributes.sensitive` | bool   | `false` | Whether the value is sensitive. If true then the something is written once and not visible thereafter.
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/api/_template.md
```diff
@@ -8,6 +8,11 @@ Follow this template to format each API method. There are usually multiple secti
 
 <!-- ^ The method and path are styled as a single code span, with global prefix (`/api/v2`) omitted and the method capitalized. -->
 
+- On success: HTTP 200 and a JSON API document (`type: "somethings"`).
```
dylanegan: In keeping with the other bits I think a table here is more concise and works better when we might have multiple success and failure scenarios each with their own status code. I usually just look up a status code in the API documentation if I'm receiving it from an endpoint.

content/source/docs/enterprise/api/_template.md
```diff
@@ -41,6 +46,7 @@ Properties without a default value are required.
 Key path                    | Type   | Default | Description
 ----------------------------|--------|---------|------------
 `data.type`                 | string |         | Must be `"somethings"`.
+`data[].type`               | string |         | ... <!-- use data[].x when data is an array of objects. -->
```
dylanegan: 👍

content/source/docs/enterprise/api/_template.md
```diff
@@ -61,6 +67,19 @@ them from unquoted values like booleans and nulls.
   list something like "(nothing)" as the default and explain in the description.
 -->
 
+### Available Related Resources
```
dylanegan: :+1:

content/source/docs/enterprise/api/_template.md
nfagerlund: hmmmmm in that case I'm inclined to move it down below the url parameter list. yeah?

dylanegan: `Status` and `Response` would tie it in to the terminology we currently use (`Sample Response`).

nfagerlund: 👍🏼

content/source/docs/enterprise/api/_template.md
```diff
@@ -21,6 +16,13 @@ Parameter            | Description
 
 If further explanation of this method is needed beyond its title, write it here, after the parameter list. -->
 
+Result  | Status and response
+--------|------------------------
+Success | HTTP 200 and a JSON API document (`type: "somethings"`).
```
dylanegan: Oh I meant `Status` and `Response`. 😜 

Status | Response
------ | ----------
`200` | JSON API document (`type: "somethings"`)
`204` | No content
`400` | JSON API document

Maybe the codes can link to official documentation too?

dylanegan: https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Busername%7D/%7Brepo_slug%7D/commit/%7Bnode%7D/statuses/build

Here you can see multiple status codes in place.

nfagerlund: @dylanegan augh, sorry! 😭 Wellllll it's merged, so, do you want to do a second PR with your proposal? Actually, maybe using real info from an endpoint would be more illustrative.

dylanegan: @nfagerlund what do you mean by real info? Like a response? It might be good, but then we'd drop the `Sample Response` part in favour of this? And have it as `Sample Responses`?

nfagerlund: @dylanegan oh, no, I meant, like, the real status codes from one of our endpoints. Like, whichever one you said was using 409?

dylanegan: @nfagerlund oh yeah, heh.

`POST /organizations/:organization_name/teams`

Status | Response | Reason(s)
---- | ------------ | -------
[200](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200) | [JSON API document](https://www.terraform.io/docs/enterprise/api/index.html#json-api-documents) | Successfully created a team
[400](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) | [JSON API error object](http://jsonapi.org/format/#error-objects) | Invalid `include` parameter
[404](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404) | [JSON API error object](http://jsonapi.org/format/#error-objects) | Organization not found, user is unauthorized to perform action
[422](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) | [JSON API error object](http://jsonapi.org/format/#error-objects) | Validation errors
[500](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500) | [JSON API error object](http://jsonapi.org/format/#error-objects) | Failure during team creation

PR #236, opened 2018-04-12
Correct link to localhost in custom provider docs

charlieegan3:


nfagerlund:
Ah, thanks for catching that.

PR #240, opened 2018-04-16
Reuse existing Gemfile+config for provider site

radeksimko:
This allows us to reduce the amount of plumbing code we need to put into each provider codebase as discussed in https://github.com/terraform-providers/terraform-provider-aws/pull/4122 and with @bflad via Slack.

### DIFF
--- a/content/config.rb
+++ b/content/config.rb
@@ -11,3 +11,10 @@ config[:file_watcher_ignore] += [/^(\/website\/)?ext\//]
 
 require "middleman_helpers"
 helpers Helpers
+
+if ENV.include?('PROVIDER_NAME')
+  provider = ENV['PROVIDER_NAME']
+  logger.info("==")
+  logger.info("==> See #{provider} docs at http://localhost:4567/docs/providers/#{provider}")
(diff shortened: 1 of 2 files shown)


content/config.rb
```diff
@@ -11,3 +11,10 @@
 
 require "middleman_helpers"
 helpers Helpers
+
+if ENV.include?('PROVIDER_NAME')
+  provider = ENV['PROVIDER_NAME']
+  logger.info "=="
```
appilon: nitpick: my ruby is rusty but I've always disliked the omission of brackets in function calls. I recall it being a very divisive style thing though so whichever everyone prefers is fine with me.

appilon (approved):

content/config.rb
catsby: I too find the omission of parenthesis jarring but I'm not going to force the issue ;)

catsby (approved):
👍 I bite my thumb at your lack of parenthesis in function calls 😆

radeksimko:
Parenthesis added :)

PR #241, opened 2018-04-16
Improve status/response table in api template

nfagerlund:
This is from @dylanegan's improvements in the comments of https://github.com/hashicorp/terraform-website/pull/231.


armchairlinguist:
This looks great. We will be at the top of the class in API docs! 😸

dylanegan:
@nfagerlund can you update a page here so I can Netlify view it and share it with people for review?

nfagerlund:
@dylanegan sure thing! I went with create a team, since I think that's what your example was from.

nfagerlund:
Hmmm.... yes, we should. 👍🏼

PR #242, opened 2018-04-16
Add default description to make for better unfurls

phinze:
My latest discovery in the Saga of Reasonable Unfurls is that
descriptions are crucial to getting a non-huge image printed.

Here, we set a default description to make it so that every page has at
least something there. Ideally we should be writing descriptions for all
our pages, but this will help as a stopgap unfurl improvement in the
meantime.


phinze:
It works!

PR #251, opened 2018-04-18
remove note about uneditable repo connections when setting up workspaces

armchairlinguist:
We can edit workspace VCS now, so this note is no longer needed. 🎉


nfagerlund:
Ah good catch, I missed this one!

PR #254, opened 2018-04-20
Updated pTFE support bundler.

jboero:
Customers don't always have hashicorp-support available.  Replicated has its own bundler.


armchairlinguist:
We usually recommend that Installer customers use the Replicated GUI to generate bundles - this would be a fallback if the GUI isn't running, only. Some customers also don't know what replicated means, so we should refer to an "Installer-based instance" or similar.

nfagerlund:
This seems to have been moved out of the FAQ and [into the Diagnostics page](https://www.terraform.io/docs/enterprise/private/diagnostics.html#installer-based-instances), so I'm closing the PR.

PR #258, opened 2018-04-24
Provide redirects for aws_codecommit_* resource pages that were fixed for naming consistency

bflad:
Reference: https://github.com/terraform-providers/terraform-provider-aws/pull/4330

The AWS provider change will go out tomorrow, 4/25, with the v1.16.0 release.


bflad:
Self-merging so its available when I release v1.16.0 of the AWS provider in the next few minutes.

PR #260, opened 2018-04-25
Exempt the preflight releases.hashicorp.com check from the stern warning about supportability

armchairlinguist:
Per discussion, this check doesn't currently use the proxy, so systems running behind a proxy may get a misleading failure. We want them to know this is probably okay.


armchairlinguist:
@evanphx Mind taking a look and making sure I'm not letting people off the hook too easily here?

PR #266, opened 2018-05-01
make: Expose provider slug as variable

radeksimko:
This is to address outliers like Atlas/TFE and DigitalOcean which have different slug & path.

https://github.com/terraform-providers/terraform-provider-atlas/pull/4
https://github.com/terraform-providers/terraform-provider-digitalocean/pull/86

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -274,8 +274,8 @@
 	path = ext/providers/vsphere
 	url = https://github.com/terraform-providers/terraform-provider-vsphere
 	branch = stable-website
-[submodule "ext/providers/do"]
-	path = ext/providers/do
+[submodule "ext/providers/digitalocean"]
+	path = ext/providers/digitalocean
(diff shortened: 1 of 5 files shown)


appilon (approved):

appilon (approved):

catsby (changes requested):
To be 💯 clear:

- Current `terraform.io/docs/providers/do/index.html`
- Remains `terraform.io/docs/providers/do/index.html`

[ ] correct?

radeksimko:
@catsby Correct, the URL path is defined by the directory name in `content/source/docs/providers` which isn't changing, the symlink will just point to a differently named submodule.

FYI: https://travis-ci.org/terraform-providers/terraform-provider-digitalocean/builds/373398993 is green (uses the branch from this PR and does link-checking via `do`)

catsby (approved):
👍

PR #269, opened 2018-05-02
scripts: Allow rejecting paths in link checker

radeksimko:
This is to unblock https://github.com/hashicorp/terraform/pull/17982 and allow exclusion of paths when checking links.

### DIFF
--- a/content/scripts/check-links.sh
+++ b/content/scripts/check-links.sh
@@ -3,10 +3,16 @@ set -e
 
 URL=$1
 ACCEPT_PATH=$2
+REJECT_PATH=$3
+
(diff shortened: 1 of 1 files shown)


content/scripts/check-links.sh
```diff
@@ -3,10 +3,16 @@ set -e
 
 URL=$1
 ACCEPT_PATH=$2
+REJECT_PATH=$3
+
+if [ "$REJECT_PATH" != "" ]; then
```
appilon: my bash isn't as lit 🔥as yours but I usually did non-empty check like this `if [ -n "$myvar" ]` I assume this works too.

appilon (approved):

PR #273, opened 2018-05-07
Minor improvements from proofreading

neodon:
I noticed a few minor issues while reading some of the documentation and figured I would make a small contribution to help fix them up.  I tried to keep the fixes as objective as possible.

I appreciate the thorough documentation!


nfagerlund:
Thank you for catching these!

PR #274, opened 2018-05-07
TFE API rate limiting

dylanegan:
CHANGELOG: Document the TFE API rate limiting

I've gone with the basic of the basics here. It'd be great to see this fleshed out, but I feel like our rate limiting implementation is fairly basic. We don't return much information and we don't vary limits across actions, it's just a simple blanketing.

### DIFF
--- a/content/source/docs/enterprise/api/index.html.md
+++ b/content/source/docs/enterprise/api/index.html.md
@@ -163,7 +163,7 @@ $ curl \
 
 ## Rate Limiting
 
-You can make up to 30 requests per second to the API as an authenticated or unauthenticated request. If you reach the rate limit then you will be throttled and an error response will be returned.
+You can make up to 30 requests per second to the API as an authenticated or unauthenticated request. If you reach the rate limit then your access will be throttled and an error response will be returned.
 
 Authenticated requests are allocated to the user associated with the authentication token. This means that a user with multiple tokens will still be limited to 30 requests per second, additional tokens will not allow you to increase the requests per second permitted.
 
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/api/index.html.md
```diff
@@ -161,6 +161,33 @@ $ curl \
 }
 ```
 
+## Rate Limiting
+
+You can make up to 30 requests per second to the API as an authenticated or unauthenticated request. If you reach the rate limit then you will be throttled and an error response will be returned.
```
nfagerlund: Throttling a person means strangling them!! 😬 Maybe let's change that to "your access will be throttled". 😂

content/source/docs/enterprise/api/index.html.md
```diff
@@ -161,6 +161,33 @@ $ curl \
...
+
+Authenticated requests are allocated to the user associated with the authentication token. This means that a user with multiple tokens will still be limited to 30 requests per second, additional tokens will not allow you to increase the requests per second permitted.
+
+Unauthenticated requests are associated with the requesting IP address.
```
nfagerlund: I had no idea we allowed unauthenticated requests! What endpoints can an anon hit?

dylanegan: @nfagerlund if you just hit an endpoint unauthenticated it still goes through to the application so all this will do is throttle those requests based on IP. I don't think we specifically have endpoints that can be accessed unauthenticated.

PR #276, opened 2018-05-07
Configurable link check delay for provider sites

appilon:
add startup delay to provider site linkchecks, my local tests showed 15 seconds was enough to not receive any `No data received` or `Connection reset by peer` messages from wget. Open to having the environment variable be named something friendlier

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ endif
 ifeq ($(PROVIDER_SLUG),)
 	$(eval PROVIDER_SLUG := $(PROVIDER_NAME))
 endif
-ifeq ($(LINKCHECK_DELAY,)
+ifeq ($(LINKCHECK_DELAY),)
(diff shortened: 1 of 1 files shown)


paddycarver:
🚲 🏚: MIDDLEMAN_STARTUP_DELAY sounds (to me) like the delay before starting middleman. LINKCHECK_DELAY? MIDDLEMAN_GRACE_PERIOD?

appilon:
@paddycarver I agree now that you mention it, `LINKCHECK_DELAY` seems like the most valid name, if the website were powered by something else the variable name would still be valid, I will change switch to that however I'm going to try and configure `wget` with some "saner" defaults so maybe this can be a zero config fix for all providers (if I have no such luck then this PR will stand)

catsby (approved):
👍

appilon:
Closed in favour of wget flag (already merged)

PR #277, opened 2018-05-08
Configure wget to back-off up to 2 minutes before failing

appilon:
This is an alternative to solving the failing CI, you will still see early `Connection reset by peer` or `No data received` messages but the max retry interval should now be more than enough for the website to startup. The advantage being there is zero config across providers as long as this number is large enough and it will not degrade performance for small providers that don't need a lot of time for middleman to startup.


catsby (approved):
👍

PR #278, opened 2018-05-08
Add minio setup guide

blalor:


content/source/docs/enterprise/private/minio-setup-guide.html.md
```diff
@@ -0,0 +1,95 @@
...
+sidebar_current: "docs-enterprise2-private-installer-minio"
+description: |-
+  This document provides an overview for setting up minio for external object storage for HashiCop Private Terraform Enterprise (PTFE).
+---
+
+## Introduction
+
+This document provides an overview for setting up minio for external object storage for HashiCop Private Terraform Enterprise (PTFE).
+
+When configured to use external services, PTFE must be connected to a storage service to persist workspace state and other file-based data.  Native support exists for Azure Blob Storage, Amazon S3, and services that are API-compatible with Amazon S3.  If you are not using Azure or a cloud provider with an S3-compatible service, or you are running PTFE in an environment without a storage service, it may be possible to use [minio](https://minio.io) instead.  Minio is an Amazon S3-compatible object storage server that can use mounted filesystems or a variety of other cloud storage solutions to persist data.  HashiCorp may add support in the future for other cloud storage solutions to meet customer requests.
```
armchairlinguist: I'd suggest we don't add promises of future support to the docs - it leads to lots of "but when?!" tickets.

content/source/docs/enterprise/private/minio-setup-guide.html.md
```diff
@@ -0,0 +1,95 @@
...
+
+### PTFE installation
+
+You may now [continue the installation in the browser](./install-installer.html#continue-installation-in-browser).
+
+1. provide a valid hostname
+2. choose the "Production" installation type
+3. choose the "External Services" production type
+4. provide the required Database URL for the PostgreSQL configuration
+5. choose "S3" for object storage
```
armchairlinguist: This looks like it might be duplicating other info. Is it reasonable to refer to the info and then instruct the user on where to drop the Minio-specific process in?

content/source/docs/enterprise/private/minio-setup-guide.html.md
blalor: ok, that's fair.

content/source/docs/enterprise/private/minio-setup-guide.html.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+### Start minio
+
+Now you'll start the minio container, mounting a volume so that you can gain access to the generated config:
+
+    docker run \
+      -d \
+      --name minio \
+      -v /run/minio/config:/root/.minio minio/minio:latest \
```
evanphx: If you break out the rest of the arguments pre line, you should break out the image too.

blalor:
All feedback items have been addressed.  Can you please re-review @evanphx ?

PR #282, opened 2018-05-09
Update and clarify PATCH behavior for workspaces api

nfagerlund:
- Set defaults to "(previous value)"
- Add top-level line for the vcs-repo object, with type "object or null".


nfagerlund:
btw this is definitely a time to use the "rendered" diff view in github.

PR #288, opened 2018-05-14
make: Wait until website is up before checking it

radeksimko:
Docker's `HEALTHCHECK` is not suitable for this purpose, because `docker run` won't wait until the the healthcheck passes. Neither is `docker wait` as it only works with container running/stopped state, not healthiness.


appilon (approved):
nice!

PR #291, opened 2018-05-22
Layouts: Make the mobile primary nav match the desktop one

nfagerlund:
This commit adds the missing "Guides" link to the mobile main nav, and removes
"Press Kit" and "Security" (which are linked at the bottom of the page on
both mobile and desktop widths).


nfagerlund:
What the heck is goin' on with that Netlify preview. 🤔

PR #293, opened 2018-05-23
PTFE AdminSettings API Documentation

caseylang:
Documentation for the new PTFE AdminSettings API. I broke the commits out into 4 slightly more manageable chunks of SMTP, SAML, General, and Twilio.

### DIFF
--- a/content/source/docs/enterprise/api/admin/settings.html.md
+++ b/content/source/docs/enterprise/api/admin/settings.html.md
@@ -10,6 +10,8 @@ sidebar_current: "docs-enterprise2-api-admin-settings"
 
 -> **Note**: These endpoints are only available in Private Terraform instances and only accessible by site administrators.
 
+-> **Pre-release**: These API endpoints are not yet available in the current Private Terraform Enterprise release.
+
 ## List SAML Settings
 `GET /api/v2/admin/saml-settings`
 


content/source/docs/enterprise/api/ptfe_settings.html.md
```diff
@@ -0,0 +1,491 @@
...
+
+```json
+{
+  "data": {
+    "id": "general-settings",
+    "type": "settings",
+    "attributes": {
+      "support_email_address": "support@example.com",
+      "security_email_address": "security@example.com",
+      "limit_user_org_creation": true
```
alisdair: I think JSON API attribute names should be `dasher-ized`, not `under_scored`.

content/source/docs/enterprise/api/ptfe_settings.html.md
```diff
@@ -0,0 +1,491 @@
...
+-> **Note**: These API endpoints are in beta and are subject to change. Additionally, these endpoints are only available in Private Terraform instances and only accessible by site administrators.
+
+## List General Settings
+`GET /api/v2/admin/settings/general`
+
+-> **Note:** This endpoint cannot be accessed with [organization tokens](../users-teams-organizations/service-accounts.html#organization-service-accounts). You must access it with a [user token](../users-teams-organizations/users.html#api-tokens) which has the `site-admins` permission.
+
+Status  | Response                                         | Reason
+--------|--------------------------------------------------|-------
+[200][] | Array of [JSON API document][]s (`type: "settings"`) | Successfully listed General settings
```
alisdair: From the examples below, I don't think this is accurate—it doesn't look like an array, just a single document? That would make sense, since we're semantically fetching a single record.

(Same comment throughout)

content/source/docs/enterprise/api/ptfe_settings.html.md
```diff
@@ -0,0 +1,491 @@
...
+`data.attributes.slo_target` | string |         | Single Log Out URL specifies the HTTP(s) endpoint on your IdP for single logout requests. This value is provided by your IdP configuration.
+`data.attributes.sso_target` | string |         | Single Sign On URL specifies the HTTP(S) endpoint on your IdP for single sign-on requests. This value is provided by your IdP configuration.
+`data.attributes.attr_groups`| string | `"MemberOf"` | Team Attribute Name specifies the name of the SAML attribute that determines team membership. The value of this attribute in the SAML assertion must be a string containing a comma-separated list of team names.
+`data.attributes.attr_site_admin`| string |`"SiteAdmin"`| Specifies the role for site admin access, overridding the "Site Admin Role" method.
+`data.attributes.site_admin_role`| string |`"site-admins"`| Specifies the role for site admin access, provided in the list of roles sent in the Team Attribute Name attribute.
+`data.attributes.sso_api_token_session_timeout`| integer | 1209600 | Specifies the Single Sign On session timeout, defaulting to 14 days.
```
alisdair: Are there maximum/minimum values for this attribute? What happens if it's elided? Can session timeout be disabled?

content/source/docs/enterprise/api/ptfe_settings.html.md
```diff
@@ -0,0 +1,491 @@
...
+This PATCH endpoint requires a JSON object with the following properties as a request payload.
+
+Properties without a default value are required if `data.attributes.enabled` is set to `true`.
+
+Key path                    | Type   | Default | Description
+----------------------------|--------|---------|------------
+`data.attributes.enabled`    | bool   | `false` | Allows SAML to be used. If true, all remaining attributes are required.
+`data.attributes.idp_cert`   | string |         | Identity Provider Certificate specifies the PEM encoded X.509 Certificate as provided by the IdP configuration.
+`data.attributes.slo_target` | string |         | Single Log Out URL specifies the HTTP(s) endpoint on your IdP for single logout requests. This value is provided by your IdP configuration.
+`data.attributes.sso_target` | string |         | Single Sign On URL specifies the HTTP(S) endpoint on your IdP for single sign-on requests. This value is provided by your IdP configuration.
```
alisdair: Typo: `HTTP(S)` here and `HTTP(s)` the line above.

(Do we really support HTTP for SSO URLs?)

content/source/docs/enterprise/api/ptfe_settings.html.md
```diff
@@ -0,0 +1,491 @@
...
+
+### Request Body
+
+This PATCH endpoint requires a JSON object with the following properties as a request payload.
+
+Properties without a default value are required if `data.attributes.enabled` is set to `true`.
+
+Key path                    | Type   | Default | Description
+----------------------------|--------|---------|------------
+`data.attributes.enabled`   | bool   | `false` | Allows SMTP to be used. If true, all remaining attributes are required.
```
alisdair: …except username & password if auth is `"none"`, I guess?

content/source/docs/enterprise/api/ptfe_settings.html.md
```diff
@@ -0,0 +1,491 @@
...
+This PATCH endpoint requires a JSON object with the following properties as a request payload.
+
+Properties without a default value are required if `data.attributes.enabled` is set to `true`.
+
+Key path                    | Type   | Default | Description
+----------------------------|--------|---------|------------
+`data.attributes.enabled`    | bool   | `false` | Allows Twilio to be used. If true, all remaining attributes are required.
+`data.attributes.account_sid`  | string   |  | The Twilio account id.
+`data.attributes.auth_token` | string   |  | The Twilio authentication token.
+`data.attributes.from_number`| string     |  | The phone number Twilio will use to send the message.
```
alisdair: It may be worth noting that this number has to be a registered Twilio number.

content/source/docs/enterprise/api/ptfe_settings.html.md
caseylang: Copy pasta error from template. Good catch, will resolve!

content/source/docs/enterprise/api/ptfe_settings.html.md
caseylang: True, which I call out in the username and password descriptions. Should I be extra explicit here as well?

content/source/docs/enterprise/api/ptfe_settings.html.md
caseylang: Will fix!
I'm not sure on the second point. Maybe @zhssh knows?

content/source/docs/enterprise/api/ptfe_settings.html.md
caseylang: I'm unsure. There appears to be no current validation on this value. @zhssh are you aware of a min/max for session timeout and if it can be nil?

caseylang:
@alisdair ready for re-review when you have a moment. Thanks for the first pass

content/source/docs/enterprise/api/ptfe_settings.html.md
```diff
@@ -0,0 +1,489 @@
...
+  https://app.terraform.io/api/v2/admin/settings/general
+```
+
+### Sample Response
+
+```json
+{
+  "data": {
+    "id": "general-settings",
+    "type": "settings",
```
alisdair: Based on a quick internal spike, the implementation is going to be easier with a couple of changes here:

- Let's make the `type` for each setting follow the pattern `settings/general`, `settings/smtp`, etc. This matches the path and the model name in our UI;
- Then for reduced redundancy, we can remove `-settings` from the `id`, i.e. `general`, `smtp`. This is for reduced redundancy.

content/source/docs/enterprise/api/ptfe_settings.html.md
```diff
@@ -0,0 +1,489 @@
...
+    "attributes": {
+      "support-email-address": "support@example.com",
+      "security-email-address": "security@example.com",
+      "limit-user-org-creation": true
```
alisdair: Could we use the full expansion of `org` here, i.e. `limit-user-organization-creation`?

content/source/docs/enterprise/api/ptfe_settings.html.md
alisdair: No, I think it's clear enough as-is!

content/source/docs/enterprise/api/ptfe_settings.html.md
caseylang: Certainly! Will fix.

content/source/docs/enterprise/api/ptfe_settings.html.md
```diff
@@ -0,0 +1,491 @@
+---
+layout: enterprise2
+page_title: "Private Terraform Settings - API Docs - Terraform Enterprise"
+sidebar_current: "docs-enterprise2-api-private-terraform-settings"
```
nfagerlund: When file moves, slug should change to `docs-enterprise2-api-admin-settings` so it lights up in the appropriate nav sidebar section.

content/source/docs/enterprise/api/ptfe_settings.html.md
```diff
@@ -0,0 +1,491 @@
...
+---
+
+# Private Terraform Settings API
```
nfagerlund: "Private Terraform Enterprise Settings API". (Likewise in `page_title` above.)

content/source/docs/enterprise/api/ptfe_settings.html.md
```diff
@@ -0,0 +1,491 @@
...
+
+-> **Note**: These API endpoints are in beta and are subject to change. Additionally, these endpoints are only available in Private Terraform instances and only accessible by site administrators.
```
nfagerlund: If you split this into two separate notes, it'll be easier to scan. (Since the "beta" note appears on every page, it becomes invisible after a few minutes. Moving the "private" note out of its shadow will make it stand out.)

content/source/docs/enterprise/api/ptfe_settings.html.md
```diff
@@ -0,0 +1,491 @@
+---
```
nfagerlund: This is part of the PTFE /admin APIs, which @davidcelis has mostly documented in a PR in the private copy of the repo. Since they're all in the same family, we should adjust this page so it lines up with the other ones! That means: 

- It should go in a subdirectory of `docs/enterprise/api`, at `docs/enterprise/api/admin`.
- Its filename should be `settings.html.md`.
- Whenever it links to other pages with a relative path, there should be an additional `../` at the front.
- It should get a nav sidebar link (source/content/layouts/enterprise2.erb) in that new admin subsection of the API links. See David's PR for example; this'll make a merge conflict once we merge both, but meh, that's easy to deal with.

content/source/docs/enterprise/api/ptfe_settings.html.md
```diff
@@ -0,0 +1,491 @@
...
+`data.attributes.slo-target` | string |         | Single Log Out URL specifies the HTTPS endpoint on your IdP for single logout requests. This value is provided by your IdP configuration.
+`data.attributes.sso-target` | string |         | Single Sign On URL specifies the HTTPS endpoint on your IdP for single sign-on requests. This value is provided by your IdP configuration.
+`data.attributes.attr-groups`| string | `"MemberOf"` | Team Attribute Name specifies the name of the SAML attribute that determines team membership. The value of this attribute in the SAML assertion must be a string containing a comma-separated list of team names.
+`data.attributes.attr-site-admin`| string |`"SiteAdmin"`| Specifies the role for site admin access, overriding the "Site Admin Role" method.
+`data.attributes.site-admin-role`| string |`"site-admins"`| Specifies the role for site admin access, provided in the list of roles sent in the Team Attribute Name attribute.
+`data.attributes.sso-api-token-session-timeout`| integer | 1209600 | Specifies the Single Sign On session timeout, defaulting to 14 days.
```
nfagerlund: Should specify the unit of that integer, in the description! ...I just did the math and it's seconds, so: "Specifies the Single Sign On session timeout in seconds, defaulting to..."

content/source/docs/enterprise/api/ptfe_settings.html.md
```diff
@@ -0,0 +1,491 @@
...
+[404]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404
+[422]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422
+[JSON API document]: https://www.terraform.io/docs/enterprise/api/index.html#json-api-documents
+[JSON API error object]: http://jsonapi.org/format/#error-objects
+
+### Request Body
+
+This PATCH endpoint requires a JSON object with the following properties as a request payload.
+
+Properties without a default value are required if `data.attributes.enabled` is set to `true`.
```
nfagerlund: If they already had a value, do they retain the previous value if you omit them while updating some other SAML settings? If so, I can come up with some way to describe that.

content/source/docs/enterprise/api/ptfe_settings.html.md
```diff
@@ -0,0 +1,491 @@
...
+      "username": "sample_user"
+    }
+  }
+}
+```
+
+## Update SMTP Settings
+`PATCH /api/v2/admin/settings/smtp`
+
+-> **Note:** This endpoint cannot be accessed with [organization tokens](../users-teams-organizations/service-accounts.html#organization-service-accounts). You must access it with a [user token](../users-teams-organizations/users.html#api-tokens) which has the `site-admins` permission.
```
nfagerlund: FYI: in David's PR, he omitted these notes and instead called out the auth requirements once, on the index page. @armchairlinguist, you have more direct contact with our readers, and you've probably seen confusion about token types firsthand; do you reckon we should go back through the other admin pages later and make sure every endpoint has this note?

content/source/docs/enterprise/api/ptfe_settings.html.md
caseylang: @nfagerlund They do retain their value. I would love words to clarify that

content/source/docs/enterprise/api/admin/settings.html.md
```diff
@@ -0,0 +1,492 @@
...
+
+This PATCH endpoint requires a JSON object with the following properties as a request payload.
+
+Properties without a default value are required if `data.attributes.enabled` is set to `true`.
+
+Key path                    | Type   | Default | Description
+----------------------------|--------|---------|------------
+`data.attributes.enabled`    | bool   | `false` | Allows SAML to be used. If true, all remaining attributes are required.
+`data.attributes.idp-cert`   | string |         | Identity Provider Certificate specifies the PEM encoded X.509 Certificate as provided by the IdP configuration.
+`data.attributes.slo-target` | string |         | Single Log Out URL specifies the HTTPS endpoint on your IdP for single logout requests. This value is provided by your IdP configuration.
```
armchairlinguist: I'd s/your/the/ for both of the yours in this line and the next line.

content/source/docs/enterprise/api/admin/settings.html.md
```diff
@@ -0,0 +1,492 @@
...
+`data.attributes.sso-target` | string |         | Single Sign On URL specifies the HTTPS endpoint on your IdP for single sign-on requests. This value is provided by your IdP configuration.
+`data.attributes.attr-groups`| string | `"MemberOf"` | Team Attribute Name specifies the name of the SAML attribute that determines team membership. The value of this attribute in the SAML assertion must be a string containing a comma-separated list of team names.
```
armchairlinguist: I don't think the second sentence is correct. The value can also list each team name in a separate item. Also it doesn't seem relevant here, so maybe just remove it or link to the documentation on it?

content/source/docs/enterprise/api/admin/settings.html.md
```diff
@@ -0,0 +1,492 @@
...
+`data.attributes.attr-site-admin`| string |`"SiteAdmin"`| Specifies the role for site admin access, overriding the "Site Admin Role" method.
+`data.attributes.site-admin-role`| string |`"site-admins"`| Specifies the role for site admin access, provided in the list of roles sent in the Team Attribute Name attribute.
```
armchairlinguist: I find these two lines confusing. If one overrides the other, are both still required? Can you clarify the behavior a bit?

content/source/docs/enterprise/api/admin/settings.html.md
```diff
@@ -0,0 +1,492 @@
...
+
+## Update SMTP Settings
+`PATCH /api/v2/admin/settings/smtp`
+
+-> **Note:** This endpoint cannot be accessed with [organization tokens](../users-teams-organizations/service-accounts.html#organization-service-accounts). You must access it with a [user token](../users-teams-organizations/users.html#api-tokens) which has the `site-admins` permission.
+
+Status  | Response                                     | Reason
+--------|----------------------------------------------|----------
+[200][] | [JSON API document][] (`type: "settings/smtp"`) | Successfully updated the SMTP settings
+[401][] | [JSON API error object][]                    | Failure during test message delivery, SMTP user credentials are invalid
```
armchairlinguist: The "failure during test message delivery" here confuses me. Are we validating the creds by sending a test message? If so, that should be included in the description of the endpoint, and then I think the error here could be abbreviated to `SMTP user credentials are invalid` and likewise for the below codes that mention the test message.

Something like:

"When this request is submitted, (what happens with the test message here). If the test message delivery fails, the API will return an error code indicating the reason for the failure."

content/source/docs/enterprise/api/admin/settings.html.md
```diff
@@ -0,0 +1,492 @@
...
+
+Properties without a default value are required if `data.attributes.enabled` is set to `true`.
+
+Key path                    | Type   | Default | Description
+----------------------------|--------|---------|------------
+`data.attributes.enabled`    | bool   | `false` | Allows Twilio to be used. If true, all remaining attributes are required.
+`data.attributes.account-sid`  | string   |  | The Twilio account id.
+`data.attributes.auth-token` | string   |  | The Twilio authentication token.
+`data.attributes.from-number`| string     |  | The Twilio registered phone number which will be used to send the message.
+`data.attributes.test-number`| string     |  | The phone number you wish to send a test SMS to. Not persisted and only used during testing.
```
armchairlinguist: I'd change this to something like "The target phone number for the test SMS."

content/source/docs/enterprise/api/ptfe_settings.html.md
armchairlinguist: I was wondering the same thing. Do we have existing language for that in other PATCH endpoints?

content/source/docs/enterprise/api/ptfe_settings.html.md
armchairlinguist: I prefer the way David did it - there's a lot less duplication. If every admin endpoint requires a user token, we can say that once and point people back there if they have questions.

caseylang: So which way should I go? David's description of authentication appears to be in the future index file. I can just remove all references to authentication from this setting file if that's what you want?

caseylang:
@skierkowski Would you mind give the SAML section a quick review, specifically the descriptions in the request body section of `Update SAML Settings`?

content/source/docs/enterprise/api/ptfe_settings.html.md
nfagerlund: Yes, I think removing the references to authentication is the way to go. 👍🏼

content/source/docs/enterprise/api/ptfe_settings.html.md
nfagerlund: @armchairlinguist @caseylang The closest thing to this is in the [update a workspace endpoint](https://www.terraform.io/docs/enterprise/api/workspaces.html#update-a-workspace). Take a look at the `data.attributes.vcs-repo` property: 

> To delete a workspace's existing VCS repo, specify null instead of an object. To modify a workspace's existing VCS repo, include whichever of the keys below you wish to modify. To add a new VCS repo to a workspace that didn't previously have one, include at least the oauth-token-id and identifier keys.

These all still have default values of "(previous value)". 

...I'll poke at this for a minute and push a commit, eh?

nfagerlund:
@caseylang I pasted a brief explanation above the various PATCH endpoints, which hopefully makes it pretty clear what's up.

content/source/docs/enterprise/api/admin/settings.html.md
```diff
@@ -0,0 +1,483 @@
...
+
+This PATCH endpoint requires a JSON object with the following properties as a request payload.
+
+If `data.attributes.enabled` is set to `true`, all remaining attributes must have valid values. You can omit attributes if they have a default value, or if a value was set by a previous update. Omitted attributes keep their previous values.
+
+Key path                    | Type   | Default | Description
+----------------------------|--------|---------|------------
+`data.attributes.enabled`    | bool   | `false` | Allows SAML to be used. If true, all remaining attributes are required.
+`data.attributes.idp-cert`   | string |         | Identity Provider Certificate specifies the PEM encoded X.509 Certificate as provided by the IdP configuration.
+`data.attributes.slo-endpoint-url` | string |         | Single Log Out URL specifies the HTTPS endpoint on your IdP for single logout requests. This value is provided by the IdP configuration.
```
skierkowski: @zhssh Does the SLO do anything? Do we just take this in case TFE would support SLO in the future?

zhssh: The SLO URL is not in use and we can remove this setting since won't be implementing it. Users are able to log out of TFE itself through the regular logout link. Having the link sign them out off all their apps could disruptive to their work, so we will keep the scope only to TFE.

skierkowski: @zhssh Do you have an Asana tracking the removal of the SLO endpoint (if not, can you open it)? Since this attribute is available now, we should keep it, but make sure we remove it when it's removed.

zhssh: @skierkowski I've created a task that includes the rationale for removing the setting.

content/source/docs/enterprise/api/admin/settings.html.md
```diff
@@ -0,0 +1,485 @@
...
+
+[200]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200
+[404]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404
+[422]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422
+[JSON API document]: https://www.terraform.io/docs/enterprise/api/index.html#json-api-documents
+[JSON API error object]: http://jsonapi.org/format/#error-objects
+
+### Request Body
+
+This PATCH endpoint requires a JSON object with the following properties as a request payload.
```
armchairlinguist: @nfagerlund's update for SAML and SMTP indicated that attributes not included retain their previous values there. I'm assuming that's also true here?

content/source/docs/enterprise/api/admin/settings.html.md
```diff
@@ -0,0 +1,485 @@
...
+`data.attributes.enabled`    | bool   | `false` | Allows SAML to be used. If true, all remaining attributes are required.
+`data.attributes.idp-cert`   | string |         | Identity Provider Certificate specifies the PEM encoded X.509 Certificate as provided by the IdP configuration.
+`data.attributes.slo-endpoint-url` | string |         | Single Log Out URL specifies the HTTPS endpoint on your IdP for single logout requests. This value is provided by the IdP configuration.
+`data.attributes.sso-endpoint-url` | string |         | Single Sign On URL specifies the HTTPS endpoint on your IdP for single sign-on requests. This value is provided by the IdP configuration.
+`data.attributes.attr-groups`| string | `"MemberOf"` | Team Attribute Name specifies the name of the SAML attribute that determines team membership.
+`data.attributes.attr-site-admin`| string |`"SiteAdmin"`| Specifies the role for site admin access, overriding the "Site Admin Role" method.
+`data.attributes.site-admin-role`| string |`"site-admins"`| Specifies the role for site admin access, provided in the list of roles sent in the Team Attribute Name attribute.
```
armchairlinguist: This bit is still not clear to me - how do these two attributes interact? What does each one mean for the format of the assertion? I think we need more of an explanation in the description, or a link to the relevant docs in the SAML section. @zhssh do you have any thoughts on the best way to word how these attributes are used?

content/source/docs/enterprise/api/admin/settings.html.md
```diff
@@ -0,0 +1,485 @@
...
+      "username": "sample_user"
+    }
+  }
+}
+```
+
+## Update SMTP Settings
+`PATCH /api/v2/admin/smtp-settings`
+
+When this request is submitted, a test message will be sent to the specified `test-email-address` address. If the test message delivery fails, the API will return an error code indicating the reason for the failure.
```
armchairlinguist: Nice.

PR #294, opened 2018-05-23
Document that configuration versions are paginated

ryanuber:
A pending API change makes the configuration versions index paginated. This documents it as such.


nfagerlund:
Oh wait, #295 has this as well...

ryanuber:
Oh that @beekus... haha! We can take his instead since it has the index update also.

PR #295, opened 2018-05-23
Pagination index update

beekus:
Add pagination docs for configuration versions.


beekus:
associated TFE pr: https://github.com/hashicorp/atlas/pull/6238

nfagerlund:
#294 has this as well, but doesn't update the index. QUESTION: Do we actually want to keep adding these pages to the index? I say no, because the admin pages API is gonna add like another five paginated endpoints iirc, so we should probably say something like "Most of the endpoints that return lists of objects support pagination."

beekus:
oh ha. woops. it was too early in the morning and i didn't see ryan's PR

beekus:
probably a good point. yeah, we can change the text there.

PR #296, opened 2018-05-23
Call out an issue with org rename and modules

caseylang:
It's not currently possible to rename an org which has modules, as those modules use the org name as their namespace. Discussion has begun on future resolutions to this issue. For now, a call out of the issue is sufficient.

Relates to https://app.asana.com/0/157256879238414/683075579461559/f

### DIFF
--- a/content/source/docs/enterprise/api/organizations.html.md
+++ b/content/source/docs/enterprise/api/organizations.html.md
@@ -221,6 +221,8 @@ curl \
 
 `PATCH /organizations/:organization_name`
 
+-> **Note**: Renaming an organization is not recommended if the organization has modules in the Private Module Registry. To rename an organization with private modules, first, delete the modules from the Private Module Registry. Then make the name change. After the change is complete, reload the modules into the Private Module Registry.
+
 Parameter            | Description
 ---------------------|------------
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/api/organizations.html.md
```diff
@@ -221,6 +221,8 @@ curl \
 
 `PATCH /organizations/:organization_name`
 
+-> **Note**: Renaming an organization is not currently supported if that organization also relies on the Private Module Registry. Existing modules will need to be deleted and later reloaded to accomplish this action. This functionality will be fully supported in a future release.
```
armchairlinguist: I think Ryan's suggestion of "isn't recommended if the organization has modules in the Private Module Registry" is the way I would go here. I'd also be more explicit in the second sentence:

"To rename an organization with private modules, first, delete the modules from the Private Module Registry. Then make the name change. After the change is complete, reload the modules into the Private Module Registry."

And I wouldn't say anything about the future. Customers email us asking "when will that happen" if you do. :) We can just delete the note when it happens.

caseylang: Updated

PR #299, opened 2018-05-29
Don't present Docker CE as an option for PTFE on RHEL

koikonom:
Although we know CE will work on RHEL installing it is not always
straightforward (CentOS package repositories often diverge from RHEL
causing install issues and support burden on us).

### DIFF
--- a/content/source/docs/enterprise/private/rhel-install-guide.html.md
+++ b/content/source/docs/enterprise/private/rhel-install-guide.html.md
@@ -11,8 +11,7 @@ This install guide is specifically for users of Private Terraform Enterprise ins
 ## Install Recommendations
 
 * RedHat Enterprise Linux version 7.3 or 7.4
-* Docker 1.13.1 (available in RHEL extras), or Docker CE or EE version 17.06 or later. The later versions are not available in the standard RHEL yum repositories.
-   * For Docker CE, you can use the official CentOS docker repository. Instructions are available here: https://docs.docker.com/install/linux/docker-ce/centos/#install-from-a-package.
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/private/rhel-install-guide.html.md
```diff
@@ -34,7 +33,7 @@ If you opt to use Docker from RHEL extras, then you must make a change to its de
 Sure! Just be sure to have at least 1.13.1 and authorization plugins disabled.
 
 ### When I run the installer, it allows me to download and install Docker CE on RedHat. Can I use that?
-Yes, Docker CE is compatible with the current installer; however it is not directly supported by RedHat. You still need to be sure that the storage backend is configured properly as by default, Docker will be using devicemapper in loopback, an entirely unsupported mode.
+Yes, Docker CE is compatible with the current installer; however it is not directly supported by RedHat (See https://access.redhat.com/articles/2726611). You still need to be sure that the storage backend is configured properly as by default, Docker will be using devicemapper in loopback, an entirely unsupported mode.
```
blalor: "See" → "see".

armchairlinguist: I'd suggest instead making this a true link, and changing the punctuation a bit:

```
Yes, Docker CE is compatible with the current installer. However, it is not directly [supported by RedHat](https://access.redhat.com/articles/2726611).
```

I'm also a little unsure what we're suggesting by saying that it's not directly supported...? What kind of hesitation should that give the customer, if we're still saying it's fine, ultimately?

PR #300, opened 2018-05-29
Screenshot updates

nfagerlund:
This is mostly about fixing screenshots that had wildly varying apparent sizes by re-shooting at the full viewport width (with roughly the same size viewport). As a bonus, it also updates the screenshots that were the most badly out-of-sync with the current UI. I also updated some terminology while I was mucking around in the VCS integration docs.


content/source/docs/enterprise/vcs/bitbucket-server.html.md
```diff
@@ -17,9 +17,9 @@ Note that Bitbucket Server requires both OAuth authentication and an SSH key. Th
 
 -> **Note:** Alternately, you can skip the OAuth configuration process and authenticate with a personal access token. This requires using TFE's API. For details, see [the OAuth Clients API page](../api/oauth-clients.html).
 
-## Step 1: On Bitbucket Server, Ensure the Webhooks Plugin is Installed
+## Step 1: On Bitbucket Server, Ensure the Webhooks Plugin is Installed (Versions < 5.4)
```
armchairlinguist: Unfortunately, the plugin has to be installed on all versions as far as I know.

content/source/docs/enterprise/vcs/github-enterprise.html.md
```diff
@@ -50,22 +50,22 @@ The rest of this page explains the GitHub Enterprise versions of these steps.
     Application Description    | Any description of your choice.
     Authorization callback URL | `https://example.com/replace-this-later` (or any placeholder; the correct URI doesn't exist until the next step.)
 
-4. Click the "Register application" button. This will take you to the application page.
+4. Click the "Register application" button, which creates the application and takes you to its page.
 
 5. Take note of two items: the **Client ID** and the **Client Secret.** You'll copy and paste these unique strings in the next step. Leave this page open in a browser tab.
```
armchairlinguist: Wouldn't hurt to upgrade this one the same as the earlier one:

> Leave this page open in a browser tab. In the next step, you will copy and paste the unique **Client ID** and the **Client Secret.**

content/source/docs/enterprise/vcs/github-enterprise.html.md
```diff
@@ -78,23 +78,23 @@ The rest of this page explains the GitHub Enterprise versions of these steps.
 
     ![TFE screenshot: add client fields](./images/ghe-tfe-add-client-fields.png)
 
-3. Click "Create connection." This will take you back to the OAuth Configuration page, which now includes your new GitHub Enterprise client.
+3. Click "Create VCS Provider." This will take you back to the VCS Provider page, which now includes your new GitHub client.
```
armchairlinguist: Hm, seems we should keep "Enterprise" here? Or is there some reason you removed it?

content/source/docs/enterprise/workspaces/run-ui.html.md
```diff
@@ -32,7 +32,7 @@ If the workspace has a plan that is still in the [plan stage](./run-basics.html#
 
 ## Confirming or Discarding Plans
 
-By default, run plans require confirmation before TFE will apply them. Users with write access on a workspace can visit the page for a run that has finished planning and click the "Confirm & Apply" or "Discard Plan" button to finish or cancel a run.
+By default, run plans require confirmation before TFE will apply them. Users with write access on a workspace can navigate to a run that has finished planning and click the "Confirm & Apply" or "Discard Plan" button to finish or cancel a run. Use the "View Plan" button for more details about what the run will change.
```
armchairlinguist: I think there's an Asana for the fact that the plan doesn't pop open automatically, so it might be more future-proof to say "If necessary, use..."

content/source/docs/enterprise/vcs/github-enterprise.html.md
nfagerlund: That was a diffing error! Thanks for catching, I'll fix it.

content/source/docs/enterprise/workspaces/run-ui.html.md
nfagerlund: 👍🏼

content/source/docs/enterprise/vcs/github-enterprise.html.md
nfagerlund: 👍🏼

nfagerlund:
@armchairlinguist Rebased!

content/source/docs/enterprise/vcs/github-enterprise.html.md
nfagerlund: I went and caught all of them, since there was a variant of that everywhere.

nfagerlund:
I'm going to ignore that Travis check, because my local preview isn't busted. We'll see whether that comes to bite me. :)

PR #302, opened 2018-05-30
Add rundeck provider

cgriggs01:


content/source/layouts/runscope.erb
```diff
@@ -0,0 +1 @@
+../../../ext/providers/PROVIDER/website/runscope.erb
```
nfagerlund: This symlink still has `PROVIDER` in its path — gotta replace it with `runscope` for this to work.

PR #306, opened 2018-06-01
Adding FlexibleEngine to website

cgriggs01:


content/source/docs/providers/flexibleengine
```diff
@@ -0,0 +1 @@
+ext/providers/flexibleengine/website/docs
```
bill-rich: I think this should be "../../../../ext/providers/flexibleengine/website/docs"

content/source/layouts/flexibleengine.erb
```diff
@@ -0,0 +1 @@
+ext/providers/flexibleengine/website/flexibleengine.erb
```
bill-rich: I think this should be "../../../ext/providers/flexibleengine/website/flexibleengine.erb"

cgriggs01: Found those, thanks!

PR #309, opened 2018-06-04
Adds Note about debugging Git server SSL certs

petems:
Something we've used a lot when debugging, would be nice to have in the docs! 😄

### DIFF
--- a/content/source/docs/enterprise/vcs/troubleshooting.html.md
+++ b/content/source/docs/enterprise/vcs/troubleshooting.html.md
@@ -84,3 +84,101 @@ This often happens when the VCS connection has been somehow broken: it might hav
 The domain name for Terraform Enterprise changed on 02/22 at 9AM from atlas.hashicorp.com to app.terraform.io. If the OAuth client was originally configured on the old domain, using it for a new VCS connection can result in this error.
 
 The fix is to update the OAuth Callback URL in your VCS provider to use app.terraform.io instead of atlas.hashicorp.com.
+
+## Debugging Certificate Errors on Private Terraform Enterprise
+
+When debugging failures of VCS connections due to certifcate errors, it's sometime helpful to run the OpenSSL command within the TFE container to get more information.
+
+Attach a bash session to the PTFE container:
+
+```
+docker exec -it ptfe_atlas sh -c "stty rows 50 && stty cols 150 && bash"
+```
+
+Then run the `openssl s_client` command, using the certificate from the container, and you'll be able to see the SSL connection, and any errors that come up.
+
+The certificate will be available under `/tmp/cust-ca-certificates.crt` in the container:
+
+```
+openssl s_client -showcerts -CAfile /tmp/cust-ca-certificates.crt -connect git-server-hostname:443
+```
+
+For example, if my GitLab server was running self-signed certificates, I might get a `verify error:num=18:self signed certificate` log like so:
+
+```
+bash-4.3# openssl s_client -showcerts -CAfile /tmp/cust-ca-certificates.crt -connect gitlab.local:443
+CONNECTED(00000003)
+depth=0 CN = gitlab.local
+verify error:num=18:self signed certificate
+verify return:1
+depth=0 CN = gitlab.local
+verify return:1
+---
+Certificate chain
+ 0 s:/CN=gitlab.local
+   i:/CN=gitlab.local
+-----BEGIN CERTIFICATE-----
+MIIC/DCCAeSgAwIBAgIJAIhG2GWtcj7lMA0GCSqGSIb3DQEBCwUAMCAxHjAcBgNV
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/vcs/troubleshooting.html.md
```diff
@@ -84,3 +84,101 @@ This often happens when the VCS connection has been somehow broken: it might hav
...
+```
+docker exec -it ptfe_atlas sh -c "stty rows 50 && stty cols 150 && bash"
+```
+
+Then run the openssl command, using the certificate from the container, and you'll be able to see the SSL connection, and any errors that come up.
+
+The certificate will be available under `/tmp/cust-ca-certificates.crt` in the container:
+
+```
+$ openssl s_client -showcerts -CAfile /tmp/cust-ca-certificates.crt -connect git-server-hostname:443
```
nfagerlund: Let's be consistent about whether to put a sigil at the start of the line, if it's a line standing by itself. 

...but I've been doing GUI stuff so long I can't remember what our standard practice is at the moment! Let's go with "not" for now.

content/source/docs/enterprise/vcs/troubleshooting.html.md
```diff
@@ -84,3 +84,101 @@ This often happens when the VCS connection has been somehow broken: it might hav
...
+
+When debugging failures of VCS connections due to certifcate errors, it's sometime helpful to run the OpenSSL command within the TFE container to get more information.
+
+Attach a bash session to the PTFE container:
+
+```
+docker exec -it ptfe_atlas sh -c "stty rows 50 && stty cols 150 && bash"
+```
+
+Then run the openssl command, using the certificate from the container, and you'll be able to see the SSL connection, and any errors that come up.
```
nfagerlund: May as well specify that we mean the `openssl s_client` command, even though we have examples below. I had a brief moment of panic reading this sentence, worried I'd have to figure out which subcommand to use. :)

content/source/docs/enterprise/vcs/troubleshooting.html.md
```diff
@@ -84,3 +84,101 @@ This often happens when the VCS connection has been somehow broken: it might hav
...
+
+The certificate will be available under `/tmp/cust-ca-certificates.crt` in the container:
+
+```
+$ openssl s_client -showcerts -CAfile /tmp/cust-ca-certificates.crt -connect git-server-hostname:443
+```
+
+For example, if my GitLab Server was running self-signed certificates, I might get a `verify error:num=18:self signed certificate` log like so:
```
nfagerlund: nocaps "server"

content/source/docs/enterprise/vcs/troubleshooting.html.md
```diff
@@ -84,3 +84,101 @@ This often happens when the VCS connection has been somehow broken: it might hav
 The domain name for Terraform Enterprise changed on 02/22 at 9AM from atlas.hashicorp.com to app.terraform.io. If the OAuth client was originally configured on the old domain, using it for a new VCS connection can result in this error.
 
 The fix is to update the OAuth Callback URL in your VCS provider to use app.terraform.io instead of atlas.hashicorp.com.
+
+### Debugging Certificate Errors on Private Terraform Enterprise
```
nfagerlund: Let's drop one of those `#`s to make this a new top-level H2. The rest of the stuff in General is "symptom -> likely problem -> fix," and this is something different that doesn't quite fit with those.

petems: Done! 👍

content/source/docs/enterprise/vcs/troubleshooting.html.md
petems: Done! 👍

content/source/docs/enterprise/vcs/troubleshooting.html.md
petems: Done! 👍

content/source/docs/enterprise/vcs/troubleshooting.html.md
petems: Done! 👍

petems:
All changes made as requested 👍

skierkowski:
Should these go somewhere under `./private` instead?

petems:
@skierkowski I think it makes sense to put them under the rest of the VCS Provider troubleshooting tips, as that's usuaully the page the user is on if they're trying to figure out git issues.

content/source/docs/enterprise/vcs/troubleshooting.html.md
armchairlinguist: We're actually not very consistent about this - it would probably be worth going through the docs at some point to clean up :)

PR #310, opened 2018-06-05
Stop using $ATLAS_TOKEN

nfagerlund:


skierkowski:
`./private/faq.html.md`
`./upgrade/index.html.md`
These two files also need to be updated too as they have references to ATLAS_TOKEN, even though not in the API Docs.

nfagerlund:
@skierkowski ALAS, those have to remain. The `atlas` backend built into terraform still uses the `$ATLAS_TOKEN` environment variable, so that's what we're stuck setting until the backend changes. 

Unless we already changed the backend at some point to accept an alternate env var! But the docs don't seem to indicate that we did.

PR #312, opened 2018-06-05
Update docs to replace filter with relationships

svanharmelen:
This change is done inline with the changes in https://github.com/hashicorp/atlas/pull/6303

### DIFF
--- a/content/source/docs/enterprise/api/variables.html.md
+++ b/content/source/docs/enterprise/api/variables.html.md
@@ -22,16 +22,25 @@ This POST endpoint requires a JSON object with the following properties as a req
 Properties without a default value are required.
 
 
-Key path                    | Type   | Default | Description
-----------------------------|--------|---------|------------
-`data.type`                 | string |         | Must be `"vars"`.
-`data.attributes.key`       | string |         | The name of the variable.
-`data.attributes.value`     | string |         | The value of the variable.
-`data.attributes.category`  | string |         | Whether this is a Terraform or environment variable. Valid values are `"terraform"` or `"env"`.
-`data.attributes.hcl`       | bool   | `false` | Whether to evaluate the value of the variable as a string of HCL code. Has no effect for environment variables.
-`data.attributes.sensitive` | bool   | `false` | Whether the value is sensitive. If true then the variable is written once and not visible thereafter.
-`filter.workspace.name`     | string |         | The name of the workspace that owns the variable.
-`filter.organization.name`  | string |         | The name of the organization that owns the workspace.
+Key path                                    | Type   | Default | Description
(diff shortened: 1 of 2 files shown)


nfagerlund:
Oh man, I remember wondering whether that was a docs error or what. Hooray for making it consistent!

content/source/docs/enterprise/api/variables.html.md
```diff
@@ -46,14 +48,21 @@ Key path                    | Type   | Default | Description
...
+        "data": {
+          "type":"organizations",
+          "id":"my-organization"
+          }
+        }
+      },
+      "workspace": {
+        "data": {
+          "type":"workspaces",
+          "id":"ws-123456789"
```
armchairlinguist: This id should look like a real workspace ID.

svanharmelen:
Thanks for the feedback @armchairlinguist! I updated the workspace ID and added a deprecation warning. The `JSONAPI` specs do not offer a way to deprecate API's or API fields, so indeed adding this to the docs seems like the right thing to do 👍

content/source/docs/enterprise/api/variables.html.md
```diff
@@ -22,16 +22,25 @@ This POST endpoint requires a JSON object with the following properties as a req
...
+`data.attributes.value`                     | string |         | The value of the variable.
+`data.attributes.category`                  | string |         | Whether this is a Terraform or environment variable. Valid values are `"terraform"` or `"env"`.
+`data.attributes.hcl`                       | bool   | `false` | Whether to evaluate the value of the variable as a string of HCL code. Has no effect for environment variables.
+`data.attributes.sensitive`                 | bool   | `false` | Whether the value is sensitive. If true then the variable is written once and not visible thereafter.
+`data.relationships.workspace.data.type`    | string |         | Must be `"workspaces"`.
+`data.relationships.workspace.data.id`      | string |         | The ID of the workspace that owns the variable.
+`data.relationships.organization.data.type` | string |         | Must be `"organizations"`.
+`data.relationships.organization.data.id`   | string |         | The ID (which is the name) of the organization that owns the workspace.
+
+**Deprecation warning**: The following properties are deprecated and will be removed from feature versions of the API!
```
armchairlinguist: This looks like a great add. I'd replace "feature" with "future", and note something like "`filter` properties have been replaced by `relationships`.

svanharmelen: That's indeed a typo, will update that and add a word about the `filter` <-> `relationships` change as well. Thought that would be obvious, doesn't hurt to write it down 👍 Thx!

svanharmelen:
@armchairlinguist tweaked it a bit more... Could you double check before I merge the PR?

PR #316, opened 2018-06-08
Revise service account docs again

nfagerlund:
Second time's a charm. Let's really drill into people that you CAN'T atlas backend with these puppies. 

This PR also includes some tweaks to the API shortcut callouts, since they came up in the process of updating this.

### DIFF
--- a/content/source/docs/enterprise/users-teams-organizations/service-accounts.html.md
+++ b/content/source/docs/enterprise/users-teams-organizations/service-accounts.html.md
@@ -6,17 +6,35 @@ sidebar_current: "docs-enterprise2-users-teams-organizations-service-accounts"
 
 # Service Accounts
 
-Terraform Enterprise provides two types of service accounts. These accounts can be granted permission to access Terraform Enterprise APIs, but cannot be used interactively. The Service Accounts are designed to support server-to-server API calls using the service identity as opposed to individual user identities.
+Terraform Enterprise (TFE) provides two types of service accounts: team, and organization. These accounts can access Terraform Enterprise APIs, but cannot be used interactively. The service accounts are designed to support server-to-server API calls using the service identity as opposed to individual user identities.
+
+Team and organization service accounts don't have API tokens by default; you must generate a token before you can use them.
+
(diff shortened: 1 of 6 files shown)


content/source/docs/enterprise/users-teams-organizations/service-accounts.html.md
```diff
@@ -6,34 +6,37 @@ sidebar_current: "docs-enterprise2-users-teams-organizations-service-accounts"
...
+Terraform Enterprise (TFE) provides two types of service accounts: team, and organization. These accounts can access Terraform Enterprise APIs, but cannot be used interactively. The service accounts are designed to support server-to-server API calls using the service identity as opposed to individual user identities.
+
+Team and organization service accounts don't have API tokens by default; you must generate a token before you can use them.
+
+Like with [user tokens](./users.html#api-tokens), service account tokens are displayed only once when they are created, and are obfuscated thereafter. If the token is lost, it must be regenerated. When a token is regenerated, the previous token immediately becomes invalid.
+
+~> **Important:** Team and organization tokens cannot authenticate with Terraform's `atlas` backend for running Terraform locally; they only work with the v2 API. If you need to use the `atlas` backend (to migrate state, for example), use a [user token](./users.html#api-tokens).
```
fprimex: Moving this to a callout is great. It also needs to be expanded to explicitly call out that not only can you not use these tokens for local backends, but you _also_ can't use them for Terraform Remote States.

It should make sense to people that this won't work, since the Terraform Remote State resource literally takes a backend configuration as a config block, but a simple "or" statement here would make it explicit.

> cannot authenticate with Terraform's `atlas` backend for running Terraform locally or authenticate the `terraform_remote_state` resource for accessing other states

Since both of these have to do with state access, it might read better to state it that way and then elaborate

> cannot be used to access states stored in workspaces using either the `atlas` backend or the `terraform_remote_state` resource. Use a user token to access states instead.

nfagerlund: ok! I guess it hadn't even occurred to me that you could use a remote state data source without already setting up an atlas backend block? ...actually, can you? I feel like I ran into issues with it at one point. Well, never mind, I'll add a note!

nfagerlund:
Ready for a re-check. That should cover all the bases.

PR #326, opened 2018-06-18
Small voice/clarity cleanup for the new cert troubleshooting section in VCS

armchairlinguist:


content/source/docs/enterprise/vcs/troubleshooting.html.md
```diff
@@ -85,25 +85,23 @@ The domain name for Terraform Enterprise changed on 02/22 at 9AM from atlas.hash
...
-
-The certificate will be available under `/tmp/cust-ca-certificates.crt` in the container:
+Then run the `openssl s_client` command, using the certificate at `/tmp/cust-ca-certificates.crt` in the container:
 
 ```
 openssl s_client -showcerts -CAfile /tmp/cust-ca-certificates.crt -connect git-server-hostname:443
 ```
 
-For example, if my GitLab server was running self-signed certificates, I might get a `verify error:num=18:self signed certificate` log like so:
+For example, the error `verify error:num=18:self signed certificate` might be returned if a GitLab server was running with a self-signed certificate. This example output is shown below.
```
nfagerlund: Mmm, I think reversing the order and adding passive voice makes this harder to parse. How about: 

> For example, a VCS server that uses a self-signed certificate might cause an error like `verify error:num=18:self signed certificate`, as shown in the output below:

armchairlinguist: How about:

> For example, a Gitlab server that uses a self-signed certificate might result in an error like `verify error:num=18:self signed certificate`, as shown in the output below:

nfagerlund: Oh yeah, works for me!

armchairlinguist:
The travis failure looks unrelated, so I'm going to go ahead and merge this.

PR #330, opened 2018-06-19
Fix broken layout in terraform versions admin API page

nfagerlund:
YAML frontmatter is very picky about the number of hyphens.


nfagerlund:
haha, it's 'cause the diff "minus" marker is disguising itself as another hyphen. STEALTH

PR #333, opened 2018-06-19
fix vars payload json issue

armchairlinguist:
We had a stray brace in there.


armchairlinguist:
Cool. I think @svanharmelen is the last touch here, and I've requested his review as well.

PR #342, opened 2018-06-29
clarify database user requirements

armchairlinguist:
A customer had a lot of trouble setting up PTFE because he hadn't created a user with the right role - as a non-DBA, he found "User with access to ownership semantics" to be opaque (as did I).

I discussed the best way to express this in a way that's general but clear with @koikonom, and we landed on this, but I'm open to changing it up as long as it's something that a technical user who isn't a database expert can act on.

Also fix a stray comma.

### DIFF
--- a/content/source/docs/enterprise/private/install-installer.html.md
+++ b/content/source/docs/enterprise/private/install-installer.html.md
@@ -264,7 +264,7 @@ When Terraform Enterprise uses an external PostgreSQL database, the
 following must be present on it:
 
 * PostgreSQL version 9.4 or greater
-* User with full access to the schemas created and the ability to run "CREATE EXTENSION" on the database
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/private/install-installer.html.md
```diff
@@ -264,7 +264,7 @@ When Terraform Enterprise uses an external PostgreSQL database, the
 following must be present on it:
 
 * PostgreSQL version 9.4 or greater
-* User with access to ownership semantics on the referenced database
+* User with full access to the schemas created and the ability to run "CREATE EXTENSION" on the database
```
koikonom: This looks fine.

If you want you can replace `full access` with something along the lines of "User with all privileges granted on the schemas created and ...".

That kinda hints readers to use the GRANT ALL command to do the right thing.

PR #347, opened 2018-07-03
Add section about data backup and rollback

evanphx:


content/source/docs/enterprise/private/migrate.html.md
```diff
@@ -9,6 +9,24 @@ sidebar_current: "docs-enterprise2-private-installer-migration"
 This document outlines the procedure for migrating from the AMI-based Private Terraform Enterprise (PTFE)
 to the Installer-based PTFE.
 
+## Backup
```
koikonom: I think we should add  a small sentence saying that they also need to backup their state file.

content/source/docs/enterprise/private/migrate.html.md
```diff
@@ -9,6 +9,24 @@ sidebar_current: "docs-enterprise2-private-installer-migration"
...
+
+Before beginning, it's best create an additional backup of your RDS database. This will allow you to rollback the data and continue to use the AMI if necessary.
```
armchairlinguist: best create > best to create

I think I prefer "roll back" for the verb (roll back the data). But since this is a specialized use of roll back, I don't feel strongly.

content/source/docs/enterprise/private/migrate.html.md
```diff
@@ -9,6 +9,24 @@ sidebar_current: "docs-enterprise2-private-installer-migration"
...
+
+To do so, go to the [Amazon RDS Instances](https://console.aws.amazon.com/rds/home?region=us-east-1#dbinstances:). You may need to change the region that you are viewing in order to see your PTFE instance. Once you find it, click on the instance name. On the next page, select **Instance Actions** and then **Take snapshot**. On the **Take DB Snapshot** page, enter a name for the snapshot such as _pre installer migration_, and then click **Take Snapshot**.
```
armchairlinguist: To do so > To create an RDS backup

Conventionally, I think we usually put literal text entry in `backticks`, so it would read:

"enter a name for the snapshot, such as `Pre-installer Migration`, and then click...

content/source/docs/enterprise/private/migrate.html.md
```diff
@@ -9,6 +9,24 @@ sidebar_current: "docs-enterprise2-private-installer-migration"
...
+
+The snapshot will take a little while to create. After it has finished, you can continue with the rest of the migration steps.
+
+### Reverting back to AMI
```
armchairlinguist: I think I'd add "the" here, "Reverting back to the AMI"

content/source/docs/enterprise/private/migrate.html.md
```diff
@@ -9,6 +9,24 @@ sidebar_current: "docs-enterprise2-private-installer-migration"
...
+
+If you wish to revert back to the AMI after running the migration script, first if you've already manipulated the state file to move the resources, you'll need to restore the original state file.
```
armchairlinguist: ~If you wish~ To revert ~back~ to...

I think we'll want to separate and number these steps, so this would go:

> ....after running the migration script:

>1. If you've already manipulated the state file to move the AMI resources, restore the original state file.
>2. With the state file in place, return...
>3. Run `terraform apply`...
>4. Return to the original hostname used for the cluster.

content/source/docs/enterprise/private/migrate.html.md
```diff
@@ -9,6 +9,24 @@ sidebar_current: "docs-enterprise2-private-installer-migration"
...
+
+With your original terraform state file in place, return to the [Amazon RDS Snapshots](https://console.aws.amazon.com/rds/home?region=us-east-1#db-snapshots:) and find the snapshot you created before migrating. Click on the snapshot and note it's **ARN** value. Open your **.tfvars** file and add `db_snapshot = "arn-value-of-snapshot"`.
```
armchairlinguist: it's **ARN** value > its **ARN** value

content/source/docs/enterprise/private/migrate.html.md
```diff
@@ -9,6 +9,24 @@ sidebar_current: "docs-enterprise2-private-installer-migration"
...
+
+Next you'll  run `terraform apply` to make sure everything is setup. This will result in a new RDS instance being built against the snapshot.. If the EC2 instance is still running from the migration process, be sure to issue a `shutdown` on it to get a new instance created.
```
armchairlinguist: everything is setup > everything is set up
snapshot.. > snapshot.

This might just be me not being familiar with the setup, but when you say "If the EC2 instance is still running from the migration process..." I wonder if it's more clear to add a step before this (so would be step 3 in the numbers) saying `If the original AMI EC2 instance is still running, issue a shutdown command on it so that this process can create a new EC2 instance.`

content/source/docs/enterprise/private/migrate.html.md
armchairlinguist: @koikonom 's note about backing up the state file might fit well in item 1 here, too?

content/source/docs/enterprise/private/migrate.html.md
rberlind: what state file?  Isn't the state in S3?

evanphx: They should already have a strategy for managing their statefile. I can be more explicit here, but since the product in deployed with terraform, they've already been managing the state file in some way.

content/source/docs/enterprise/private/migrate.html.md
```diff
@@ -9,6 +9,28 @@ sidebar_current: "docs-enterprise2-private-installer-migration"
 This document outlines the procedure for migrating from the AMI-based Private Terraform Enterprise (PTFE)
 to the Installer-based PTFE.
 
+## Terraform State
+
+To run this procedure, you'll need the terraform state file used to create the AMI based installation. Additionaly, we strongly suggest you backup this state file before proceeding with this process in the event that you need to revert back to the AMI.
```
armchairlinguist: AMI based > AMI-based
Additionaly > Additionally
backup > back up
Add comma after "process"

evanphx: nit pick away! Fixed!

koikonom:
:shipit:

PR #353, opened 2018-07-11
no need to delete/recreate workspaces anymore, yay

armchairlinguist:
Now that we can edit workspaces, fortunately, we don't need to delete them :)


nfagerlund:
Deleting documentation is my favorite thing

PR #355, opened 2018-07-13
No migration between internal and external vault

armchairlinguist:
Add a callout on the Vault page and a section on the install page to alert users that the Vault decision is one and done.


armchairlinguist:
@evanphx any chance you could take a look at this during this week?

PR #360, opened 2018-07-18
Document updated postgresql params for automated install

blalor:


content/source/docs/enterprise/private/automating-the-installer.html.md
```diff
@@ -85,7 +85,6 @@ ptfe$ cat settings.json
         "value": "poc"
     },
     "placement": {},
-    "postgres_url": {},
```
armchairlinguist: Should this output also be updated with the replacements?

content/source/docs/enterprise/private/automating-the-installer.html.md
```diff
@@ -146,7 +145,11 @@ The following apply to every installation:
 
 #### The following apply if you've chosen `external` for `production_type`:
 
-- `postgres_url` — (Required) the database URL
+- `pg_user` — (Required) PostgreSQL user name to connect as
```
armchairlinguist: I might omit "name" here since we just use user in the next line, but I don't feel strongly.

content/source/docs/enterprise/private/automating-the-installer.html.md
```diff
@@ -146,7 +145,11 @@ The following apply to every installation:
...
+- `pg_password` — (Required) The password for the PostgreSQL user
+- `pg_netloc` — (Required) The `hostname:port` of the target PostgreSQL server
```
armchairlinguist: This makes sense to me, but I wonder if it's better to be extremely explicit and say "The hostname and port of the target...in the format `hostname:port`."

content/source/docs/enterprise/private/automating-the-installer.html.md
blalor: I thought about that, but this output is really just a representative sample (and not a very good one, at that).  I'll revisit this example output.

blalor:
@armchairlinguist I've addressed your feedback; mind giving it another 👀?

PR #358, opened 2018-07-18
TFE Admin API updates/refinements

armchairlinguist:
Mostly just updates for consistency/clarity, plus some added info and fixes to out of date info.

PR #361, opened 2018-07-19
Remove the two-factor endpoints from the API docs

zhssh:


armchairlinguist:
Oh right I forgot we were doing this when I submitted my other admin PR. This is likely to result in merge conflicts with https://github.com/hashicorp/terraform-website/pull/358 - do you want me to just do it there?

zhssh:
Reopening because the other PR will affect other docs not these ones.

PR #363, opened 2018-07-19
ADFS integration guide

koikonom:
How to configure ADFS in order to be used by PTFE for SAML logins.


zhssh:
Thank you for adding this!

content/source/docs/enterprise/saml/configuration.html.md
```diff
@@ -13,6 +13,10 @@ SAML requires the configuration of two parties:
 
 Terraform Enterprise is configured as the Service Provider.
 
+
+
+~> For MS ADFS specific instructions follow [this link](./adfs.html)
```
armchairlinguist: Instead of linking directly to ADFS here, I'm thinking we should say "For instructions for specific IdPs, see [Identity Provider Configuration](https://www.terraform.io/docs/enterprise/saml/identity-provider-configuration.html)."

Thoughts?

content/source/docs/enterprise/saml/identity-provider-configuration-adfs.html.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Configure PTFE
+
+1. Visit `https://<TFE HOSTNAME>/app/admin/saml`.
+2. Set "Single Sign-on URL" to `https://<ADFS hostname>/<URL Path>`.
+3. Set "Single Log-out URL" to `https://<ADFS hostname>/<URL Path>?wa=wsignout1.0`.
+4. Paste the contents of the saved certificate in "IDP Certificate".
+5. Scroll to the bottom of the screen and click "Save SAML Settings".
+
+## Configure ADFS
```
armchairlinguist: I wonder if we want to move this up above Configure PTFE, the way Zacc has it arranged in the OneLogin doc? That would save some of the repeats, and maybe be more efficient.

content/source/docs/enterprise/saml/identity-provider-configuration-adfs.html.md
```diff
@@ -0,0 +1,103 @@
...
+layout: enterprise2
+page_title: "ADFS Configuration - Terraform Enterprise"
+sidebar_current: "docs-enterprise2-private-saml-identity-provider-configuration-adfs"
+---
+
+# ADFS Configuration
+
+This guide explains how to configure Active Directory Federated Services (ADFS) in order to use it as an Identity Provider (IdP) for PTFE's SAML authentication feature. The screenshots below were taken on Windows Server 2016, and the UI may not look the same on previous Windows versions.
+
+This document assumes that you have already installed and configured ADFS and that you are using PTFE version v201807-1 or later.
```
armchairlinguist: Is this mostly about using the new SAML settings endpoint? I wonder how we should handle this...the OneLogin one is older; should we update that, or move this one back to v1, or...?

koikonom: This is a guide on how to configure SAML when your IdP is MS ADFS end to end. The fact that the metadata URL is there was just an added bonus.

I think it makes more sense to update the OneLogin one too.

content/source/docs/enterprise/saml/configuration.html.md
koikonom: Yeah I Was about to suggest that we remove these lines entirely, but I think your idea is better!

content/source/docs/enterprise/saml/identity-provider-configuration-adfs.html.md
koikonom: PTFE needs to be configured before we point ADFS to the metadata URL, otherwise it will not work.

armchairlinguist: Oh, got it! I totally missed the metadata URL use, very clever. This makes sense, then.

content/source/docs/enterprise/saml/identity-provider-configuration-adfs.html.md
armchairlinguist: Yeah, understanding that you're using the metadata URL, I agree. I'll write a quick task to remind us to do that at some point.

content/source/docs/enterprise/saml/configuration.html.md
armchairlinguist: Awesome.

phinze:
It think I addressed the last outstanding issues. Gonna merge this!

koikonom:
Thanks!

PR #370, opened 2018-07-25
Update Admin User Impersonation docs

caseylang:
A few quick updates to the admin user impersonation docs:
* Adds `is-service-account` to the API
* Requests the `:id` parameter instead of `:username`
* Adds `reason` property to the request

### DIFF
--- a/content/source/docs/enterprise/api/admin/users.html.md
+++ b/content/source/docs/enterprise/api/admin/users.html.md
@@ -479,15 +479,33 @@ Because of the requirement to provide a valid admin user session cookie in order
 
 Status  | Response                  | Reason
 --------|---------------------------|----------
-[204][] | Empty body                | Successfully impersonated the user
+[204][] | Empty body                | Successfully impersonated the user.
+[400][] | [JSON API error object][] | A reason for impersonation is required.
 [403][] | [JSON API error object][] | Client is already impersonating another user.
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/api/admin/users.html.md
```diff
@@ -473,23 +479,43 @@ Because of the requirement to provide a valid admin user session cookie in order
 
 Status  | Response                  | Reason
 --------|---------------------------|----------
-[204][] | Empty body                | Successfully impersonated the user
+[204][] | Empty body                | Successfully impersonated the user.
+[400][] | [JSON API error object][] | A reason for impersonation is required.
 [403][] | [JSON API error object][] | Client is already impersonating another user.
+[403][] | [JSON API error object][] | User is a Service Account which cannot be impersonated.
```
davidcelis: Admins are also not able to be impersonated!

armchairlinguist: That's not the case anymore; there was a change a while back (at my request) that removed that restriction.

caseylang: I was gonna say, I don't see that in the code

davidcelis: Oh! My mistake, then

PR #373, opened 2018-07-26
Add user impersonation restriction note

caseylang:
Adds a note to Admin User Impersonation on which accounts can be impersonated.


content/source/docs/enterprise/api/admin/users.html.md
```diff
@@ -471,6 +471,8 @@ Parameter   | Description
 
 Impersonation allows an admin to begin a new session as another user in the system. This can be helpful in reproducing issues that a user is experiencing with their account that the admin cannot reproduce themselves. While an admin is impersonating a user, any actions that are logged to the audit log will reflect that an admin was acting on another user's behalf. The `"actor"` key will reference the impersonated user, and an added `"admin"` key will contain the username of the admin acting on the user's behalf. For more information, see the [audit logging documentation][audit logging].
 
+-> **Note**: It is not possible to impersonate service accounts or your own account.
```
armchairlinguist: I think we usually mark up notes with `~>` which invokes some kind of nice blue block thing.

caseylang: So `~> **Note**` instead of `-> **Note**`? I pulled the latter version from the top of this doc but I guess that's the yellow style?

armchairlinguist: Ignore me, I was confused by looking at the GH-rendered MD instead of the real stuff! This is good to go.

PR #378, opened 2018-07-30
Update postgres connection settings reference for ptfe

blalor:
The mention of the URL is outdated, and it's worth repeating the
restrictions on the sslmode param.


nfagerlund:
Should this be merged now?

armchairlinguist:
@nfagerlund yas pls!

PR #379, opened 2018-07-30
Fixed external vault commands to use more modern syntax

rberlind:
I had my Git email or username wrong on my system somehow, so first commit came through as "Test User" but I then fixed so it shows up as from rberlind (me).

3 changes to Vault commands to use more modern syntax that won't give yellow warnings when running as the current ones do.

### DIFF
--- a/content/source/docs/enterprise/private/vault.html.md
+++ b/content/source/docs/enterprise/private/vault.html.md
@@ -25,10 +25,10 @@ Vault options.
 
 Use the following as a guide to configure an external Vault instance:
 
-1. Enable AppRole: `vault auth-enable approle`
-1. Enable transit: `vault mount transit`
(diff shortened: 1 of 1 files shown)


nfagerlund:
Hey @armchairlinguist, do you have a good suggestion for someone to double-check these commands real quick? I don't know Vault well enough.

armchairlinguist:
I added a couple review requests! It's certainly not me. :)

rberlind:
https://www.vaultproject.io/docs/commands/auth/enable.html
https://www.vaultproject.io/docs/commands/secrets/enable.html
https://www.vaultproject.io/docs/commands/policy/write.html

nfagerlund:
@rberlind Oh hey, thank you. That... seems clear enough that I don't think we need another reviewer.

PR #388, opened 2018-08-01
update/reorg support & bundle generation content

armchairlinguist:
I changed the FAQ to refer to the diagnostics page instead of only having the AMI info (and a duplicate of it at that). Moved the info we didn't initially have on the diagnostics page over as well.

### DIFF
--- a/content/source/docs/enterprise/private/diagnostics.html.md
+++ b/content/source/docs/enterprise/private/diagnostics.html.md
@@ -6,8 +6,7 @@ sidebar_current: "docs-enterprise2-private-diagnostics"
 
 # Private Terraform Enterprise Diagnostics
 
-This document contains information on how to provide HashiCorp with diagnostic information about a Private Terraform Enterprise (PTFE)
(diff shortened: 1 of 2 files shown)


content/source/docs/enterprise/private/diagnostics.html.md
```diff
@@ -82,10 +81,40 @@ hashicorp-support.tar.gz.enc
...
+        --cipher-algo AES256 \
+        --compress-algo ZLIB \
+        -o hashicorp-support.tar.gz.enc \
+        hashicorp-support.tar.gz
+
+You will note that we first create a support bundle using the normal procedure,
+extract it, remove the files we want to omit, and then create a new one.
+
+
+## Windows
```
blalor: here, you dropped this: `#`

armchairlinguist: Ah yes. This was intentional, but mistaken, since it only applies to the AMI.

PR #389, opened 2018-08-02
Nav: Expand PTFE items into the top nav level, and add clarifying headers

nfagerlund:
**Discuss before merging,** since this might be mildly controversial. 

@armchairlinguist pointed out that the PTFE docs seem somewhat hidden, or somewhat second-class; or at the least, harder to navigate to than it seems like they should be, given their importance. 

So, with this adjustment, I'm trying to make them easier to work with in multiple ways at once: 

- If you definitely need a document within the PTFE nav hierarchy, you have a bigger initial target, plus we saved you a whole click. 
- If you don't KNOW whether you need a document within the PTFE docs, exposing the top-level topics should make the division between app concerns and operational concerns a lot more transparent.
- If you don't know whether you need PTFE at all, the summary of topics quietly helps you evaluate that; right off the bat, I'm seeing that it would give me SAML sign-ins, some additional logging, and more control over my data security (for good and ill). 

screen shot 2018-08-01 at 5 46 05 pm

FAQ: 

- Why aren't those big headers clickable? Well, the default link styling within that sidebar made them look kind of wack. Can easily fix that in CSS, but for the proof of concept I just put "home" links underneath.

### DIFF
--- a/content/source/docs/enterprise/private/index.html.md
+++ b/content/source/docs/enterprise/private/index.html.md
@@ -1,7 +1,7 @@
 ---
 layout: "enterprise2"
 page_title: "Private Terraform Enterprise"
-sidebar_current: "docs-enterprise2-private"
+sidebar_current: "docs-enterprise2-private-home"
 ---
 
 # Private Terraform Enterprise
--- a/content/source/layouts/enterprise2.erb
+++ b/content/source/layouts/enterprise2.erb
@@ -15,6 +15,12 @@
 
       <hr>
 
+      <li><strong>Terraform Enterprise</strong></li>
+
+      <li<%= sidebar_current("docs-enterprise2-home") %>>
+        <a href="/docs/enterprise/index.html">Home</a>
+      </li>
+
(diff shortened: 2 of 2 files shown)


phinze:
I think the "nestled" PTFE docs are somewhat intentional as their audience is PTFE Site Administrators while the rest of the docs have an audience of all TFE/PTFE Users.

The difference in size between these two groups is measured in orders of magnitude (hundreds or thousands of TFE users for every one PTFE Site Administrator).

Given that, I think it's reasonable to keep the docs on how to use TFE front and center and establish the PTFE docs as sort of a "site administrator's corner".

nfagerlund:
@phinze That's a good point about readership size. 

I'm wondering now if maybe we should keep the nesting, but either add some extra text or some kind of decoration to the PTFE docs link — just calling it "Private Terraform Enterprise" sort of makes it sound like it's relevant to anyone who's using PTFE, but you're right, really it's about administering a PTFE instance, and the name/decoration probably ought to reflect that.

nfagerlund:
\*mortal kombat "finish him" voice sample\*

PR #398, opened 2018-08-08
Add docs for deleting an OAuth Client

davidcelis:
We got a request to add documentation for other OAuth Client API endpoints, and this is the only one not documented.

### DIFF
--- a/content/source/docs/enterprise/api/oauth-clients.html.md
+++ b/content/source/docs/enterprise/api/oauth-clients.html.md
@@ -109,3 +109,29 @@ curl \
   }
 }
 ```
+
+## Destroy an OAuth Client
+
+`DELETE /organizations/:organization_name/oauth-clients/:id`
+
+Parameter            | Description
+---------------------|------------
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/api/oauth-clients.html.md
```diff
@@ -109,3 +109,29 @@ curl \
...
+## Destroy an OAuth Client
+
+`DELETE /organizations/:organization_name/oauth-clients/:id`
+
+Parameter            | Description
+---------------------|------------
+`:organization_name` | The name of the organization that owns the OAuth connection.
+`:id`                | The ID of the OAuth client to destroy.
+
+This endpoint allows you to remove an existing connection between an organization and a VCS provider (GitHub, Bitbucket, or GitLab). Note that removing the OAuth Client will unlink workspaces that use this connection from their repositories, and these workspaces will need to be manually linked to another repository.
```
armchairlinguist: I'd suggest we should use `~> **Note**: Removing...` in a separate para here, to get the callout markup for the warning.

PR #402, opened 2018-08-14
Breakout section for reference architectures

roooms:
This PR separates the reference architectures from the other install content to make it easier to find. This also aligns with changes coming to the other product sites to surface the reference architectures and deployment guides more clearly in the navigation.

A future update will rename the files and introduce redirects so they are no longer called setup guides.


roooms:
I'll merge master into this once #437 is merged and then it should be good to go. #437 is more important than this one.

roooms:
Thanks, @armchairlinguist that will happen in future (already discussed with Jack and Geoffrey). As you say, it will require some link and filename changes and I'll be getting Jack to help with redirects etc to make sure we don't have search results leading to 404s.

roooms:
🎉

PR #403, opened 2018-08-16
Extend > Resources

appilon:
Adds a page in the sidebar after `writing custom providers` and `schemas`, covers the advanced features of Resources.

Preview:
http://tf-alex-preview2.s3-website-us-west-2.amazonaws.com/docs/extend/resources.html

I will squash the commits before merging

### DIFF
--- a/content/source/docs/extend/resources.html.md
+++ b/content/source/docs/extend/resources.html.md
@@ -1,0 +1,370 @@
+---
+layout: "extend"
+page_title: "Resources - Guides"
+sidebar_current: "docs-extend-resources"
+description: |-
+  Resources are a key component to provider development. This guide covers using advanced resource APIs.
+---
+
+# Resources
+
+A key component to Provider development is defining a resource. Aspects of this component have already been covered in [Writing Custom Providers](/docs/extend/writing-custom-providers.html) and [Schemas](/docs/extend/schemas/), this document will cover the other features of `schema.Resource`.
+
+## Table of Contents
+
+- [State Migrations](#state-migrations)
+- [Retry Helpers](#retry-helpers)
+  - [Retry](#retry)
+  - [StateChangeConf](#statechangeconf)
+- [Importers](#importers)
+- [Customizing Differences](#customizing-differences)
+
+## State Migrations
+
+Resources define the data types and API interactions required to create, update, and destroy infrastructure with a cloud vendor while the [Terraform state](/docs/state/index.html) stores mapping and metadata information for those remote objects. There are several reasons why a resource implementation needs to change: backend APIs Terraform interacts with will change overtime, or the current implementation might be incorrect or unmaintainable. Some of these changes may not be backward compatible and a migration is needed for resources provisioned in the wild with old schema configurations.
+
+For this task provider developers should use a resource's `SchemaVersion` and `MigrateState` function. Resources do not have these options set on first implementation, the `SchemaVersion` defaults to `0`.
+
+```go
+package example
+
+import "github.com/hashicorp/terraform/helper/schema"
+
+func resourceExampleInstance() *schema.Resource {
+    return &schema.Resource{
+        Create: resourceExampleInstanceCreate,
+        Read:   resourceExampleInstanceRead,
+        Update: resourceExampleInstanceUpdate,
+        Delete: resourceExampleInstanceDelete,
+      
+        Schema: map[string]*schema.Schema{
+            "name": {
+                Type:     schema.TypeString,
+                Required: true,
+            },
+        },
+    }
+}
+```
+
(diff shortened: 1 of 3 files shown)


content/source/docs/extend/resources.html.md
```diff
@@ -0,0 +1,368 @@
...
+        if resp.Status != "CREATED" {
+            return resource.RetryableError(fmt.Errorf("Expected instance to be created but was in state %s", resp.Status))
+        }
+
+        return resource.NonRetryableError(resourceExampleInstanceRead(d, meta))
+    })
+}
+```
+
+### StateChangeConf
```
radeksimko: I think the header should be something like `Retry Helpers`?

content/source/docs/extend/resources.html.md
```diff
@@ -0,0 +1,368 @@
...
+
+`resource.Retry` is useful for simple scenarios, but sometimes handling an APIs latency or eventual consistency requires more fine tuning. `resource.Retry` is in fact a wrapper for a another helper: `resource.StateChangeConf`. Use `resource.StateChangeConf` when your resource has several pending states, the retry timing and backoff requires fine-tuning, or you want to ensure a minimum number of occurrences of a target state is reached (this is very common when dealing with eventually consistent APIs, where a response can reply back with an old state between calls before becoming consistent).
```
radeksimko: I think we should explain the main difference between the two:

`resource.Retry` is useful for retrying on errors or more generally when dealing with **two** states (error / success).
`resource.StateChangeConf` should be preferred when the resource can go through **multiple** states as it provides more fine grained control and set expectations around what state can the resource be in.

It doesn't have to be word by word, but this is how I understand the two differ, and how I would decide which one to use.

content/source/docs/extend/resources.html.md
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    d.Set("name", obj.Name)
+    return nil
+}
+```
+
+There are other cases where the `READ` function uses configuration parameters as the identifier or support for importing multiple resources, as seen in the [AWS Provider](https://github.com/terraform-providers/terraform-provider-aws/blob/d3fe7e9907263b1aa41ddc0736a34b42899d1536/aws/import_aws_dx_gateway.go#L12) is needed. In general it is advised to stick to the passthrough importer when possible.
+
+## Customizing Differences
```
radeksimko: Not necessarily a blocker, but are we going to leave out `DiffSuppressFunc` here? I feel like it's even more commonly used than `CustomizeDiff` and probably deserves a few sentences?

appilon: DiffSuppressFunc is covered in schemas:
https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#diffsuppressfunc

This doc focuses on Resource APIs, as far as I know DiffSuppressFunc is at the schema level?

content/source/docs/extend/resources.html.md
appilon: 👍

content/source/docs/extend/resources.html.md
radeksimko: Ah, doh, you're right, I missed that.

content/source/docs/extend/resources.html.md
radeksimko: The higher-level one, sorry for the confusion. I think headers should use a human-friendly language - something the reader can understand and then we can dive into terms like `StateChangeConf` below.

content/source/docs/extend/resources.html.md
```diff
@@ -0,0 +1,368 @@
...
+        }
+
+        return resource.NonRetryableError(resourceExampleInstanceRead(d, meta))
+    })
+}
+```
+
+### StateChangeConf
+
+`resource.Retry` is useful for simple scenarios, particularly when the API response is either success or failure, but sometimes handling an APIs latency or eventual consistency requires more fine tuning. `resource.Retry` is in fact a wrapper for a another helper: `resource.StateChangeConf`. Use `resource.StateChangeConf` when your resource has multiple states to progress though, you require fine grained control of retry and delay timing, or you want to ensure a minimum number of occurrences of a target state is reached (this is very common when dealing with eventually consistent APIs, where a response can reply back with an old state between calls before becoming consistent).
```
radeksimko: Just a nitpick - I'd put _Use `resource.StateChangeConf` when _ to a separate paragraph, i.e. insert two newlines.

radeksimko (approved):
Just one nitpick/minor suggestion, but feel free to :shipit:

content/source/docs/extend/resources.html.md
appilon: 👍

PR #406, opened 2018-08-16
Migrating state from oss to remote

beekus:
Update the migrating state from oss to tfe.
screenshot2
screenshot1

### DIFF
--- a/content/source/docs/enterprise/migrate/index.html.md
+++ b/content/source/docs/enterprise/migrate/index.html.md
@@ -9,6 +9,7 @@ sidebar_current: "docs-enterprise2-migrating"
 [cli-workspaces]: /docs/state/workspaces.html
 [user-token]: ../users-teams-organizations/users.html#api-tokens
 [remote-backend]: /docs/backends/types/remote.html
+[cli-credentials]: /docs/commands/cli-config.html#credentials
 
 # Migrating State from Terraform Open Source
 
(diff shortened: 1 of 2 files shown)


content/source/docs/enterprise/migrate/index.html.md
```diff
@@ -27,74 +28,77 @@ Make sure you have all of the following:
 - A TFE user account which is a member of your organization's owners team, so you can create workspaces.
 - A [user API token][user-token] for your TFE user account. (Organization and team tokens will not work; the token must be associated with an individual user.)
 
-    In your shell, set an `ATLAS_TOKEN` environment variable with your API token as the value.
+    In your shell, set a `TFE_TOKEN` environment variable with your API token as the value.
 
     ``` bash
-    export ATLAS_TOKEN=<USER TOKEN>
+    export TFE_TOKEN=<USER TOKEN>
```
phinze: Going forward we're consolidating credential storage here https://www.terraform.io/docs/commands/cli-config.html#credentials - the ENV variable actually won't work in this case, so we should point users to the config file to set their API token.

content/source/docs/enterprise/migrate/workspaces.html.md
```diff
@@ -31,120 +31,91 @@ Make sure you have all of the following:
 - A TFE user account which is a member of your organization's owners team, so you can create workspaces.
 - A [user API token][user-token] for your TFE user account. (Organization and team tokens will not work; the token must be associated with an individual user.)
 
-    In your shell, set an `ATLAS_TOKEN` environment variable with your API token as the value.
+    In your shell, set a `TFE_TOKEN` environment variable with your API token as the value.
 
     ``` bash
-    export ATLAS_TOKEN=<USER TOKEN>
+    export TFE_TOKEN=<USER TOKEN>
```
phinze: Same as above, point users to the credentials stanza in the config file instead of an env var.

PR #408, opened 2018-08-16
extend: Best Practices: Add Deprecations, Removals, and Renames guide

bflad:
The synopsis says it all. 😄 

Over time, remote services evolve and better workflows are designed.
Terraform's plugin system has functionality to aid in iterative improvements.
In `Deprecations, Removals, and Renames`, we cover procedures for
backwards compatible code and documentation updates to ensure that operators
are well informed of changes ahead of functionality being removed or renamed.

### DIFF
--- a/content/source/docs/extend/best-practices/deprecations.md
+++ b/content/source/docs/extend/best-practices/deprecations.md
@@ -1,0 +1,737 @@
+---
+layout: "extend"
+page_title: "Extending Terraform: Deprecations, Removals, and Renames Best Practices"
+sidebar_current: "docs-extend-best-practices-deprecations"
+description: |-
+  Recommendations for deprecations, removals, and renames.
+---
+
+# Deprecations, Removals, and Renames
+
+Terraform is trusted for managing many facets of infrastructure across many organizations. Part of that trust is due to consistent versioning guidelines and setting expectations for various levels of upgrades. Ensuring backwards compatibility for all patch and minor releases, potentially in concert with any upcoming major changes, is recommended and supported by the Terraform development framework. This allows operators to iteratively update their Terraform configurations rather than require massive refactoring.
+
+This guide is designed to walk through various scenarios where existing Terraform functionality requires future removal, while maintaining backwards compatibility. Further information about the versioning terminology (e.g. `MAJOR`.`MINOR`.`PATCH`) in this guide can be found in [the versioning guidelines documentation](/docs/extend/best-practices/versioning.html).
+
+~> **NOTE:** Removals should only ever occur in `MAJOR` version upgrades.
+
+## Table of Contents
+
+- [Provider Attribute Removal](#provider-attribute-removal)
+- [Provider Attribute Rename](#provider-attribute-rename)
+  - [Renaming a Required Attribute](#renaming-a-required-attribute)
+  - [Renaming an Optional Attribute](#renaming-an-optional-attribute)
+  - [Renaming a Computed Attribute](#renaming-a-computed-attribute)
+- [Provider Data Source or Resource Removal](#provider-data-source-or-resource-removal)
+- [Provider Data Source or Resource Rename](#provider-data-source-or-resource-rename)
+
+## Provider Attribute Removal
+
+The recommended process for removing an attribute from a data source or resource in a provider is as follows:
+
+1. Add `Deprecated` in the attribute schema definition. After an operator upgrades to this version, they will be shown a warning with the message provided when using the attribute, but the Terraform run will still complete.
+1. Ensure the changelog has an entry noting the deprecation.
+1. Release a `MINOR` version with the deprecation.
+1. In the next `MAJOR` version, remove all code associated with the attribute except for its schema definition.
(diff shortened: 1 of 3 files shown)


content/source/docs/extend/best-practices/deprecations.md
```diff
@@ -0,0 +1,737 @@
...
+- Allows the operator to migrate the configuration to the new attribute at the same time requiring that any other references only work with the new attribute. This is to prevent a situation with Terraform showing a difference when the existing attribute is configured, but the new attribute is saved into the Terraform state. For example, in `terraform plan` output format:
+
+```
+existing_attribute: "" => "value"
+new_attribute:      "value" => ""
+```
+
+The recommended process is as follows:
+
+1. Replace `Required: true` with `Optional: true` in the existing attribute schema definition.
```
appilon: numbering got messed up, everything step 1!

content/source/docs/extend/best-practices/deprecations.md
```diff
@@ -0,0 +1,737 @@
...
+In general, the procedure here allows the operator to migrate the configuration to the new attribute at the same time requiring that any other references only work with the new attribute. This is to prevent a situation with Terraform showing a difference when the existing attribute is configured, but the new attribute is saved into the Terraform state. For example, in `terraform plan` output format:
+
+```text
+existing_attribute: "" => "value"
+new_attribute:      "value" => ""
+```
+
+The recommended process is as follows:
+
+1. Duplicate the schema definition of the existing attribute, renaming one of them with the new attribute name.
```
appilon: numbering

content/source/docs/extend/best-practices/deprecations.md
```diff
@@ -0,0 +1,737 @@
...
+}
+```
+
+### Renaming a Computed Attribute
+
+~> **NOTE:** If the schema definition contains `Optional` see the [Renaming an Optional Attribute section](#renaming-an-optional-attribute) instead. If the schema definition contains `Required` see the [Renaming a Required Attribute section](#renaming-a-required-attribute) instead.
+
+The recommended process is as follows:
+
+1. Duplicate the schema definition of the existing attribute, renaming one of them with the new attribute name.
```
appilon: numbering

content/source/docs/extend/best-practices/deprecations.md
```diff
@@ -0,0 +1,737 @@
...
+    // ... other logic ...
+    return nil
+}
+```
+
+## Provider Data Source or Resource Removal
+
+The recommended process for removing a data source or resource from a provider is as follows:
+
+1. Add `DeprecationMessage` in the data source or resource schema definition. After an operator upgrades to this version, they will be shown a warning with the message provided when using the deprecated data source or resource, but the Terraform run will still complete.
```
appilon: numbering

content/source/docs/extend/best-practices/deprecations.md
```diff
@@ -0,0 +1,737 @@
...
+}
+```
+
+## Provider Data Source or Resource Rename
+
+When renaming a resource from one name to another, it is important to keep backwards compatibility with both existing Terraform configurations and the Terraform state while operators migrate. To accomplish this, there will be some duplicated logic to support both resources until the next `MAJOR` release. Once both resources are appropriately handled, the process for deprecating and removing the old resource is the same as noted in the [Provider Data Source or Resource Removal section](#provider-data-source-or-resource-removal).
+
+The recommended process is as follows:
+
+1. Duplicate the code of the existing resource, renaming (and potentially modifying) functions as necessary.
```
appilon: numbering

content/source/docs/extend/best-practices/index.html.md
```diff
@@ -46,5 +54,6 @@ Terraform would expect a resource to behave.
 
 [Click here to learn more.][drift]
```
appilon: can you delete this line for me? I don't think its needed and it's confusingly placed.

bflad: I need to add a synopsis for the versioning page -- let me address this as well in that upcoming PR after this one. 👍

bflad:
@appilon according to the [Markdown syntax for lists](https://daringfireball.net/projects/markdown/syntax#list)

> It’s important to note that the actual numbers you use to mark the list have no effect on the HTML output Markdown produces. The HTML Markdown produces from the above list is:
> 
> ```
> <ol>
> <li>Bird</li>
> <li>McHale</li>
> <li>Parish</li>
> </ol>
> ```
> 
> If you instead wrote the list in Markdown like this:
>
> ```
> 1.  Bird
> 1.  McHale
> 1.  Parish
> ```
> 
> or even:
> 
> ```
> 3. Bird
> 1. McHale
> 8. Parish
> ```
>
> you’d get the exact same HTML output.

You can verify this with `make website`:

screen shot 2018-08-17 at 12 43 58 pm

Personally, I prefer the repeated `1.` so you do not need to constantly reorder all the numbers when the list is changed, but if there is some precedent or guideline for appropriately numbering a list in the actual Markdown, I can change these. 👍

content/source/docs/extend/best-practices/deprecations.md
appilon: :( I'll go away now...

PR #410, opened 2018-08-17
sentinel: Update documentation to account for new data namespace

vancluever:
This updates the documentation to reflect the new data source namespace,
which will be available in TFE soon.

The namespace will be available in all 3 imports and is implemented in
the same way resources are. Specifically, this is actually a splitting
of the previously shared "resources" namespace, which will now only
contain resources going forward.

### DIFF
--- a/content/source/docs/enterprise/sentinel/import/tfstate.html.md
+++ b/content/source/docs/enterprise/sentinel/import/tfstate.html.md
@@ -72,8 +72,8 @@ from name to data source. For the documentation on the value, see
 
 This returns a list of data sources and behaves exactly how a resource lookup
 behaves with the m.resources.TYPE.NAME syntax above. All fields that are
-accessible within resources are available within a data source, save `tainted`,
-as data sources cannot be tainted.
+accessible within resources are available within a data source, except for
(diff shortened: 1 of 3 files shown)


content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -52,6 +57,24 @@ This returns the list of resources with the given type and name. The
...
+
+This returns all the data sources in the module with the given type as a map
+from name to data source. For the documentation on the value, see
+`m.data.TYPE.NAME`
+
+### m.data.TYPE.NAME
+
+This returns a list of data sources and behaves exactly how a resource lookup
+behaves with the m.resources.TYPE.NAME syntax above. All fields that are
+accessible within resources are available within a data source, save `tainted`,
```
armchairlinguist: I might use "except for" instead of "save", it's less poetic but more common. :)

vancluever: Updated!

vancluever:
Thanks @armchairlinguist! Merging this now and it will be available in the next website release.

PR #414, opened 2018-08-20
Add new attributes to Plans API

davidcelis:
These are now live and reflect the number of resources that a Terraform run would add, change, and destroy.


nfagerlund:
👍🏼

PR #421, opened 2018-08-22
Amy hashi ptfe vmguide

amy-hashi:


content/source/docs/enterprise/private/vmware-setup-guide.html.md
```diff
@@ -30,13 +30,12 @@ Depending on the chosen [operational
 mode](https://www.terraform.io/docs/enterprise/private/install-installer.html#operational-mode-decision),
 the infrastructure requirements for PTFE range from a single virtual machine
 for demo or proof of concept installations, to multiple virtual machines
-hosting postgres, access to S3 or Azure storage, and external Vault servers for
+hosting postgres, access to S3, Azure or Google Cloud storage, and external Vault servers for
```
blalor: This parses kind of weird for me.  Maybe a comma after "Azure", but also, they don't need a VM for "access to S3"; they'd either need the VM to _have_ access, or to provide their own storage service, like ceph.

content/source/docs/enterprise/private/vmware-setup-guide.html.md
```diff
@@ -92,14 +91,16 @@ the internal database or vault may result in serious performance issues.
...
-application. This S3 bucket must be access via the network to the PTFE virtual
+application. This object storage must be accessible via the network to the PTFE virtual
 machine as well as the postgres server. Vault is used to encrypt all
-application data stored in the S3 bucket. This allows for further [server-side
+application data stored in this location. This allows for further [server-side
 encryption](https://docs.aws.amazon.com/AmazonS3/latest/dev/serv-side-encryption.html)
-by S3 if required by your security policy.
+by S3 if required by your security policy. 
+
+S3-compatible storage can be Google Cloud storage, Azure blob storage, or [ceph](https://ceph.com/), among many others. Please feel free to reach out to [support](https://www.hashicorp.com/support) with questions.
```
blalor: This is somewhat incorrect.  We have first-class support for S3, GCS, and Azure's storage, or you can use an S3-_compatible_ service such as ceph.  If you can use the first-class storage service support, you should.  In fact, Google's S3-compatibility is pretty marginal, from what I've heard.

content/source/docs/enterprise/private/vmware-setup-guide.html.md
```diff
@@ -141,16 +142,16 @@ The above diagram show the infrastructure components at a high-level.
 ### Storage Layer
 
 The Storage Layer is composed of multiple service endpoints (datastores,
-S3, Vault) all configured with or benefitting from inherent resiliency
-provided by ESX and AWS (in the case of S3) or assumed resiliency
+object storage, Vault) all configured with or benefitting from inherent resiliency
+provided by ESX and your storage provider, or assumed resiliency
```
blalor: "assumed resiliency" makes me giggle.

content/source/docs/enterprise/private/vmware-setup-guide.html.md
amy-hashi: Yeah... that is odd, actually... maybe "to multiple virtual machines hosting the Terraform Enterprise application, postgres and external Vault servers for..." ?

content/source/docs/enterprise/private/vmware-setup-guide.html.md
amy-hashi: Maybe I could alter that to read "Supported storage options include AWS S3, Azure, Google Cloud Storage, and ceph, among others." ?

content/source/docs/enterprise/private/vmware-setup-guide.html.md
blalor: yeah.  Also, I think @armchairlinguist  mentioned this earlier, but we should probably use "PostgreSQL", as well.

content/source/docs/enterprise/private/vmware-setup-guide.html.md
blalor: Yes, but I'm hesitant to say we "support" ceph.  I'm not entirely sure we should remove the minio reference, as well.  The lack of versioning in minio is important, but I'm not really comfortable implying ceph is the only solution.  I'm also not sure we call out the need (well, desire) for versioning outside of the S3 bucket setup…

content/source/docs/enterprise/private/vmware-setup-guide.html.md
```diff
@@ -92,14 +91,16 @@ the internal database or vault may result in serious performance issues.
...
 
-An [S3 Standard](https://aws.amazon.com/s3/storage-classes/) bucket must be
+An [S3 Standard](https://aws.amazon.com/s3/storage-classes/) bucket, or compatible storage, must be
 specified during the PTFE installation for application data to be stored
 securely and redundantly away from the virtual servers running the PTFE
-application. This S3 bucket must be access via the network to the PTFE virtual
-machine as well as the postgres server. Vault is used to encrypt all
-application data stored in the S3 bucket. This allows for further [server-side
+application. This object storage must be accessible via the network to the PTFE virtual
+machine as well as the PostgreSQL server. Vault is used to encrypt all
```
blalor: "as well as the PostgreSQL server": postgresql would never talk to the object store on its own.

PR #428, opened 2018-08-27
Update the default pagination size to 150

beekus:
Change default pagination to 150


beekus:
@nfagerlund i made a change because we are now only defaulting the workspaces index endpoint to 150. everything else defaults back to 20.

PR #429, opened 2018-08-27
Correct workspace lock description

chrisarcand:
There's a contradiction on this page; previously it says that users with write access can lock/unlock workspaces. I've confirmed that write users can indeed lock/unlock workspaces.


nfagerlund:
Yeah, that matches my current understanding as well. Thanks for catching this!

PR #433, opened 2018-08-27
Update tfconfig.html.md

philbal611:
Updating doc to provide more useful and accurate information on tfconfig import.

- tfconfig.modules does not exist. description matches tfconfig.module_paths
- types 'm', 'p', 'r', 'd', and 'v' represent 'module', 'providers', 'resources', 'data', and 'variables'

### DIFF
--- a/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
@@ -20,7 +20,7 @@ providers, resources, data sources, modules, variables, etc. Note that since
 this is the configuration and not an invocation of Terraform, you can't see
 values for variables, state, etc.
 
-### tfconfig.modules
+### tfconfig.module_paths
 
 All the module paths represented in the config. This can be used along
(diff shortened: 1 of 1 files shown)


vancluever:
Hey @philbal611, thank you very much for this pull request!

We very much understand that the current documentation for the TFE Sentinel functionality is, to say the least, not helpful, and we appreciate the effort to try and make it better!

There were some things missed here with the updates that still don't necessarily communicate the correct method of use for certain features. As an example, `module` is a function, so a lot of the items that reference it will not work in the examples shown - as an example, `module.resources` would need to be `module([]).resources`.

After discussing internally, we've decided that we are going to work towards a larger overhaul of the documentation, sooner rather than later, and ultimately this will mean that the docs will look a lot different than they do today - with not just abbreviations corrected, but with a completely different format, so that the imports are documented more to a level of quality of our OSS providers.

As such, I'm just going to close this now. Again, thank you for making the effort - and check back soon for a much better documentation layout!

philbal611:
Awesome, that sounds great. Thanks for getting back!

PR #435, opened 2018-08-28
Fix usage attributes in /admin/terraform-versions API

nfagerlund:
Usage counts are provided in each object, not in the meta fields. Confirmed by
hitting the API on my zone (release d2a7cb2).


davidcelis:
Pinging @caseylang since I think he's the one who made this change

nfagerlund:
👍🏼

PR #440, opened 2018-08-29
Add force-cancel API docs

ryanuber:
Added force-cancel API docs for site admin as well as the runs API. I also added a missing endpoint (`GET /runs/:run_id`) because I wanted to link to it. I noticed a few other little JSON errors in here too so I fixed those up.

### DIFF
--- a/content/source/docs/enterprise/api/admin/runs.html.md
+++ b/content/source/docs/enterprise/api/admin/runs.html.md
@@ -143,6 +143,22 @@ Status  | Response                               | Reason
 [JSON API document]: https://www.terraform.io/docs/enterprise/api/index.html#json-api-documents
 [JSON API error object]: http://jsonapi.org/format/#error-objects
 
+### Request body
+
+This POST endpoint allows an optional JSON object with the following properties as a request payload.
+
+Key path  | Type   | Default | Description
+----------|--------|---------|------------
+`comment` | string | `null`  | An optional explanation for why the run was force canceled.
+
+### Sample Payload
+
+```json
+{
+  "comment": "This run was stuck and would never finish."
+}
+```
+
 ### Sample Request
 
 ```shell
@@ -150,6 +166,7 @@ curl \
   --header "Authorization: Bearer $TOKEN" \
   --header "Content-Type: application/vnd.api+json" \
   --request POST \
+  --data @payload.json \
   "https://app.terraform.io/api/v2/admin/runs/run-VCsNJXa59eUza53R/actions/force-cancel"
 ```
 
--- a/content/source/docs/enterprise/api/admin/runs.html.md
+++ b/content/source/docs/enterprise/api/admin/runs.html.md
@@ -125,13 +125,13 @@ curl \
 
 ## Force a run into the "cancelled" state
 
-`POST /admin/runs/:id/actions/cancel`
+`POST /admin/runs/:id/actions/force-cancel`
 
 Parameter | Description
 ----------|------------
 `:id`     | The ID of the run to cancel.
 
-This endpoint forces a run (and its plan/apply, if applicable) into the `"errored"` state. This action should only be performed for runs that are stuck and no longer progressing normally, as there is a risk of lost state data if a progressing apply is force-canceled. Healthy runs can be [requested for cancellation by end-users](../../run/states.html).
+This endpoint forces a run (and its plan/apply, if applicable) into the `"canceled"` state. This action should only be performed for runs that are stuck and no longer progressing normally, as there is a risk of lost state data if a progressing apply is force-canceled. Healthy runs can be [requested for cancellation by end-users](../../run/states.html).
 
 Status  | Response                               | Reason
 --------|----------------------------------------|----------
@@ -150,7 +150,7 @@ curl \
   --header "Authorization: Bearer $TOKEN" \
   --header "Content-Type: application/vnd.api+json" \
   --request POST \
-  "https://app.terraform.io/api/v2/admin/runs/run-VCsNJXa59eUza53R/actions/cancel"
+  "https://app.terraform.io/api/v2/admin/runs/run-VCsNJXa59eUza53R/actions/force-cancel"
 ```
 
 ### Sample Response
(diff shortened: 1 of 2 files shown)


content/source/docs/enterprise/api/run.html.md
```diff
@@ -277,7 +279,8 @@ curl \
         "run-events": {...},
         "policy-checks": {...},
         "comments": {...},
-        "workspace-run-alerts": {...},
+        "workspace-run-alerts": {...}
```
nfagerlund: good catch.

content/source/docs/enterprise/api/run.html.md
```diff
@@ -346,14 +428,66 @@ Parameter | Description
...
+
+`POST /runs/:run_id/actions/force-cancel`
+
+Parameter | Description
+----------|------------
+`run_id`  | The run ID to cancel
+
+The `force-cancel` action is like [cancel](#cancel-a-run), but ends the run immediately. Once invoked, the run is placed into a `canceled` state, and the running Terraform process is terminated. The workspace is immediately unlocked, allowing further runs to be queued.
+
+This endpoint enforces a prerequisite that a [non-forceful cancel](#cancel-a-run) is performed first. Once a non-forceful cancel has been performed, an additional cool-off period is enforced. The time remaining in the cool-off period can be found using the [run details endpoint](#get-run-details), in the key `force_cancel_delay_seconds`. Once this time period expires, if the run has still not ended, the force-cancel endpoint may be used successfully.
```
nfagerlund: User-brain questions: 

* What's the full path to this key? Is it `data.attributes.force_cancel_delay_seconds`, or is it inside a sub-object of `attributes`?
* I don't see the delay key in the sample run objects; does that mean it's only present in the run object while the cooldown delay is in effect? 
* Is the client meant to use the `permissions` and `actions` objects to determine whether force-cancel is an option? How does that work, and does it interact with the cooldown timer?

content/source/docs/enterprise/api/run.html.md
```diff
@@ -346,14 +428,66 @@ Parameter | Description
...
+
+This endpoint represents an action as opposed to a resource. As such, it does not return any object in the response body.
+
+-> **Note:** This endpoint cannot be accessed with [organization tokens](../users-teams-organizations/service-accounts.html#organization-service-accounts). You must access it with a [user token](../users-teams-organizations/users.html#api-tokens) or [team token](../users-teams-organizations/service-accounts.html#team-service-accounts).
```
nfagerlund: Can anyone with write access on the workspace force-cancel, or does it require admin or something?

content/source/docs/enterprise/api/run.html.md
ryanuber: Good questions, and I can try to clarify in the language here in the docs!

* Your full path is correct (`data.attributes.force_cancel_delay_seconds`).
* The key is indeed only present in the run output if the run has been canceled previously. If the run has not yet been canceled, there is no cool-off timer in the payload. After cool-off has expired the value will remain `0` after that.
* Permissions and actions can totally be used to figure out whether your user is allowed to force cancel, and if the run is in a state that would allow that to be successful. I can add that to the docs to clarify.

nfagerlund: 🎉 thanks!

content/source/docs/enterprise/api/run.html.md
```diff
@@ -346,14 +428,66 @@ Parameter | Description
...
 This endpoint queues the request to perform a cancel; the cancel might not happen immediately. After canceling, the run is completed and later runs can proceed.
 
 This endpoint represents an action as opposed to a resource. As such, it does not return any object in the response body.
 
 -> **Note:** This endpoint cannot be accessed with [organization tokens](../users-teams-organizations/service-accounts.html#organization-service-accounts). You must access it with a [user token](../users-teams-organizations/users.html#api-tokens) or [team token](../users-teams-organizations/service-accounts.html#team-service-accounts).
 
+Status  | Response                  | Reason(s)
+--------|---------------------------|----------
+[202][] | none                      | Successfully queued a cancel request.
+[409][] | [JSON API error object][] | Run was not planning or applying, has not been canceled non-forcefully, or the cool-off period has not yet passed.
```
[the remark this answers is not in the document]
ryanuber: Yikes, my bad. This should have gone into the other section. Fixing that.

ryanuber:
Took another swipe at this - let me know what you think.

content/source/docs/enterprise/api/admin/runs.html.md
```diff
@@ -143,14 +143,31 @@ Status  | Response                               | Reason
 [JSON API document]: https://www.terraform.io/docs/enterprise/api/index.html#json-api-documents
 [JSON API error object]: http://jsonapi.org/format/#error-objects
 
+### Request body
+
+This POST endpoint allows an optional JSON object with the following properties as a request payload.
+
+Key path  | Type   | Default | Description
+----------|--------|---------|------------
+`comment` | string | `null`  | An optional explanation for why the run was force canceled.
```
armchairlinguist: probably should have hyphen in force-canceled

content/source/docs/enterprise/api/run.html.md
```diff
@@ -388,6 +470,58 @@ curl \
...
+This endpoint represents an action as opposed to a resource. As such, it does not return any object in the response body.
+
+-> **Note:** This endpoint cannot be accessed with [organization tokens](../users-teams-organizations/service-accounts.html#organization-service-accounts). You must access it with a [user token](../users-teams-organizations/users.html#api-tokens) or [team token](../users-teams-organizations/service-accounts.html#team-service-accounts).
+
+~> **Warning:** This endpoint has potentially dangerous side-effects, including loss of any in-flight state in the running Terraform process. Use this operation with extreme caution.
+
+Status  | Response                  | Reason(s)
+--------|---------------------------|----------
+[202][] | none                      | Successfully queued a cancel request.
+[409][] | [JSON API error object][] | Run was not planning or applying, has not been canceled non-forcefully, or the cool-off period has not yet passed.
```
armchairlinguist: Should we have 404 here as well for permissions failure? (Is that what would happen?)

ryanuber: It is - I noticed it was not documented in the cancel endpoint, so figured that permissions are probably talked about as a concept some place, and that 404 would have been mentioned there. 404 could be permissions but it could also be that the run was not found. Grepping around it looks like that's not the case and other endpoints do mention 404 so I'll just add it.

PR #444, opened 2018-08-31
Adding back force-cancel docs

ryanuber:
These were already reviewed/merged in #440 , but I reverted it until the feature flag is lifted. We can merge this once that happens.


armchairlinguist:
@ryanuber Will these need to be updated to handle the change to the cooldown time format?

ryanuber:
@armchairlinguist yes that is on my todo list! Should be a pretty minor update but will take care of that today.

ryanuber:
Functionality is in prod for all users now

PR #445, opened 2018-09-01
Fix and expand docs about users and organizations

nfagerlund:
I have some similar edits for the teams page in progress, but wasn't able to finish those before leaving for the week, so here's what I've got. 

These are all remedial updates instead of upcoming stuff, so if there isn't anything I need to fix, I'm fine with it being merged in my absence. Otherwise, I'll catch any revisions when I come back on the 10th.


content/source/docs/enterprise/users-teams-organizations/organizations.html.md
```diff
@@ -4,14 +4,82 @@ page_title: "Organizations - Terraform Enterprise"
...
+
+Once you have created an organization, you can add other [users][] by adding them to one or more [teams][].
+
+-> **Note:** On the SaaS version of TFE, any user can create a new organization. On private installs of TFE, the administrators can restrict this ability, so that only site admins can create organizations. See [Administration: General Settings](../private/admin/general.html#organization-creation) for more details.
+
+## Organization Settings
+
+-> **API:** See the [Organizations API](../api/organizations.html).
+
+You can view and manage an organization's settings by clicking the main "Settings" link in the top navigation bar.
```
armchairlinguist: I'd probably delete main here, it doesn't really provide any additional distinction.

content/source/docs/enterprise/users-teams-organizations/organizations.html.md
```diff
@@ -4,14 +4,82 @@ page_title: "Organizations - Terraform Enterprise"
...
+
+Only members of the organization's "owners" team can manage an organization's settings. However, other users can use this section to view the organization's contact email, view the membership of any teams they belong to, and view the organization's authentication policy.
+
+Most of the organization settings are documented near the specific workflows they enable. What follows is a brief summary with links to more relevant sections of the documentation.
+
+### Profile / Delete
+
+The profile page shows the organization's name and contact email, but does not allow you to change them.
+
+The profile page is also where you can **delete your organization.**
```
armchairlinguist: Should the period be outside the bold here?

content/source/docs/enterprise/users-teams-organizations/organizations.html.md
```diff
@@ -4,14 +4,82 @@ page_title: "Organizations - Terraform Enterprise"
...
+
+### Teams
+
+The teams page allows organization owners to manage the organization's teams, including creating and delete teams, managing membership, and generating team API tokens. See [Teams][] for more information.
```
armchairlinguist: Should we mention it's where others see their membership?

content/source/docs/enterprise/users-teams-organizations/users.html.md
```diff
@@ -4,47 +4,85 @@ page_title: "Users - Terraform Enterprise"
...
-Once on this page, you can change your password and generate or revoke API tokens.
+Once on this page, can use the navigation on the left to choose which settings to manage.
 
 ![The user settings menu item, in the upper right menu.](./images/user-settings.png)
 
+### Profile
+
+TFE user profiles are very small, consisting only of a username and an email address. You can change either of these from the "Profile" page of the user settings.
+
+~> **Important:** Changing your username can cause important operations to fail. This is because it is used in URL paths to various resources. If external systems make requests to these resources, you'll need to update them prior to making a change.
```
armchairlinguist: I'm not sure this is true anymore, actually. We should get this confirmed.

content/source/docs/enterprise/users-teams-organizations/users.html.md
```diff
@@ -4,47 +4,85 @@ page_title: "Users - Terraform Enterprise"
...
+
+Additionally, TFE uses [Gravatar](http://en.gravatar.com) to display a user icon if you have associated one with your email address. For details about changing your user icon, see [Gravatar's documentation](http://en.gravatar.com/support/).
+
 ### Password Management
 
-Users manage their own passwords. To change your password, click the "Security" tab of the user settings page. You'll need to confirm your current password, and enter your new password twice.
+Users manage their own passwords. To change your password, click the "Password" page of the user settings. You'll need to confirm your current password, and enter your new password twice.
 
 Password management isn't available if your organization uses [SAML single sign on](../saml/index.html).
```
armchairlinguist: We probably shouldn't use "organization" here (although I see we were before) because it's overloading a term. I'd say "If your TFE installation uses".

content/source/docs/enterprise/users-teams-organizations/organizations.html.md
```diff
@@ -4,14 +4,82 @@ page_title: "Organizations - Terraform Enterprise"
...
+TFE is a paid product offered to organizations on a subscription basis; for more information about billing, please speak to a HashiCorp sales representative. TFE is also available as a free trial to organizations evaluating its features.
+
+When a free trial has expired, or when paid billing has been canceled, an organization becomes inactive. Inactive organizations display a banner reading "TRIAL EXPIRED — Upgrade Required" in the top navigation bar:
+
+![screenshot: the trial expiration banner](./images/org-inactive.png)
 
-If you navigate to TFE before you're a member of any organizations, you'll be asked to create one. You can also create a new organization at any time by going to an existing organization's settings, then clicking the "Create new organization" button in the sidebar navigation.
+Members of an inactive organization can still log into TFE, view data about past runs, manage team memberships, and manage access permissions on workspaces. However, inactive organizations cannot initiate new [Terraform runs](../run/index.html).
 
-Invite users to join an organization by sending them the
```
armchairlinguist: I think we should leave a link here to the new home for this info, since org owners commonly wonder how to "invite" people.

content/source/docs/enterprise/users-teams-organizations/users.html.md
armchairlinguist: @hashicorp/tfe-rails-reviewers Would one of you be able to confirm whether changing one's username in TFE still incurs the risk of breaking other systems? As I understand it, that should no longer be the case in V2.

content/source/docs/enterprise/users-teams-organizations/organizations.html.md
nfagerlund: The way I learned it, you should usually extend a text weight/style to include an adjacent punctuation mark; that's because a roman mark next to italic text can end up in slightly the wrong horizontal position, and a normal mark next to bold text can look glitchy or end up absorbed by a letterform, depending on the specific font involved. (Cases where the text style carries very very specific semantic meaning, like code spans, get an exception, kind of like the exception for bringing punctuation inside of quotation marks, which is falling out of favor regardless, but hey.)

So if this was a case where the bold style indicated a UI label, then yes; otherwise (by traditional typesetting rules, _I think_) no... but, none of our readers will really care either way, so I can do it the other way if I gotta. 😄

content/source/docs/enterprise/users-teams-organizations/organizations.html.md
nfagerlund: yes!

content/source/docs/enterprise/users-teams-organizations/users.html.md
nfagerlund: Oh heck, you're right, it slipped my mind that you can't configure that per-org.

content/source/docs/enterprise/users-teams-organizations/organizations.html.md
nfagerlund: Got lost in the shuffle somehow, but yeah, definitely. I added an "adding users to orgs" header.

content/source/docs/enterprise/users-teams-organizations/organizations.html.md
armchairlinguist: Woo, I learned something! I can totally see that. Since the text here is generic vs exact, I'm happy with your reasoning.

content/source/docs/enterprise/users-teams-organizations/users.html.md
armchairlinguist: You can't edit your username if SAML is enabled, so that part is okay. :)
But yeah, I think you're probably right to stick to the warning here for now.

PR #448, opened 2018-09-05
Add TFE<->Sentinel policy examples

chrisarcand:
This adds direct links to the terraform-guides repository on GitHub.  The links are intentionally set to the file path and not by commit SHA (as you should normally do when referencing things like this) to not create another step to updating the website docs when the examples in the repo are updated.  In the future we may want to do more of an integrated thing to show the examples inline on the site, but this is a good way to get give the extremely helpful terraform-guides docs some much needed visibility, quickly.

### DIFF
--- a/content/source/layouts/enterprise2.erb
+++ b/content/source/layouts/enterprise2.erb
@@ -194,6 +194,9 @@
               </li>
             </ul>
           </li>
+          <li<%= sidebar_current("docs-enterprise2-sentinel-examples") %>>
+            <a href="/docs/enterprise/sentinel/examples.html">Example Policies</a>
+          </li>
         </ul>
       </li>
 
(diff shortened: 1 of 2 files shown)


content/source/layouts/enterprise2.erb
```diff
@@ -194,6 +194,9 @@
               </li>
             </ul>
           </li>
+          <li<%= sidebar_current("docs-enterprise2-sentinel-examples") %>>
```
dylanegan: That syntax highlighting. 🥇

chrisarcand:
Ah ha, just noticed the review request for @skierkowski; Lemme know if you want anything changed or added in there, Maciej!

nfagerlund:
Well this came in while I was gone, but here's my post-facto 👍🏼 🙌🏼 🙌🏼 🎉

PR #450, opened 2018-09-08
sentinel: Re-write import documentation

vancluever:
This vastly improves the documentation for all three Terraform Sentinel imports: `tfconfig`, `tfplan`, and `tfstate`.

Some specific features of the rewrite include:

- Much more verbose explaniations for every part of the namespace.
- A map of each import's namespace after the synopsis.
- Examples for every value and function in the import namespace.

There will be other things to follow (namely mocks), but this is complete enough to be merged in now, the mocks can be merged in later.

### DIFF
--- a/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
@@ -195,12 +195,12 @@ for the root module. They are essentially the equivalent of running
 
 ## Namespace: Resources/Data Sources
 
-The **resource namespace** is a namespace shared between resources and data
-sources, and is identical in structure and behavior regardless of which
-classification you are loading.
+The **resource namespace** is a namespace _type_ that applies to both resources
+(accessed by using the `resources` namespace key) and data sources (accessed
+using the `data` namespace key).
 
-Accessing an individual resource or data source within this namespace can be
-accomplished by specifying the type and name, in the syntax
+Accessing an individual resource or data source within each individual namespace
+can be accomplished by specifying the type and name, in the syntax
 `[resources|data].TYPE.NAME`, depending on whether or not you are loading the
 namespace for a resource (accessed in the `resources` name space) or data
 sources (accessed in the `data`) namespace.
--- a/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
@@ -183,7 +183,6 @@ get_vms = func() {
 }
 ```
 
-
 ## Namespace: Module
 
 The **module namespace** can be loaded by calling [`module()`](#function-module-)
--- a/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
@@ -208,8 +208,7 @@ It can be used to load the following child namespaces:
 
 The root-level `data`, `modules`, `providers`, `resources`, and `variables` keys
 all alias to their corresponding namespaces within the module namespace, loaded
-for the root module. They are essentially the equivalent of running
-`module([]).KEY`.
+for the root module. They are the equivalent of running `module([]).KEY`.
 
 ## Namespace: Resources/Data Sources
 
--- a/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
@@ -216,16 +216,20 @@ The **resource namespace** is a namespace _type_ that applies to both resources
 (accessed by using the `resources` namespace key) and data sources (accessed
 using the `data` namespace key).
 
-Accessing an individual resource or data source within each individual namespace
+Accessing an individual resource or data source within each respective namespace
 can be accomplished by specifying the type and name, in the syntax
-`[resources|data].TYPE.NAME`, depending on whether or not you are loading the
-namespace for a resource (accessed in the `resources` name space) or data
-sources (accessed in the `data`) namespace.
+`[resources|data].TYPE.NAME`.
 
-In addition, each of these namespace levels is a map, so to access all items
-_matching_ a certain type, you can use `[resources|data].TYPE`, and to access
-all data sources _grouped by_ type, you can use a simple bareword `resources` or
-`data` map.
+In addition, each of these namespace levels is a map, allowing you to filter
+based on type and name. Some examples of multi-level access are below:
+
+* To fetch all `aws_instance` resources within the root module, you can specify
+  `tfconfig.resources.aws_instance`. This would give you a map of resource
+  namespaces indexed off of the names of each resource (`foo`, `bar`, and so
+  on).
+* To fetch all resources within the root module, irrespective of type, use
+  `tfconfig.resources`. This is indexed by type, as shown above with
+  `tfconfig.resources.aws_instance`, with names being the next level down.
 
 Further explanation of the namespace will be in the context of resources. As
 mentioned, when operating on data sources, use the same syntax, except with
--- a/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
@@ -428,6 +428,17 @@ This namespace is indexed by provider type and _only_ contains data about
 providers when actually declared. If you are using a completely implicit
 provider configuration, this namespace will be empty.
 
+This namespace is populated based on the following criteria:
+
+* The top-level namespace [`config`](#value-config-3) and
+  [`version`](#value-version) values are populated with the configuration and
+  version information from the default provider (the provider declaration that
+  lacks an alias).
+* Any aliased providers are added as namespaces within the
+  [`alias`](value-alias) value.
+* If a module lacks a default provider configuration, the top-level `config` and
+  `version` values will be empty.
+
 ### Value: `alias`
 
 * **Value Type:** A map of [provider namespaces](#namespace-providers), indexed
--- a/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
@@ -101,6 +101,11 @@ Hence, a module with an address of simply `foo` (or `root.foo`) would be
 `["foo"]`, and a module within that (so address `foo.bar`) would be read as
 `["foo", "bar"]`.
 
+[`null`][ref-null] is returned if a module address is invalid, or if the module
+is not present in the configuration.
+
+[ref-null]: https://docs.hashicorp.com/sentinel/language/spec#null
+
 As an example, given the following module block:
 
 ```hcl
@@ -133,9 +138,13 @@ main = rule { subject.module(["foo"]).resources.null_resource.foo.config.trigger
 * **Value Type:** List of a list of strings.
 
 The `module_paths` value within the [root namespace](#namespace-root) is a list
-of all of the modules within the Terraform configuration. This data is
-represented as a list of a list of strings, with the inner list being the module
-address, split on the period (`.`).
+of all of the modules within the Terraform configuration.
+
+Modules not present in the configuration will not be present here, even if they
+are present in the diff or state.
+
+This data is represented as a list of a list of strings, with the inner list
+being the module address, split on the period (`.`).
 
 The root module is included in this list, represented as an empty inner list.
 
--- a/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
@@ -107,8 +107,10 @@ module "foo" {
 If the module contained the following content:
 
 ```hcl
-resource "aws_instance" "foo" {
-  ami = "ami-1234567"
+resource "null_resource" "foo" {
+  triggers = {
+    foo = "bar"
+  }
 }
 ```
 
@@ -118,7 +120,7 @@ The following policy would evaluate to `true`:
 ```python
 import "tfconfig"
 
-main = rule { tfconfig.module(["foo"]).resources.aws_instance.foo.config.ami is "ami-1234567" }
+main = rule { subject.module(["foo"]).resources.null_resource.foo.config.triggers[0].foo is "bar" }
 ```
 
 ### Value: `module_paths`
@@ -155,7 +157,7 @@ Here is an example of a function to retrieve all resources of a particular type
 from all modules. Note the use of `else []` in case some modules don't have any
 resources; this is necessary to avoid the function returning undefined.
 
-```
+```python
 import "tfconfig"
 
 get_vms = func() {
@@ -223,27 +225,27 @@ directly map to Terraform config keys and values.
 As a consequence of the mapping of this key to raw Terraform configuration,
 complex structures within Terraform configuration are grouped _per-instance_ as
 they are represented within the actual configuration itself. This has
-implications when defining policy correctly.
+implications when defining policy correctly. This applies to all complex
+structures - lists, sets, and maps.
 
 As an example, consider the following resource block:
 
-```
-resource "aws_instance" "foo" {
-  # ...
-
-  tags {
+```hcl
+resource "null_resource" "foo" {
+  triggers = {
     foo = "one"
   }
-
-  tags {
+  
+  triggers = {
     bar = "two"
   }
 }
 ```
 
 In this example, one would need to access both
-`tfconfig.resources.aws_instance.foo.config.tags[0].foo` and
-`tfconfig.resources.aws_instance.foo.config.tags[1].bar` to reach both tags.
+`tfconfig.resources.null_resource.foo.config.triggers[0].foo` and
+`tfconfig.resources.null_resource.foo.config.triggers[1].bar` to reach both
+triggers.
 
 This can be better represented by a loop. Given the above example, the following
 policy would evaluate to `true`:
@@ -252,8 +254,8 @@ policy would evaluate to `true`:
 import "tfconfig"
 
 main = rule {
-	all subject.resources.aws_instance.foo.config.tags as tags {
-		all tags as _, value {
+	all tfconfig.resources.null_resource.foo.config.triggers as triggers {
+		all triggers as _, value {
 			value in ["one", "two"]
 		}
 	}
@@ -289,8 +291,8 @@ represents the values of the keys within the provisioner.
 
 As an example, given the following resource block:
 
-```
-resource "aws_instance" "foo" {
+```hcl
+resource "null_resource" "foo" {
   # ...
 
   provisioner "local-exec" {
@@ -305,7 +307,7 @@ The following policy would evaluate to `true`:
 import "tfconfig"
 
 main = rule {
-	tfconfig.resources.aws_instance.foo.provisioners[0].config.command is "echo ${self.private_ip} > file.txt"
+	tfconfig.resources.null_resource.foo.provisioners[0].config.command is "echo ${self.private_ip} > file.txt"
 }
 ```
 
@@ -318,8 +320,8 @@ represents the type of the specific provisioner.
 
 As an example, in the following resource block:
 
-```
-resource "aws_instance" "foo" {
+```hcl
+resource "null_resource" "foo" {
   # ...
 
   provisioner "local-exec" {
@@ -330,10 +332,10 @@ resource "aws_instance" "foo" {
 
 The following policy would evaluate to `true`: 
 
-```
+```python
 import "tfconfig"
 
-main = rule { tfconfig.resources.aws_instance.foo.provisioners[0].type is "local-exec" }
+main = rule { tfconfig.resources.null_resource.foo.provisioners[0].type is "local-exec" }
 ```
 
 ## Namespace: Module Configuration
@@ -422,7 +424,7 @@ question loaded. The `alias` key will not be available within this namespace.
 
 As an example, given the following provider declaration block:
 
-```
+```hcl
 provider "aws" {
   alias  = "east"
   region = "us-east-1"
@@ -448,7 +450,7 @@ value](#value-version).
 
 As an example, given the following provider declaration block:
 
-```
+```hcl
 provider "aws" {
   region = "us-east-1"
 }
@@ -472,7 +474,7 @@ the pessimistic operator.
 
 As an example, given the following provider declaration block:
 
-```
+```hcl
 provider "aws" {
   version = "~> 1.34"
 }
@@ -516,7 +518,7 @@ be confused with [`undefined`][ref-sentinel-undefined]).
 
 As an example, given the following variable blocks:
 
-```
+```hcl
 variable "foo" {
   default = "bar"
 }
@@ -546,7 +548,7 @@ represents the description of the variable, as provided in configuration.
 
 As an example, given the following variable block:
 
-```
+```hcl
 variable "foo" {
   description = "foobar"
 }
--- a/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
@@ -167,9 +167,14 @@ main = rule { tfconfig.module_paths contains ["foo"] }
 
 #### Iterating through modules
 
-Here is an example of a function to retrieve all resources of a particular type
-from all modules. Note the use of `else []` in case some modules don't have any
+Iterating through all modules to find particular resources can be useful. This
+example shows how to use `module_paths` with the [`module()`
+function](#function-module-) to retrieve all resources of a particular type from
+all modules (in this case, the [`azurerm_virtual_machine`][ref-tf-azurerm-vm]
+resource). Note the use of `else []` in case some modules don't have any
 resources; this is necessary to avoid the function returning undefined.
+
+[ref-tf-azurerm-vm]: /docs/providers/azurerm/r/virtual_machine.html
 
 ```python
 import "tfconfig"
--- a/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
@@ -422,7 +422,7 @@ The **provider namespace** represents data on the declared providers within a
 namespace.
 
 This namespace is indexed by provider type and _only_ contains data about
-providers when actually declared. Hence, if you are using a completely implicit
+providers when actually declared. If you are using a completely implicit
 provider configuration, this namespace will be empty.
 
 ### Value: `alias`
--- a/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
@@ -28,9 +28,9 @@ The following is a tree view of the import namespace. For more detail on a
 particular part of the namespace, see below.
 
 -> Note that the root-level alias keys shown here (`data`, `modules`,
-`providers`, `resources`, and `variables`) are shortcuts to the [module
-namespace](#namespace-module). For more details, see the section on [root
-namespace aliases](#root-namespace-aliases).
+`providers`, `resources`, and `variables`) are shortcuts to a [module
+namespace](#namespace-module) scoped to the root module. For more details, see
+the section on [root namespace aliases](#root-namespace-aliases).
 
 ```
 tfconfig
--- a/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
@@ -193,7 +193,7 @@ get_vms = func() {
 The **module namespace** can be loaded by calling [`module()`](#function-module-)
 for a particular module.
 
-It's a namespace that can be used to load the following data:
+It can be used to load the following child namespaces:
 
 * `data` - Loads the [resource namespace](#namespace-resources-data-sources),
   filtered against data sources.
--- a/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
@@ -148,7 +148,16 @@ module "foo" {
 }
 ```
 
-The following policy would evaluate to `true`:
+The value of `module_paths` would be:
+
+```
+[
+	[],
+	["foo"],
+]
+```
+
+And the following policy would evaluate to `true`:
 
 ```python
 import "tfconfig"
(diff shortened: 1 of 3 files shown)


content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
 
 The tfstate import provides access to the Terraform state.
 
-### tfstate.module_paths
+This is the Terraform state as it exists at plan time. Depending on the state of
+a Terraform workspace, the data in this import may be incomplete (such as when a
```
armchairlinguist: I think I might suggest leaving off the initial clause here, to avoid questions about what type of workspace is involved here (TFE or Terraform). I don't think it's important as a qualifier?

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
+resource has not been created yet).
+
+Depending on what you are looking for, the [`tfconfig`][import-tfconfig] and
+[`tfplan`][import-tfplan] imports can assist with providing a more complete view
+of the state and configuration of the workspace as it exists currently, or in
+the future as well. `tfplan` also has aliases to both the `tfconfig` and
```
armchairlinguist: "as it exists currently or will exist after the apply"?

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
+`tfplan` imports to help simplify policy.
+
+[import-tfconfig]: /docs/enterprise/sentinel/import/tfconfig.html
+[import-tfplan]: /docs/enterprise/sentinel/import/tfplan.html
+
+-> To reiterate, this import works with the Terraform state that _currently
```
armchairlinguist: Per Nick's recommendations, this should have a **bold** styled header saying what it is (`Note:` I think, in this case)

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
+exists_ at plan time, not what it will look like _after_ the plan is applied.
+This means that resources that have not been created yet (and hence, can have no
+state) will not be present in the namespaces in this import. If you are looking
+for something state-like but also contains future data, see the
```
armchairlinguist: state-like that also

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
+│               └── tainted (boolean)
+│
+├── module_paths ([][]string)
+├── terraform_version (string)
+│
+├── data (root module alias)
+├── outputs (root module alias)
+├── path (root module alias)
+└── resources (root module alias)
+```
```
armchairlinguist: same comments here as for the other two

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
+  triggers = {
+    foo = "bar"
+  }
+}
+```
 
-The module function finds the module at the given path.
 
-## Type: m
+The following policy would evaluate to `true`, if the resource was present in
```
armchairlinguist: no comma needed here

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
-type, the value is the same as `m.data.TYPE`.
+```hcl
+module "foo" {
+  # ...
+}
+```
 
-### m.data.TYPE
+The following policy would evaluate to `true`, _only_ if the diff had changes
+for that module:
```
armchairlinguist: Hm, is this maybe copypasta? It seems like it should say "The following policy would evaluate to true only if the module is present in the state:"

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
-### o.sensitive
+### Value: `terraform_version`
 
-True if the output is marked as sensitive
+* **Value Type:** String.
 
-### o.type
+The `terraform_version` value within the [root namespace](#namespace-root)
+represents the version of Terraform in use when the state was saved. This can be
+used to enforce a specific version of Terraform in a policy check.
```
armchairlinguist: I think this might also be copy-pasta. If the version is in the state, it cannot be used as an enforcement check since that already happened - it might make more sense to say it can be used to check if there is a large jump in the old vs new versions?

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
 
-The data type of this output.
+As an example, the following policy would evaluate to `true`, as long as the
```
armchairlinguist: comma not needed after true

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
+corresponding namespaces within the module namespace, loaded for the root
+module. They are essentially the equivalent of running `module([]).KEY`.
 
+### Value: `path`
+
+* **Value Type:** List of strings.
+
+The `path` value within the [module namespace](#namespace-module) contains the
+path of the module that the namespace represents. This is represented as a list
+of stings.
```
armchairlinguist: More stings! 🐝

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
+As an example, if the following module block was present within a Terraform
+configuration:
+
+```hcl
+module "foo" {
+  # ...
+}
+```
+
+The following policy would evaluate to `true`, _only_ if the module was present
```
armchairlinguist: comma not needed here

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
+  `tfstate.resources.aws_instance`. This would be indexed off of the names of
+  each resource (`foo`, `bar`, and so on), with each of those maps containing
+  instances indexed on count index as per above.
+* To fetch all resources indiscriminately within the root module,
+  `tfstate.resources` will do. This is then indexed by resource type
+  (`aws_instance`, for example), with the other maps cascading down from there.
+
+Further explanation of the namespace will be in the context of resources. As
+mentioned, when operating on data sources, use the same syntax, except with
+`data` in place of `resources`.
```
armchairlinguist: Same comments here as in the other location this is used

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
+
+```hcl
+resource "null_resource" "foo" {
+  triggers = {
+    foo = "bar"
+  }
+}
+```
+
+The following policy would evaluate to `true`, if the resource was in the state:
```
armchairlinguist: comma not needed here

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
+
+### Value: `depends_on`
+
+* **Value Type:** A list of strings.
+
+The `depends_on` value within the [resource
+namespace](#namespace-resources-data-sources) contains the dependencies for the
+resource.
+
+This is a list of strings which will contain the resource address relative to
```
armchairlinguist: list of strings that contains

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
+resource "null_resource" "bar" {
+  # ...
+
+  depends_on = [
+    "null_resource.foo",
+  ]
+}
+```
+
+The following policy would evaluate to `true`, if the resource was in the state:
```
armchairlinguist: comma not needed

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
+### Value: `id`
+
+* **Value Type:** String.
+
+The `id` value within the [resource
+namespace](#namespace-resources-data-sources) contains id of the resource.
+
+As an example, given the following data source:
+
+-> Note that we use a _data source_ here because the
```
armchairlinguist: I'd suggest moving this note to somewhere that isn't in the middle of the example flow. Maybe just interchange the line above and this note. And (similar to other comment earlier) use the convention `-> **Note**: We use`

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
+
+main = rule { tfstate.resources.null_resource.foo[0].tainted }
+```
+
+## Namespace: Outputs
+
+The **output namespace** represents all of the outputs present within a
+[module](#namespace-module). This would be any outputs that would have been
+saved to state either as the result of a previous apply, or from the pre-plan
+refresh if their values were known at plan-time.
```
armchairlinguist: I would change the second sentence here to: "Outputs are present in a state if they were saved during a previous apply, or if they were updated with known values during the pre-plan refresh."

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
+
+Note that this can be used to fetch both the outputs of the root module, and the
+outputs of any module in the state below the root, allowing one to see outputs
+that have not been threaded to the root module
```
armchairlinguist: This is odd phrasing. I'd put a period after "below the root" and make the last sentence: "This makes it possible to see outputs that have not been threaded through to the root module."

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
+
+This namespace is indexed by output name.
+
+### Value: `sensitive`
+
+* **Value Type:** Boolean.
+
+The `sensitive` value within the [output namespace](#namespace-outputs) is
+`true` when the output has been marked as sensitive.
```
armchairlinguist: How does one do this? Should we link to documentation here?

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
+main = rule { tfstate.outputs.foo.sensitive }
+```
+
+### Value: `type`
+
+* **Value Type:** String.
+
+The `type` value within the [output namespace](#namespace-outputs) gives the
+output's type. This will be one of either `string`, `list`, or `map`, which are
+currently the only types available for outputs in Terraform.
```
armchairlinguist: I'd scrub out either (since it's usually used if there are two, not three, possibilities), and end the sentence after map, then say "These are currently the only..."

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -10,103 +10,526 @@ description: |-
...
+    "bar",
+  ]
+}
+
+output "map" {
+  value = {
+    foo = "bar"
+  }
+}
+```
```
armchairlinguist: I wonder if there's any value in making these more realistic? I notice we have kind of a mix of realistic and toy examples, but it's maybe a little confusing for the string output to also be called string.

content/source/docs/enterprise/sentinel/import/tfconfig.html.md
```diff
@@ -8,124 +8,556 @@ description: |-
...
+│               ├── default (value)
+│               └── description (string)
+├── module_paths ([][]string)
+│
+├── data (root module alias)
+├── modules (root module alias)
+├── providers (root module alias)
+├── resources (root module alias)
+└── variables (root module alias)
+```
```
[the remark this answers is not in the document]
vancluever: Just putting a mention in here that I've discussed this with @armchairlinguist offline. I think having this in addition to the more human-readable tree view here is import to have. However, we actually have an internal task for mocks into the documentation after this is complete, which will address these concerns.

The things that I'll do to address though:

* The mock will be linked to from the namespace overview here (in addition to the sidebar).
* I will more clearly address the aliased keys here as per @armchairlinguist's suggestions.

content/source/docs/enterprise/sentinel/import/tfplan.html.md
```diff
@@ -8,258 +8,505 @@ description: |-
...
 
-The Terraform plan import also allows you to access the configuration
-files as well as the Terraform state at the time the plan was run.
-You can also access an "applied" state that merges the plan with the
-state to create the planned state after apply. Note that any computed
-values will not be visible in this state.
+The Terraform plan import also allows you to access the configuration files (an
+alias to the [`tfconfig`][import-tfconfig] import) as well as the Terraform
+state (an alias to the [`tfstate`][import-tfstate] import) at the time the plan
+was run.  You can also access an "applied" state that merges the plan with the
```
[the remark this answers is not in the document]
vancluever: Just adding a note here on that we also want to be more specific about how while (in current TFE) this data is all sourced from the plan's saved config and data, at a later time (pending possible implementations of policy checks at different parts of the run lifecycle) the root-level `tfconfig` and `tfstate` imports may source their respective data from _actual_ config (ie: a `*.tf` config bundle) and state (analogous to a `.tfstate` file).

content/source/docs/enterprise/sentinel/import/tfconfig.html.md
```diff
@@ -8,124 +8,556 @@ description: |-
...
+The **resource namespace** is a namespace shared between resources and data
+sources, and is identical in structure and behavior regardless of which
+classification you are loading.
 
-### m.resources.TYPE
+Accessing an individual resource or data source within this namespace can be
+accomplished by specifying the type and name, in the syntax
+`[resources|data].TYPE.NAME`, depending on whether or not you are loading the
+namespace for a resource (accessed in the `resources` name space) or data
+sources (accessed in the `data`) namespace.
```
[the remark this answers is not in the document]
vancluever: I'm re-writing the first paragraph to stress that this is a _type_, in an attempt to try to not imply that this namespace is shared.

content/source/docs/enterprise/sentinel/import/tfconfig.html.md
```diff
@@ -8,124 +8,556 @@ description: |-
...
+* `variables` - Loads the [variable namespace](#namespace-variables).
 
-### m.providers.NAME
+### Root Namespace Aliases
 
-This returns the configuration for the provider with the given name.
+The root-level `data`, `modules`, `providers`, `resources`, and `variables` keys
+all alias to their corresponding namespaces within the module namespace, loaded
+for the root module. They are essentially the equivalent of running
+`module([]).KEY`.
```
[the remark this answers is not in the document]
vancluever: Maybe ;) But I've added the qualifier in the root namespace tree too that links to this part as well!

content/source/docs/enterprise/sentinel/import/tfplan.html.md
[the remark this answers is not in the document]
vancluever: I've opted to put most of the information for this in its own section, in addition to re-wording the initial synopsis to direct people to that section for more information on accessing that data in the appropriate section.

content/source/docs/enterprise/sentinel/import/tfplan.html.md
```diff
@@ -8,258 +8,505 @@ description: |-
...
-    vms = []
-    for tfplan.module_paths as path {
-        vms += values(tfplan.module(path).resources.azurerm_virtual_machine) else []
-    }
-    return vms
+	vms = []
+	for tfplan.module_paths as path {
+		vms += values(tfplan.module(path).resources.azurerm_virtual_machine) else []
+	}
+	return vms
```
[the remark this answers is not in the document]
vancluever: Yeah, this is a byproduct of [`sentinel fmt`](https://docs.hashicorp.com/sentinel/commands/fmt), which I've been using to auto-format the Sentinel snippets. The Terraform samples are sent through `terraform fmt`, but it looks like the HCL formatter uses spaces.

I think this is safe to leave for now. Created https://github.com/hashicorp/sentinel/issues/109 to mull whether or not we want to switch the formatter to spaces though.

content/source/docs/enterprise/sentinel/import/tfplan.html.md
```diff
@@ -8,258 +8,505 @@ description: |-
...
+```
+
+### Value: `variables`
+
+* **Value Type:** A string-keyed map of values.
+
+The `variables` value within the [root namespace](#namespace-root) represents
+all of the variables that were set when creating the plan.
+
+As only root module variables when creating a plan, `variables` will only
```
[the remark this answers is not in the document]
vancluever: Haha wow. Yeah, I'll re-word that. I'm also wondering if the assumption that the plan only contains root module variables is actually correct - looking into that.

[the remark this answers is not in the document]
vancluever: Confirmed that only root-level variables are part of a plan. The majority of the plan-relevant data in a plan is the [diff](https://github.com/hashicorp/terraform/blob/c64d8bdf354eb27125416de89ee199a740ff8d35/terraform/diff.go#L37-L42) which ultimately links down to [ModuleDiff](https://github.com/hashicorp/terraform/blob/c64d8bdf354eb27125416de89ee199a740ff8d35/terraform/diff.go#L206-L212). This does not contain any variables, as these will have been interpolated by the time they made it down there.

content/source/docs/enterprise/sentinel/import/tfplan.html.md
```diff
@@ -8,258 +8,505 @@ description: |-
...
 
-`data` returns a map of all the data sources in the root module.
-This is identical to `tfplan.module([]).data`. If you want to
-validate a policy against all data sources of a particular type,
-you will need to use logic similar to that in the second example in
-the [tfplan.module_paths](#tfplan-module_paths) section above.
+* `data` - Loads the [resource namespace](#namespace-resources-data-sources),
+  filtered against data sources.
+* `resources` - Loads the [resource
+  namespace](#namespace-resources-data-sources), filtered against resources.
```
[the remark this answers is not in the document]
vancluever: This has been clarified in all imports :)

content/source/docs/enterprise/sentinel/import/tfplan.html.md
```diff
@@ -8,258 +8,505 @@ description: |-
...
 
-### tfplan.module(path)
+### Root Namespace Aliases
 
-The module function finds the module at the given path and returns
-the type "module". See the documentation for the type "module" for
-available attributes.
+The root-level `data`, and `resources` keys both alias to their corresponding
```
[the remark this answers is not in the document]
vancluever: This should be fixed by adding some new context. New text reads: 

> In addition to this, the root-level `data`, `path`, and `resources` keys alias to their corresponding namespaces or values within the module namespace.

content/source/docs/enterprise/sentinel/import/tfplan.html.md
```diff
@@ -8,258 +8,505 @@ description: |-
...
+
+* To fetch all `aws_instance.foo` resource instances within the root module, you
+  can specify `tfplan.resources.aws_instance.foo`. This would then be indexed
+  off of the resource count index (`0`, `1`, `2`, and so on). Note that as
+  mentioned above, these elements must be accessed using square-bracket map
+  notation (so `["0"]`, `["1"]`, `["2"]`, and so on) instead of dotted notation.
+* To fetch all `aws_instance` resources within the root module, you can specify
+  `tfplan.resources.aws_instance`. This would be indexed off of the names of
+  each resource (`foo`, `bar`, and so on), with each of those maps containing
+  instances indexed on count index as per above.
```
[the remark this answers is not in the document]
vancluever: Changed this to "indexed by resource count index" which is the wording we have in the bullet before this. It is kind of awkward to express, but hopefully this helps a bit.

content/source/docs/enterprise/sentinel/import/tfplan.html.md
```diff
@@ -8,258 +8,505 @@ description: |-
...
+* To fetch all resources indiscriminately within the root module,
+  `tfplan.resources` will do. This is then indexed by resource type
+  (`aws_instance`, for example), with the other maps cascading down from there.
```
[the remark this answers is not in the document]
vancluever: This is correct! Corrected it.

content/source/docs/enterprise/sentinel/import/tfplan.html.md
```diff
@@ -8,258 +8,505 @@ description: |-
...
+the map is only 1 level deep with each key possibly representing a diff for a
+particular complex value within the resource.
 
-This is a boolean value that is true if this attribute's new value is
-a computed value. A computed value is a value that is not known until
-apply time. This can be because this is a dynamic value from this
-resource itself such as the IP address of a new compute instance, or
-because it is a parameter that is being populated from a dynamic
-value from another resource.
+See the diff namespace for specific examples.
```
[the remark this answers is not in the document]
vancluever: This is more referring to the **diff namespace** as a concept, versus the key, so backticks aren't really appropriate here.

To eliminate confusion though, I'll re-word it.

content/source/docs/enterprise/sentinel/import/tfplan.html.md
```diff
@@ -8,258 +8,505 @@ description: |-
...
 
-When a value is removed, its `new` value will be an empty string.
+* **Value Type:** Boolean.
 
-## Type: module
+The `computed` value within the [diff namespace](#namespace-resource-diff) is
+`true` if a value is currently unknown in the diff, but is changing. This
+happens when a value depends on a value belonging to a resource that either does
+not exist yet or is changing state in a way that the new value will not be known
+until the apply for that resource completes.
```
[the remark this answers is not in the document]
vancluever: I'd like to keep the syntax/grammar of the first sentence, so I've blended this in to it. New paragraph:

> The `computed` value within the diff namespace is `true` if the resource key in question depends on another value that isn't yet known. Typically, that means the value it depends on belongs to a resource that either doesn't exist yet, or is changing state in such a way as to affect the dependent value so that it can't be known until the apply is complete.

content/source/docs/enterprise/sentinel/import/tfplan.html.md
```diff
@@ -8,258 +8,505 @@ description: |-
...
+    foo_id = "${null_resource.foo.id}"
+  }
+}
+```
 
-An example of accessing a resource named "app" of type "aws_instance"
-in a module named "web".
+The following policy would evaluate to `true`, if the resource was in the diff,
+the value of the tag was changing, and the `id` of `null_resource.foo` was
+currently not known (example: the resource has not been created yet):
```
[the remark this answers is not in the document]
vancluever: This example needs modification anyway as the wording is no longer relevant to the resource example. Thanks for the catch!

content/source/docs/enterprise/sentinel/import/tfplan.html.md
```diff
@@ -8,258 +8,505 @@ description: |-
...
 
-Note for a data source to be present here, it must be in the plan, which
-generally means it must have some unknown data at plan time.
+The `new` value within the [diff namespace](#namespace-resource-diff) contains
+the new value of a changing attribute, _if_ the value is known at plan time.
 
-## Type: resource
+If the value is currently unknown, this field will be blank. Use the
+[`computed`](#value-computed) value to determine if the value contained here is
+actually a known zero value or not.
```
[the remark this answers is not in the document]
vancluever: New text: 

> `new` will be an empty string if the attribute's value is currently unknown. For more details on detecting unknown values, see `computed`.

content/source/docs/enterprise/sentinel/import/tfstate.html.md
vancluever: Agree, especially as a future-proofing measure, ie: when this import processes things outside of the plan.

content/source/docs/enterprise/sentinel/import/tfstate.html.md
vancluever: Just talked about this offline - there's a number of info boxes where this would need to be updated if this is a convention we want to stick to. Leaving this here for now, pending @nfagerlund's comments.

vancluever: Ironically, a re-write put context into the info box that more warranted a **NOTE** heading. :stuck_out_tongue:

content/source/docs/enterprise/sentinel/import/tfstate.html.md
vancluever: This isn't a copy/paste, but indeed it's confusing. This is exposed in both the plan and state (and indeed saved in both as well). Which one do you check? Definitely the plan would be the better one, to be honest as it would technically be more current - so maybe an info box here would be a good idea noting this, keeping in mind that we might have future scenarios where a plan won't be available (example: periodic policy checks).

To be honest, I don't even know if the version is a good idea to expose in imports - I'd personally think this might be better in a `tfe` import or something along those lines that can get run data from TFE itself, which would be the ultimate source of truth.

content/source/docs/enterprise/sentinel/import/tfconfig.html.md
```diff
@@ -8,124 +8,556 @@ description: |-
...
+namespace for a resource (accessed in the `resources` name space) or data
+sources (accessed in the `data`) namespace.
 
-This returns all the resources in the module with the given type
-as a map from name to resource. For the documentation on the
-value, see `m.resources.TYPE.NAME`
+In addition, each of these namespace levels is a map, so to access all items
+_matching_ a certain type, you can use `[resources|data].TYPE`, and to access
+all data sources _grouped by_ type, you can use a simple bareword `resources` or
+`data` map.
```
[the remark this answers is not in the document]
vancluever: Yeah, I've made this one look similar to the other two imports now, sans count key.

vancluever:
@armchairlinguist all of the items should be addressed now - if you want to take another look that would be huge! Thanks again for your help on this one - super useful feedback!

content/source/docs/enterprise/sentinel/import/tfconfig.html.md
```diff
@@ -8,124 +8,577 @@ description: |-
...
+main = rule { tfconfig.modules.foo.source is "./foo" }
+```
+
+### Value: `config`
+
+* **Value Type:** A string-keyed map of values.
+
+The `config` value within the [module configuration
+namespace](#namespace-module-configuration) represents the values of the keys
+within the module configuration.
```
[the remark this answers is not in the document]
vancluever: They aren't - I've added some clarification on this one on both `source` for modules and `alias` in providers just to ensure that there's 100% understanding as to what is not included.

content/source/docs/enterprise/sentinel/import/tfconfig.html.md
```diff
@@ -8,124 +8,577 @@ description: |-
...
+main = rule { tfconfig.providers.aws.config.region is "us-east-1" }
+```
+
+### Value: `version`
+
+* **Value Type:** String.
+
+The `version` value within the [provider namespace](#namespace-providers)
+represents the explicit expected version of the supplied provider. This includes
+the pessimistic operator.
```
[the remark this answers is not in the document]
vancluever: Yeah, there's no easy answer to this right now. Meaningful comparison would be something along the lines of taking the pessimistic constraint defined, determining its upper and lower bounds, and then comparing that against another constraint to make sure that it doesn't violate it. I've outlined this in https://github.com/hashicorp/sentinel-plugin-terraform/issues/12 as something we could possibly do in the future.

content/source/docs/enterprise/sentinel/import/tfplan.html.md
```diff
@@ -8,258 +8,538 @@ description: |-
...
+
+### Value: `module_paths`
+
+* **Value Type:** List of a list of strings.
+
+The `module_paths` value within the [root namespace](#namespace-root) is a list
+of all of the modules within the Terraform _diff_ for the current plan.
+
+Note the distinction here - this means if there are no changes for any resources
+within the module of concern, the module will not show up in `module_paths`.
```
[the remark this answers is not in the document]
vancluever: It does. I'm re-doing this section to be a little less animated, so that it matches tone with what I will be adding for all 3 of the module instances in each import (ie: an invalid module or module unknown to config/plan/state will return null).

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -8,105 +8,551 @@ description: |-
...
+
+Further explanation of the namespace will be in the context of resources. As
+mentioned, when operating on data sources, use the same syntax, except with
+`data` in place of `resources`.
+
+### Value: `attr`
+
+* **Value Type:** A string-keyed map of values.
+
+The `applied` value within the [resource
```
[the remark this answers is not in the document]
vancluever: Great catch! Thanks!

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -8,105 +8,551 @@ description: |-
...
+* **Value Type:** List of a list of strings.
 
-This returns all the data sources in the module as a map. The map key is the
-type, the value is the same as `m.data.TYPE`.
+The `module_paths` value within the [root namespace](#namespace-root) is a list
+of all of the modules within the Terraform state at plan-time.
 
-### m.data.TYPE
+Note the distinction here - this means if a module is not present in state, it
+will not be present in `module_paths`.
```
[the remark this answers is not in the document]
vancluever: (Been modified with the other work)

Also, yeah, state only gives the state as it exists the time that Sentinel is run. There's no predictive features in any of the imports right now that will tell you what _modules_ will be added. Best way to compare that would probably be compare `tfstate.module_paths` with `tfplan.module_paths`.

vancluever:
Alright, I've addressed all of @nfagerlund's feedback. Given this, and @armchairlinguist's approval, I'm going to go ahead and merge. Any other issues we can resolve later.

Thank you very much to both of you for your thorough reviews and feedback!

nfagerlund:
👍🏼

PR #456, opened 2018-09-12
docs/sentinel: Add mocks

vancluever:
This update adds mocks for all 3 imports. They are linked in the sidebar, and also in each import via an info box in the namespace outline.

This PR is dependent on #450. The PR should probably be re-based off of master post-merge to make the commit load much less overwhelming.

### DIFF
--- a/content/source/docs/enterprise/sentinel/import/mock-tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/mock-tfconfig.html.md
@@ -1,0 +1,128 @@
+---
+layout: enterprise2
+page_title: "Sentinel - Terraform Enterprise"
+sidebar_current: "docs-enterprise2-sentinel-imports-tfconfig-mock"
+description: |-
+    This page provides a mock for the tfconfig Sentinel import.
+---
+
+# Mocking tfconfig Data
+
+Below is a sample of what a mock for the [`tfconfig`][ref-tfconfig] import
+would look like. You can use this data with the mocking or testing features of
+the [Sentinel Simulator][ref-sentinel-simulator] to create test data for rules.
+
+[ref-tfconfig]: /docs/enterprise/sentinel/import/tfconfig.html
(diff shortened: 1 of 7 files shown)


content/source/docs/enterprise/sentinel/import/mock-tfconfig.html.md.erb
```diff
@@ -0,0 +1,27 @@
...
+
+For more details on how to work with mock data in Sentinel, see the section on
+[Mocking Imports][ref-mocking-imports] in the [official Sentinel
+documentation][ref-official-sentinel-documentation].
+
+[ref-mocking-imports]: https://docs.hashicorp.com/sentinel/writing/imports#mocking-imports
+[ref-official-sentinel-documentation]: https://docs.hashicorp.com/sentinel/
+
+```json
+<%= IO.read("#{__dir__}/mock-tfconfig.json").chomp %>
```
nfagerlund: Huh, intriguing!

- I'm guessing you've built the preview site with this to make sure, right? 
- Is there prior art for doing this in our product websites? Using `IO.read` seems a little much...; tbh there should be a dedicated helper tag for this sort of thing that includes some error handling and safety guardrails, but... the future.
- Why not paste it inline? The main value for doing this is if you're pulling the same snippet into multiple pages, or you want to link the raw JSON file from somewhere, but this looks like the only place it's used.

vancluever: I think I'll fix this to do it inline for now. However, for further justification, I wanted to make this easy to work with _outside_ of the documentation template. These are large JSON documents, and pasting them inline usually means you lose the ability to work with them in an editor.

FYI - we may actually end up doing more of this work later however, as we may want to do some things like automate the synchronization of examples from the Sentinel plugin to the documentation. For this kind of stuff, ERB will be valuable as we can just use it to source the examples from the places we need to.

I did also build the preview. :)

vancluever:
Thanks @nfagerlund! Will merge now

PR #457, opened 2018-09-13
Add “Creating and Managing Terraform Policies”

thrashr888:
This is a new final page in our Enterprise “Getting Started” guide. It adds a walkthrough for adding Sentiel policies to an organization. In this case, we’re just adding a simple passthrough policy that should always pass and not block a run. It references the main Sentinel docs, as well as the other Enterprise Sentinel docs. The content closely follows other pages in the guide, namely the workspaces page.

Here's a link to the rendered page: https://github.com/hashicorp/terraform-website/blob/8d221edc9ac9327a952196617c3804248e4acb6b/content/source/docs/enterprise/getting-started/policies.html.md

screenshot_2018-09-12_hashicorp_terraform-website


content/source/docs/enterprise/getting-started/policies.html.md
```diff
@@ -0,0 +1,76 @@
...
+## About Sentinel Policies
+
+_Policies_ in TFE are composed of [Sentinel policies](../sentinel/index.html) with some extra features. [Sentinel](https://www.hashicorp.com/sentinel) is an embedded policy-as-code framework integrated with the HashiCorp Enterprise products. It enables fine-grained, logic-based policy decisions, and can be extended to use information from external sources. Within TFE, you can use Sentinel to apply checks to your runs.
+
+A policy consists of:
+
+- The Sentinel policy code
+- An enforcement mode that changes how a policy affects the run lifecycle
+
+Currently, policies in an organization apply to **all** workspaces. You'll want to consider that when writing your policies so they won't fail unrelated workspace runs or configurations. In the future, TFE will provide a way to manage policies so you can apply them per-workspace.
```
vancluever: > You'll want to consider that when writing your policies so they won't fail unrelated workspace runs or configurations.

I think this could be re-worded so that "fail" isn't a verb? Maybe:

> You'll want to consider that when writing your policies so they don't cause unintended failures.

content/source/docs/enterprise/getting-started/policies.html.md
```diff
@@ -0,0 +1,76 @@
...
+
+![navigate](./images/policy-navigate.png)
+
+The policy list shows all of the policies you have access to; if you haven't created any, it's empty.
+
+To create your first policy, click the "Create new policy" button in the upper right.
+
+![new button](./images/policy-new-button.png)
+
+On the "Create a new Policy" page, you need to enter at least two items: a policy name, and the policy code. When you've finished, click the "Create Policy" button.
```
vancluever: > the "Create a new Policy" page

Looks like the page is actually titled "Create new Sentinel Policy", as per the screenshot?

content/source/docs/enterprise/getting-started/policies.html.md
```diff
@@ -0,0 +1,76 @@
...
+
+![fields on new policy page](./images/policy-new-fields.png)
+
+### Policy Name
+
+A policy name should tell your colleagues what the policy is for. Most policies are _a particular environment_ of _a particular Terraform configuration,_ so the name should include both the name of the configuration and the name of the environment.
```
vancluever: > Most policies are a particular environment of a particular Terraform configuration

* Should probably read "Most policies are _for_ a particular environment of a particular Terraform configuration".

On a more fundamental level, I would be careful in assuming that even _most_ policies are going to be specific for environments or workspaces - while there will be a number that do for sure, I would expect a well-laid out governance structure to treat these as exceptions rather than the rule, as governance and policy should apply to the organization globally (based on the principle that a large amount of exceptions dilutes and hinders good governance). I think we should be encouraging policy authors to structure their policies and governance so that they can catch a good number of their cases without making exceptions.

Some examples of good organization-wide governance material:

* The classic example of ensuring that all AWS assets have tags assigned to them (as we have on the Sentinel website). This could even be further drilled into to ensure that the tag matches a certain pattern (I used to encourage a pattern of `ORG/PROJECT`).
* Ensuring all deployed network assets contain a general set of network ACLs.
* Ensuring all of your `aws_instance` data sources [are looking up AMIs on a specific owner](https://github.com/terraform-providers/terraform-provider-aws/pull/5576).

I think we should probably not necessarily imply a best practice here, and save this material for later. Maybe we could just edit this to:

> Policy Name
>
> A policy name should tell your colleagues what the policy is for. Examples could be "Require tags on all instances" or "Enforce network ACLs". If a policy is for a particular workspace, or environment within a workspace, include the name of the workspace and/or environment.

content/source/docs/enterprise/getting-started/policies.html.md
```diff
@@ -0,0 +1,76 @@
...
+
+For this example, we'll just create a sample passthrough policy that will allow all runs to "PASS" our policy check. 
+
+In this example, we're using a configuration named "minimum" and we're deploying it in a production environment, so we named it `passthrough`.
+
+### Enforcement Mode
+
+Enforcement Mode alters how a policy result affects your run; "hard-mandatory" will always stop a run if a policy fails, "soft-mandatory" will pause a run and allow a failure to be overridden, and "advisory" will log failures but always allow a run to continue. Use "soft-mandatory" for now so we can see
```
vancluever: Looks like the last sentence ended early.

content/source/docs/enterprise/getting-started/policies.html.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Policy Code
+
+You can paste the following code into the code input box. It always resolves to `true` and will allow all runs to pass. This is a good way to see policy checks applied to your run without having a policy "FAIL".
+
+```python
+main = rule { true }
+```
+
+Later, you can switch `true` to `false` to see how a "FAIL" during a policy check can affect your run. You can find many more examples in our [Example policies doc](../sentinel/examples.html).
```
vancluever: > Example policies doc

Can we change this to "Example Policies section"?

Also, can this be changed to an absolute link within the site? (`/docs/enterprise/sentinel/examples.html`)

content/source/docs/enterprise/getting-started/policies.html.md
```diff
@@ -0,0 +1,76 @@
...
+
+## What Happens in a New Policy
+
+When you create a new policy, it will be applied to all future runs. Runs that are currently queued or in progress will not be affected.
+
+## See a Policy Check in a Run
+
+Once your policy is created, you can view its effect on the run page. Start a new run to watch it play out. Again, you can start a run with the "Queue Plan" at the upper right of the workspace page, or using the "Save & Plan" button when editing the workspace's variables.
```
vancluever: > with the "Queue Plan" at the upper right of the workspace page

Should probably be changed to "with the "Queue Plan" _button_ at the upper right of the workspace page"

content/source/docs/enterprise/getting-started/policies.html.md
thrashr888: Oh dang. That was actually left over from copy/pasting the Workspace text. Will fix. Thanks!

content/source/docs/enterprise/getting-started/policies.html.md
thrashr888: I thought I saw other pages using similar relative links. Will recheck and fix. Thanks.

thrashr888: We seem to only use absolute links like that in the nav, but relative in the actual document pages. Any particular reason to make it absolute? I used relative since I was working in the context of this area of the site and it made sense to me at the time.

vancluever: My (personal) use cases for doing "absolute" relative links:

* It ensures moving the doc in question around does not break the link
* Easy copy/paste when making documentation (visit page you want to link, grab link, paste, edit out protocol/host so just URI remains)
* It has prior art in the style guide

I think we need to come to a generally acceptable consensus on this, but the results would probably be codified in the style guide right now which mainly documents relative links using full URIs.

vancluever:
LGTM as well! Great work @thrashr888!

PR #458, opened 2018-09-13
docs/sentinel: Some post-update things

vancluever:
A few things:

* It's worth noting that we can't mock functions right now, so `module()` cannot be mocked. This puts a notice into each mock doc mentioning that specifically.
* Clarified that the `module()` null return case is when a module is not present in the _diff_, versus the _plan_.
* Removed a backup file that was used to resolve merge conflicts that crept into the source tree by mistake.


vancluever:
Added a new commit that updates the resource and data source namespace addressing in the `tfstate` diagram.

PR #461, opened 2018-09-13
adding netlify tf provider

cgriggs01:
Adding submodule and links to the newly released netlify provider


alvin-huang:
The `netlify` provider is on a branch ref and not a `master` ref.

PR #462, opened 2018-09-14
Improve documentation of teams

nfagerlund:
Companion PR to the users and orgs one I submitted a while ago.


content/source/docs/enterprise/users-teams-organizations/teams.html.md
```diff
@@ -4,14 +4,83 @@ page_title: "Teams - Terraform Enterprise"
...
+
 # Teams
 
-Teams are groups of users that reflect your company's group
-structure. Organization owners can create teams and manage user membership.
-Organization owners can give teams read, write, and admin permissions to
-workspaces.
+Teams are groups of Terraform Enterprise (TFE) [users][] within an [organization][organizations]. To delegate provisioning work, the organization's owners can grant workspace permissions to specific teams.
+
+Teams can only have permissions on workspaces within their organization, although any user in the team can belong to teams in other organizations.
```
armchairlinguist: I think I might say "any user in a team", since there's no specific team in question here.

content/source/docs/enterprise/users-teams-organizations/teams.html.md
```diff
@@ -4,14 +4,83 @@ page_title: "Teams - Terraform Enterprise"
...
+
+If a user belongs to at least one team in an organization, they are considered a member of that organization.
+
+## The Owners Team
+
+Every organization has a team named `owners`, whose members have special permissions. In TFE's documentation and UI, members of the owners team are sometimes called organization owners.
+
+An organization's creator is the first member of its owners team; other members can be added or removed in the same way as other teams. Unlike other teams, the owners team can't be deleted and can't be empty; if there is only one member, you must add another before removing them.
```
armchairlinguist: It might be slightly clearer to say "you must add another member before removing the current member", since "them" is a bit ambiguous here.

content/source/docs/enterprise/users-teams-organizations/teams.html.md
```diff
@@ -4,14 +4,83 @@ page_title: "Teams - Terraform Enterprise"
...
+
+Members of the owners team have [admin permissions](./permissions.html#admin) on every workspace in the organization. Additionally, the following tasks can only be performed by organization owners:
```
armchairlinguist: Although it might be factually correct to say that they have admin permissions, it's not technically true, and future permissions could diverge. I think I'd say something like "have full access to every workspace".

content/source/docs/enterprise/users-teams-organizations/teams.html.md
```diff
@@ -4,14 +4,83 @@ page_title: "Teams - Terraform Enterprise"
...
+
+- Creating new workspaces
+- Managing Sentinel policies
+- Overriding `soft mandatory` Sentinel policies
+- Creating and deleting teams
+- Managing team membership
+- Viewing the full list of teams, including teams the user does not belong to
```
armchairlinguist: I might leave out the second bit here. It's certainly accurate that owners see teams they don't belong to, but it doesn't seem necessary to say and is a bit mind-bending.

content/source/docs/enterprise/users-teams-organizations/teams.html.md
```diff
@@ -4,14 +4,83 @@ page_title: "Teams - Terraform Enterprise"
...
+- Managing [organization settings][]
+
+## Managing Teams
+
+Teams are managed in the [organization settings][]. Click the "Settings" link in the top navigation bar, then click the "Teams" link in the left sidebar.
+
+![Screenshot: the teams page, displaying a list of teams. Each team's entry shows how many members it has.](./images/teams-list.png)
+
+The teams page includes a list of the organization's teams. Clicking a team's name in the list loads its team settings page, where you can manage its membership and settings:
```
armchairlinguist: The shift from passive/impersonal to active/personal here caught me by surprise. Do with that information what you will. :)

content/source/docs/enterprise/users-teams-organizations/teams.html.md
```diff
@@ -4,14 +4,83 @@ page_title: "Teams - Terraform Enterprise"
...
+
+![Screenshot: a team's settings page](./images/teams-team-settings.png)
+
+The team settings page includes a list of the team's current members, with badges to indicate which users have [two-factor authentication](./2fa.html) enabled.
+
+Only organization owners can manage teams; other users can view the teams page in read-only mode. When a non-owner views the teams page, it only includes the current user's teams.
```
armchairlinguist: "it only includes that user's teams"? or "only includes teams the user is a member of"?

content/source/docs/enterprise/users-teams-organizations/teams.html.md
```diff
@@ -4,14 +4,83 @@ page_title: "Teams - Terraform Enterprise"
...
+
+### Creating and Deleting Teams
+
+-> **API:** See the [Teams API](../api/teams.html).
```
armchairlinguist: Do you feel like this needs a callout? I might just have a simple statement at the bottom like "Teams can also be managed by [API]." I think I'd take this approach for all the callouts in the text that point to the API.

content/source/docs/enterprise/users-teams-organizations/teams.html.md
```diff
@@ -4,14 +4,83 @@ page_title: "Teams - Terraform Enterprise"
...
+
+To delete a team, go to the target team's settings page and click the "Delete TEAM NAME" button.
+
+### Managing Team Membership
+
+-> **API:** See the [Team Members API](../api/team-members.html).
+
+Organization owners can use a team's settings page to add and remove users from the team.
+
+The structure of your TFE teams should resemble your company's group structure.
```
armchairlinguist: This seems very prescriptive, do we need it? I think if I was going to introduce this at all, it would be "Typically, teams will mirror your company's group structure. For example..." 

but we don't really get questions about this, in practice. The only questions we get are when permission levels don't meet people's needs.

content/source/docs/enterprise/users-teams-organizations/teams.html.md
```diff
@@ -4,14 +4,83 @@ page_title: "Teams - Terraform Enterprise"
...
+
+To add a user, enter their username in the "Username" text field (located under the "Add a New Team Member" header) and click the "Add member" button. (You must know the user's TFE username; users cannot be added using email addresses or other personal information.)
+
+To remove a user, click the "🗑" (trash can) button by their entry in the member list.
+
+### API Tokens
+
+-> **API:** See the [Team Tokens API](../api/team-tokens.html).
+
+Each team can have a special service account API token that is not associated with a specific user. You can manage this API token from a team's settings page. See [Service Accounts](./service-accounts.html) for more information.
```
armchairlinguist: In this case I would say "from the team's settings page" because "each team" establishes a specific frame of reference. But again, do as you will!

content/source/docs/enterprise/users-teams-organizations/teams.html.md
```diff
@@ -4,14 +4,83 @@ page_title: "Teams - Terraform Enterprise"
...
+A team can be given read, write, or admin permissions on one or more workspaces.
+
+- Use any workspace's "Access" tab to manage team permissions on that workspace. For full instructions, see [Managing Access to Workspaces](../workspaces/access.html).
+- For detailed information about the available permissions levels, see [Permissions](./permissions.html).
 
-To create a team, navigate to the "Teams" section of your organization settings.
-To give a team permissions on a specific workspace, navigate to the "Access
-settings" on the workspace and add the team. A team can be given read, write, or
-admin [permissions on a workspace](./permissions.html).
+When determining whether a user can take an action on a resource, TFE uses the highest permission level from that user's teams. (For example, if a user belongs to a team with read permissions on a workspace and another team with admin permissions on that workspace, that user has admin permissions.)
```
armchairlinguist: Very nice callout here.

content/source/docs/enterprise/users-teams-organizations/users.html.md
```diff
@@ -36,6 +36,10 @@ To add a user to an organization, a member of that organization's owners team mu
 
 Adding a user to a team requires only their username.
 
+## Site Admin Permissions
+
+On private installs of TFE, some user accounts have a special "site admin" permission that allows administration of the entire instance. For more information, see [Administering Private Terraform Enterprise](../private/admin/index.html).
```
armchairlinguist: I'd prefer we use "installations" or "instances" or "On Private TFE installations".

I'd also love to add a note here that site admin permissions are disjoint from organization membership permissions; admins have admin permission for everything, but regular user type permission only for those resources they have ordinary access to. This question commonly comes up with API token permissions, but it would be nice to clarify here generally.

content/source/docs/enterprise/workspaces/access.html.md
```diff
@@ -8,7 +8,7 @@ sidebar_current: "docs-enterprise2-workspaces-access"
 
 Terraform Enterprise (TFE) workspaces can only be accessed by users with the correct permissions. You can manage permissions for a workspace on a per-team basis.
 
-Managing access requires [admin privileges](../users-teams-organizations/permissions.html) on the affected workspace.
+Workspace access should be managed by [organization owners](../users-teams-organizations/teams.html#the-owners-team). (Although any user with [admin privileges](../users-teams-organizations/permissions.html) on the target workspace can edit its permissions, non-owners can only add permissions for teams they are members of.)
```
armchairlinguist: I think "on a workspace" makes more sense here than "on the target workspace".

chrisarcand: What's the reasoning for this tweak? Honestly I think the original is more clear to me than the change (I had to reread a bunch to figure out the distinction you're making here, but then it seems obvious now that I understand it). It seems like this is just a recommendation. As an owner user, isn't it plausible that I give admin access to someone with the intention of them managing that workspace's access for others thereafter?

content/source/docs/enterprise/users-teams-organizations/teams.html.md
nfagerlund: Ah, good point.

content/source/docs/enterprise/users-teams-organizations/teams.html.md
nfagerlund: ha, one of those passages where I reconsidered each half of it separately and it got weird. Let's go with

> Only organization owners can manage teams or view the full list of teams. Other users can view the teams page in read-only mode and view any teams they are members of.

content/source/docs/enterprise/users-teams-organizations/teams.html.md
nfagerlund: I do think callouts (or some other visually-distinct unit) are the way here, and have been planning to add more throughout the docs. 

My reasoning is that if you end up on a UI document but you actually want to be on an API page (probably because the UI docs ended up with more google juice), I want to send you there with a method that's 1: consistent, and 2: doesn't require actually reading the page (since this isn't even the page you want to be on). The problem with linking from an undistinguished sentence is that you have to read way more text before you find the link that was supposed to save you from having to read all this text. 

Basically, my mental model of switching between ui/api pages is that it's still part of the process of navigating to the right page, rather than the process of understanding the content, and as such it should happen in skimming-mode instead of reading-mode. I could be wrong about that, but I feel pretty confident about it. 

y/n?

content/source/docs/enterprise/users-teams-organizations/teams.html.md
nfagerlund: One of the few holdovers from the previous iteration of these pages, and it came in when everyone involved was neck-deep in the "recommended practices guide" and very much in a "we must explain to people how to think about this stuff" mode. 

Eh, sure, may as well soften it up a bit. I can just link out to the reccos guide, too.

content/source/docs/enterprise/users-teams-organizations/users.html.md
nfagerlund: Good point.

content/source/docs/enterprise/workspaces/access.html.md
nfagerlund: I went with 

> Workspace access should be managed by [organization owners](../users-teams-organizations/teams.html#the-owners-team). (Users with [admin privileges](../users-teams-organizations/permissions.html) on a workspace can make minor adjustments to its permissions, but only owners can manage permissions for any team in the organization.)

content/source/docs/enterprise/users-teams-organizations/teams.html.md
armchairlinguist: Yep, seems fine! I don't know if I'd do it, but I don't feel strongly about it, and I see the argument for consistency.

nfagerlund:
🙌🏼

PR #463, opened 2018-09-14
docs/sentinel: examples, add aws_ami owner enforcement

vancluever:
This adds the enforce-ami-owners.sentinel aws_ami enforcement example
added in hashicorp/terraform-guides#54 to the Sentinel policy examples
list.

For some extra background on the nature of the policy and the reason for
its necessity, see the referenced PR.

### DIFF
--- a/content/source/docs/enterprise/sentinel/examples.html.md
+++ b/content/source/docs/enterprise/sentinel/examples.html.md
@@ -12,6 +12,7 @@ and are far from exhaustive.
 
 ### Amazon Web Services
 
+* [Enforce owner allow list on `aws_ami` data source](https://github.com/hashicorp/terraform-guides/blob/master/governance/aws/enforce-ami-owners.sentinel)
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/sentinel/examples.html.md
```diff
@@ -12,6 +12,7 @@ and are far from exhaustive.
 
 ### Amazon Web Services
 
+* [Enforce owner allow list on `aws_ami` data source](https://github.com/hashicorp/terraform-guides/blob/master/governance/aws/enforce-ami-owners.sentinel)
```
chrisarcand: This sounds a little weird but I tried to keep the list order 'super generic' => 'more specific', think this might make more sense at the end of the list?

chrisarcand: Cool, sounds good then!

PR #464, opened 2018-09-14
Switch deployment to netlify

jescalan:
This PR adds a terraform script added that provisions Netlify's configuration. This has been tested previously and is working nicely, PRs are also going out to all other project sites. Any of the variables are absolutely up for change if need be.

The Makefile has been updated to use Netlify's build image directly. The first run, this takes a bit longer to download due to it being larger than the hashicorp-middleman image, Netlify being a larger service that can deploy many types of sites. However, on subsequent runs its pretty quick and carries the additional benefit of cacheing dependencies between builds.

This PR also removes of all old deploy script pieces that I could find. If there are any I missed please let me know and I'll remove them.

We will be setting this up on HashiCorp's internal TFE instance to keep track of state and serve as a bit of a buffer to Netlify config changes, which can easily cause downtime if not done carefully.

---

*Update:* I took my best shot at updating all the Makefile scripts and they seem to be working, but it would be great to get a second perspective on it since I'm not super familiar with some of the provider-specific and testing tasks.


vancluever:
@jescalan I noticed that you mentioned a Makefile change - but I don't see any in the changeset?

jescalan:
Yeah, so for this specific one I didn't mess with the makefile yet, just because there are a bunch of extra custom scripts I want to make sure I understand and also upgrade them without messing anything up. If there's someone that would be willing to pair with me on this, that would be amazing.

nfagerlund:
Closing this; checked with the web platform team, and it makes sense to just wait for the platform transition.

PR #465, opened 2018-09-17
docs/sentinel: tfstate, fix broken link

vancluever:
The link to `null_data_source` was broken - it does not follow the normal
convention that most other Terraform resources/data sources follow.


cgriggs01:
LGTM, 

thanks for making the change!

PR #469, opened 2018-09-18
Describe format and extensions for key and cert

rberlind:


content/source/docs/enterprise/private/install-installer.html.md
```diff
@@ -29,7 +29,7 @@ Before you begin, you'll need to prepare data files and a Linux instance.
 ### Data Files
 
 * TLS private key and certificate
-  * The installer allows for using a certificate signed by a public or private CA. If you do not use a trusted certificate, your VCS provider will likely reject that certificate when sending webhooks.
+  * The installer allows for using a certificate signed by a public or private CA. If you do not use a trusted certificate, your VCS provider will likely reject that certificate when sending webhooks. The key and certificate should have the X.509 format. The key should have extension \*.pem or \*.key while the certificate should have extension \*.cert or \*.crt.
```
nfagerlund: X.509 isn't quite a file format! It describes the layout of the content for certs and CRLs, but that content can be encoded into a variety of formats, including PEM (base-64), PKCS#12, and others. 

From my understanding, the X.509 cert and the private key for the installer need to both be in PEM format (text files beginning with `---BEGIN PRIVATE KEY---` or whatever), as well as having the specified extensions... yeah? Or do we support other valid X.509 formats like PKCS#12?

rberlind: This was what the guy who was setting up certs on a PTFE server at customer said. 

It might be best to check with the PTFE team to determine the exact internal format and allowed file extensions.

nfagerlund: @hashicorp/private-tfe , can I get some guidance from someone on allowed cert formats?

rberlind: Based on some reading I did and Nick's comments, I think I should have written "The key and certificate should have ASCII PEM encoding."  Regarding the extension, I wonder if the certificate could also have the *.pem extension.

evanphx: Specifying the extensions the files are on the customers system is probably wrong, because there are plenty of systems that store both keys and certs as .pem for both.

Looking at the docs, I think it's safe to simplify specify that the certificate and key be PEM encoded. The other encoding is DER, which is just the raw binary version of the thing that PEM encodes.

rberlind: Thanks @evanphx.  I'll modify.

nfagerlund:
Sweet, thanks all!

rberlind:
Thanks

PR #474, opened 2018-09-20
Explicitly state how to override providers

rileykarson:
There may be a better home for this in these docs - just let me know if I should put it elsewhere/I missed documentation that already covers this!

I want to make it more clear how this directory interacts with HashiCorp distributed providers, allowing users to use their development copies instead of release versions if they are present.

### DIFF
--- a/content/source/docs/extend/how-terraform-works.html.md
+++ b/content/source/docs/extend/how-terraform-works.html.md
@@ -84,6 +84,9 @@ your user's home directory.
 
 [`terraform init`][3] will search
 this directory for additional plugins during plugin initialization.
+Providers that are distributed by HashiCorp will use the local copy
+in this directory instead of downloading a release copy during 
+`terraform init`. 
 
 The naming scheme for plugins is `terraform-<type>-NAME_vX.Y.Z`, where `type` is
 either `provider` or `provisioner`. Terraform uses the `NAME` to understand the


nfagerlund:
I think being more explicit about this makes sense. 👍🏼 Here's my thinking right now: 

- There's another occurrence of this text at https://www.terraform.io/docs/configuration/providers.html, which lives in the `website` folder of https://github.com/hashicorp/terraform. That could use new text as well, in a separate PR after we finish here.
- The text in this PR is still a little ambiguous, and I was about to suggest some adjustments when I realized I don't understand how this overriding interacts with provider versioning. Specifically: 
    - How does it work without a specified provider version? Will it always use the local plugin even if there's a newer one available on releases?
    - How does it work with an explicit version? Will it only use the local copy if it matches the version requirement, and get it from releases otherwise? (Hopefully it wouldn't use a "wrong" local version...)
    - How does it interact with version constraints if the local version meets the constraint, but there's a newer version on releases that also meets the constraint? 
    - Same as previous, but what happens during `terraform init -upgrade`? 

    I think those mostly boil down to whether Terraform see plugins in the local directory as being "installed," or "available."

    Do you (or anyone passing by) happen to know the answers to all that? I could probably rig up some tests to figure it out, if not.

rileykarson:
I don't know how it interacts with provider versioning and things `init -upgrade` either. My assumption is that the discovery directory always wins?

nfagerlund:
@rileykarson I did some fairly extensive investigation, and ended up revising a whole bunch of related content in two repos. 

I'm about to close this PR, and I'm submitting a new one that includes your edit in its commit history. Thanks for getting the ball rolling on this!

PR #476, opened 2018-09-21
add 'Issuer of assertion not found or multiple' to SAML troubleshooting

armchairlinguist:
We've run into this a few times, so I thought it was worth adding to the list.


nfagerlund:
:shipit:

PR #480, opened 2018-09-24
Update policy API docs for clarity

ryanuber:
1. Cleanup up an invalid example JSON
2. Fixed default enforcement mode which is hard-mandatory
3. Clarified the enforcements map, why it is an array, behavior of updating it, and the serialized results the API returns.


content/source/docs/enterprise/api/policies.html.md
```diff
@@ -68,7 +69,7 @@ Key path                         | Type            | Default          | Descript
           "mode": "hard-mandatory"
         }
       ],
-      "name": "my-example-policy",
```
nfagerlund: Good catch, and also I've _got_ to figure out a way to make this less difficult. Could probably rig up an editor plugin somehow.........

PR #482, opened 2018-09-26
Fix a paragraph that trailed off into nowhere

nfagerlund:
This paragraph ending in a colon used to be followed by instructions for a
manual snapshot-and-restore process, which was obsoleted by the installer's
automated snapshot/restore feature. The instructions were removed in commit
32a70f3f9, but the colon remained to confuse passerby.


nfagerlund:
@blalor, tagging you because you were the one who deleted that bit. 😛 Nothing else got lost along the way, right? It was just the manual snapshot thing?

PR #484, opened 2018-09-26
Update TFE Team Membership API docs to expect username

svendlarsen:
## Context
The docs currently state that the Team Membership API endpoints accept a user ID. The endpoints in fact accept a username, and do not accept a user ID, returning `400 Bad Request` when provided a user ID instead of a username.

## Changes
* Updated the docs to reflect that the API endpoints expect a username in the `data[].id` field.


nfagerlund:
Tested the API on production, and yup, you're right.

PR #485, opened 2018-09-26
Add initial capacity and performance planning guide

evanphx:


content/source/docs/enterprise/private/capacity.html.md
```diff
@@ -0,0 +1,71 @@
...
+
+By default, PTFE allocates 256 MB of memory to each Terraform run, with a default concurrency of 10 parallel runs.
+Therefore, by default PTFE requires 2.6 GB of memory reserved for runs.
+
+After factoring in the memory needed to run the base services that make up the application, the default memory footprint of PTFE is approximately 4 GB.
+
+### Settings
+
+The settings for per-run memory and concurrency are available in the dashboard on port 8800, on the Settings page, under the Capacity section. They can also be set via
+the [application settings JSON file when using the automated install procedure](https://www.terraform.io/docs/enterprise/private/automating-the-installer.html#available-settings).
```
armchairlinguist: @evanphx This is probably the one substantive change I made, since this was referring to the replicated.conf but I think it's in the app settings. Can you confirm?

evanphx: Oh yer. Yes! Thank you.

content/source/docs/enterprise/private/capacity.html.md
```diff
@@ -0,0 +1,71 @@
...
+provided by the Linux instance it is installed on. There are a few settings that allow the capacity that PTFE
+uses to be adjusted to suit the instance.
+
+## Memory + Concurrency
+
+The amount of memory to allocate to a Terraform run and the number of concurrent runs are the primary elements in
+understanding capacity above the base services.
+
+By default, PTFE allocates 256 MB of memory to each Terraform run, with a default concurrency of 10 parallel runs.
+Therefore, by default PTFE requires 2.6 GB of memory reserved for runs.
```
roooms: What happens if not enough memory is available? Does PTFE reserve less, raise a warning, or not work at all?

content/source/docs/enterprise/private/capacity.html.md
```diff
@@ -0,0 +1,71 @@
...
+
+The settings for per-run memory and concurrency are available in the dashboard on port 8800, on the Settings page, under the Capacity section. They can also be set via
+the [application settings JSON file when using the automated install procedure](https://www.terraform.io/docs/enterprise/private/automating-the-installer.html#available-settings).
+
+## Increasing Capacity
+
+To increase the number of concurrent runs, adjust the **Capacity** setting. Note that this setting is not limited by
+system checks; it depends on the operator to provide enough memory to the system to accommodate the requested
+concurrent capacity. For instance, if **Capacity** is set to `100`, the instance would require, at a minumum,
+26 GB of memory reserved for Terraform runs.
```
roooms: If the capacity is set to 100 but there is never more than 10 concurrent runs, what is the behaviour? This is somewhat linked to the above question, how does PTFE handle over-commitment that's under-used.

content/source/docs/enterprise/private/capacity.html.md
```diff
@@ -0,0 +1,71 @@
...
+
+The default memory limit of 256 MB per Terraform run is also configurable. Note that this setting is not limited by
+system checks; it depends on the operator to provide enough memory to the system to accommodate the requested limits.
+If the memory limit is adjusted to 512 MB with the default capacity of 10, the instance would require, at a minumum,
+5.2 GB of memory reserved for Terraform runs.
+
+### Downward Adjustment
+
+We do not recommend adjusting the memory limit below 256 MB. Memory is Terraform's primary resource and it
+becomes easy for it to go above smaller limits and be terminated mid-run by the Linux kernel.
```
roooms: Is this an OOM kill?

content/source/docs/enterprise/private/capacity.html.md
evanphx: It doesn't provide any feedback about that kind of misconfiguration atm.

content/source/docs/enterprise/private/capacity.html.md
evanphx: If the limit is set insanely high but only goes up to 10, then everything works fine. It's just that the system is basically at time bomb if it can't actually handle the high limit.

content/source/docs/enterprise/private/capacity.html.md
evanphx: Yes.

PR #495, opened 2018-10-02
add infobox for CentOS install guide on preflight page

armchairlinguist:
Puts an info box for the CentOS-specific guide below the RHEL warning box. This guide is not very visible, and this should improve its visibility.


armchairlinguist:
The failing checks are a missing link in the providers area, so not related to this.

PR #496, opened 2018-10-02
specify SSH requirement and clone behavior for BBS

armchairlinguist:
Provide clarity on the necessary connectivity between TFE and BBS.


armchairlinguist:
@nfagerlund ☎️ 😆

armchairlinguist:
Confirmed with @amy-hashi that this looks good. Check failure is a provider issue, unrelated.

PR #497, opened 2018-10-02
Reconcile docs about plugin discovery and downloading

nfagerlund:
I'm attempting to keep things simple for normal users while making sure we've
got the full behavior written down somewhere for plugin developers.

This PR doesn't stand alone; it's paired with a PR in the
terraform repo, to deal with a bunch of related content: https://github.com/hashicorp/terraform/pull/18973


nfagerlund:
This mass edit initially came out of https://github.com/hashicorp/terraform-website/pull/474.

PR #498, opened 2018-10-02
call out the browser port 8800 connection, resection

armchairlinguist:
We're finding that occasionally the browser connection is a concern, so I wanted to call it out more obviously, if that makes sense.

While there I noticed we have an extra "Installation" section so I removed the top one, as it was more a collection of installation-related configuration notes.


armchairlinguist:
Failure on this is not related to the changes, just a provider link issue.

PR #499, opened 2018-10-02
Add documentation for the `plan` permission

davidcelis:
This is a new permission level found between `read` and `write`, which allows users to create and plan a new run, but not apply it.


davidcelis:
The broken link in that test failure looks unrelated and was already failing, so I'm gonna merge

PR #503, opened 2018-10-03
Remove latest_run, which has been deprecated from the API

chrisarcand:
It hasn't actually been removed, but aliased to its replacement, current_run.
There's no need to ever use latest_run at this point, use current_run instead.

### DIFF
--- a/content/source/docs/enterprise/api/_template.md
+++ b/content/source/docs/enterprise/api/_template.md
@@ -93,8 +93,8 @@ This GET endpoint can optionally return related resources, if requested with [th
 Resource Name      | Description
 -------------------|------------
 `organization`     | The full organization record.
-`latest_run`       | Additional information about the last run.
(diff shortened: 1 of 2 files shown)


content/source/docs/enterprise/api/_template.md
```diff
@@ -93,8 +93,6 @@ This GET endpoint can optionally return related resources, if requested with [th
 Resource Name      | Description
 -------------------|------------
 `organization`     | The full organization record.
-`latest_run`       | Additional information about the last run.
-`latest_run.plan ` | The plan used in the last run.
```
nfagerlund: Could you please replace these with `current_run` equivalents, so the template will still have multiple resource type to show how to do it?

chrisarcand: Good call.

PR #505, opened 2018-10-04
Adding steps for updating license

petems:
Quick help page on updating licens

@koikonom I'm not sure of the Replicated version that was required for the cli upload?

### DIFF
--- a/content/source/docs/enterprise/private/updating-license.html.md
+++ b/content/source/docs/enterprise/private/updating-license.html.md
@@ -1,0 +1,39 @@
+---
+layout: "enterprise2"
+page_title: "Updating License Key"
+sidebar_current: "docs-enterprise2-private-installer-license"
+---
+
+# Updating License Key
+
+After renewing your commercial license for Private Terraform Enterprise (PTFE), follow the instructions below to update the license file for your installation.
+
+If you are missing the new license Key, please talk to your **Solutions Engineer** (pre-sales, POC or trial) or **Technical Account Manager** (existing customers).
+
+## Requirements
+
+* Shell access to the VM running Private Terraform Enterprise
+* The license file (with the extension `.rli`)
+
+## Steps
+
+### Online Installation
+
+For the online installation mode, it's not necessary for a customer to manually load a new license file to the PTFE servers. The "Sync License" process will update the license:
+
+1. In the PTFE Admin Console, select the "View License" menu under the gear icon (or navigate directly to <ptfe-hostname>:8800/license.
+1. Click the "Sync License" button.
+1. Go to the dashboard using the Dashboard menu.
+1. Restart the PTFE application in the dashboard.
+
+### Airgap installation
+
+For an Airgap installation, a new license file will be provided by the customer's Technical Account Manager or Solutions Engineer.
+
+1. Copy the new license file to the PTFE instance (for example with `scp`)
+2. `ssh` to the PTFE instance
+3. Load the license, by running `cat <license_file> | replicatedctl license-load`
+4. In the PTFE Admin Console, select the "View License" menu under the gear icon (or navigate directly to <host>:8800/license.
+5. Click the "Sync License" button.
+6. Go to the dashboard using the Dashboard menu.
+7. Restart the PTFE application in the dashboard.


rberlind:
The process described here does not seem entirely correct

In particular, as noted in some comments in the Asana ticket that asked for this documentation, https://app.asana.com/0/331272348117180/792178027569517, the process will differ depending on whether PTFE was installed in online or airgapped mode.  And I think the division between these two situations should be the primary division.

If the customer installed in online mode, then after someone at HashiCorp changes the expiration date of the license on https://vendor.replicated.com, all the customer has to do is the following:

1. In the PTFE Admin Console, select the "View License" menu under the gear icon (or navigate directly to <host>:8800/license.
2. Click the "Sync License" button.
3. Go to the dashboard using the Dashboard menu.
4. Restart the PTFE application in the dashboard.

Note that for the online installation mode, it is not necessary for anyone at HashiCorp to download and send the customer a new license file.  It is also not necessary for the customer to manually load a new license file to their PTFE servers.  That is done for them by the "Sync License" button.

For airgapped installations, I'm assuming that you might have gotten the instructions for updating the license from Replicated.  But I wonder for the situation where version is > `Version` whether syncing the license in the Admin Console is really needed; I guess it depend on what the "Sync License" button does: if it tries to sync the license from vendor.replicated.com, then I would think it would fail in airgapped installation and that it is not needed if the license was already scp-ed and then applied with `cat <license_file> | replicatedctl license-load`.  I strongly suspect that this command does what Sync License would do, making using that button unnecesary in this situation.

For the airgapped situation where license is < `Version`, it seems we need to add a step to restart the PTFE application in the Admin Console at the end.

Finally, did any PTFE versions that use Replicated actually use a version < `Version`? If not, we could avoid describing that procedure.

Overall, I recommend a structure that would be:
If Online
If Airgapped
  If version > `Version`
  If version < `Version`

koikonom:
@petems I've had a chat with the replicated folks recently and things have become a lot easier.

Specifically on an online installation you just click "refresh" and on an airgap installation you copy the license file over the old one and click, again, refresh license.

rberlind:
Could you be more specific @koikonom ?  There is no "refersh" button.  I assume you mean the "Sync License" button.  Also, does the PTFE app still need to be restarted (especially if the license had expired and all containers were stopped)?

And if we're going to say to replace license file with new one for airgap, we need to say where it is.  I searched once and could not find it.

rberlind:
(I originally posted this before the comment from @koikonom, but then deleted and re-posted without an internal HashiCorp link.)

The process described here does not seem entirely correct

In particular, as noted in some comments in the Asana ticket that asked for this documentation, , the process will differ depending on whether PTFE was installed in online or airgapped mode. And I think the division between these two situations should be the primary division.

If the customer installed in online mode, then after someone at HashiCorp changes the expiration date of the license on , all the customer has to do is the following:

In the PTFE Admin Console, select the "View License" menu under the gear icon (or navigate directly to :8800/license.
Click the "Sync License" button.
Go to the dashboard using the Dashboard menu.
Restart the PTFE application in the dashboard.
Note that for the online installation mode, it is not necessary for anyone at HashiCorp to download and send the customer a new license file. It is also not necessary for the customer to manually load a new license file to their PTFE servers. That is done for them by the "Sync License" button.

For airgapped installations, I'm assuming that you might have gotten the instructions for updating the license from Replicated. But I wonder for the situation where version is > Version whether syncing the license in the Admin Console is really needed; I guess it depend on what the "Sync License" button does: if it tries to sync the license from vendor.replicated.com, then I would think it would fail in airgapped installation and that it is not needed if the license was already scp-ed and then applied with cat <license_file> | replicatedctl license-load. I strongly suspect that this command does what Sync License would do, making using that button unnecessary in this situation.

For the airgapped situation where license is < Version, it seems we need to add a step to restart the PTFE application in the Admin Console at the end.

Finally, did any PTFE versions that use Replicated actually use a version < Version? If not, we could avoid describing that procedure.

Overall, I recommend a structure that would be:
If Online
If Airgapped
If version > Version
If version < Version

content/source/docs/enterprise/private/updating-license.html.md
```diff
@@ -0,0 +1,51 @@
+---
+layout: "enterprise2"
+page_title: "Updating License Key"
+sidebar_current: "docs-enterprise2-private-installer-rhel"
+---
+
+# Updating License Key
+
+This guide is for updating the License file for an installation of Private Terraform Enterprise.
```
nfagerlund: "license" and "license key" should be lowercase. 

This intro sentence should include a brief word about when an admin would want to/be able to update a license file.

content/source/docs/enterprise/private/updating-license.html.md
```diff
@@ -0,0 +1,51 @@
...
+
+If you are missing the new License Key, please talk to to your **Solutions Engineer** (pre-sales, POC or trial) or **Technical Account Manager** (existing customers).
+
+## Requirements
+
+* shell access to the VM running Private Terraform Enterprise
```
nfagerlund: capital "Shell"

content/source/docs/enterprise/private/updating-license.html.md
```diff
@@ -0,0 +1,51 @@
...
+* The license file (with the extension `.rli`)
+
+## Steps
+
+### If using Replicated > `Version`
```
nfagerlund: - Is this going to be replaced by a version number? 
- Would it make sense to use PTFE release numbers instead of Replicated versions, or is the Replicated version exposed clearly to admins? (I haven't yet installed a full PTFE instance, so I haven't used the Replicated dashboard.)

content/source/docs/enterprise/private/updating-license.html.md
```diff
@@ -0,0 +1,51 @@
...
+6. Go to the dashboard using the Dashboard menu.
+7. Restart the PTFE application in the dashboard.
+
+### If using Replicated < `Version`
+
+1. Copy the new license file to the PTFE instance (for example with `scp`)
+2. `ssh` to the PTFE instance
+3. Set the license file location with the command line: `replicatedctl params set LicenseFileLocation --value '/full/path/to/license.rli`
+4. Stop PTFE in the Web UI under `<host>:8800/`
+4. Stop Replicated:
```
nfagerlund: For these code blocks nested inside a list item:

- Add a blank line before and after the code block.
- Indent the entire code block 4 spaces, including the triple-backtick delimiters. 
- Use `shell` as the language, and prefix the lines with `$ `.

content/source/docs/enterprise/private/updating-license.html.md
```diff
@@ -0,0 +1,51 @@
+---
+layout: "enterprise2"
+page_title: "Updating License Key"
+sidebar_current: "docs-enterprise2-private-installer-rhel"
```
nfagerlund: This needs to be changed to something unique.

petems:
Hey, I was on PTO for a few weeks so missed this, will try and pickup on Monday! 👍

rberlind:
Hi @petems : Did you ever get to work on this?
Also, I think we still need clarification on whether one actually has to sync the license in the admin console if one uses the `cat <license_file> | replicatedctl license-load` command for airgap installations.

Roger

petems:
completely Forgot about this! Let me do some testing and see if I can test both see rails

content/source/docs/enterprise/private/updating-license.html.md
petems: Fixed 👍

content/source/docs/enterprise/private/updating-license.html.md
petems: Replicated uses a different release pattern than PTFE, the version is exposed either via the CLI or the replicated UI. 

I do need to find out the version number where this changed however...

content/source/docs/enterprise/private/updating-license.html.md
petems: Fixed! 👍

content/source/docs/enterprise/private/updating-license.html.md
petems: Copypasta error, fixed!

content/source/docs/enterprise/private/updating-license.html.md
petems: Done! 👍

petems:
@nfagerlund Made the style changes you requested. Still need to find out what version of replicated had the new improved way of doing things 😄 

@rberlind Ok, I've added the split between Airgap and Online, I didn't realise it could auto-sync with an online install, that's pretty neat. Can you review?

@koikonom Is it possible to find out from Replicated engineers when the process changed?

content/source/docs/enterprise/private/updating-license.html.md
```diff
@@ -0,0 +1,69 @@
...
+## Requirements
+
+* Shell access to the VM running Private Terraform Enterprise
+* The license file (with the extension `.rli`)
+
+## Steps
+
+### Online Installation
+
+> For the online installation mode, it is not necessary for a customer to manually load a new license file to their PTFE servers. That is done for them by the "Sync License" button.
```
nfagerlund:
```suggestion
For the online installation mode, it is not necessary for a customer to manually load a new license file to their PTFE servers. That is done for them by the "Sync License" button.
```

content/source/docs/enterprise/private/updating-license.html.md
```diff
@@ -0,0 +1,69 @@
...
+
+1. In the PTFE Admin Console, select the "View License" menu under the gear icon (or navigate directly to <ptfe-hostname>:8800/license.
+1. Click the "Sync License" button.
+1. Go to the dashboard using the Dashboard menu.
+1. Restart the PTFE application in the dashboard.
+
+### Airgap installation
+
+> For an Airgap installation, a new license file will have to be provided by the customer's Technical Account Manager or Solutions Engineer.
```
nfagerlund:
```suggestion
For an Airgap installation, a new license file will have to be provided by the customer's Technical Account Manager or Solutions Engineer.
```

content/source/docs/enterprise/private/updating-license.html.md
```diff
@@ -0,0 +1,69 @@
+---
+layout: "enterprise2"
+page_title: "Updating License Key"
+sidebar_current: "docs-enterprise2-private-installer-license"
+---
+
+# Updating License Key
+
+This guide is for updating the license file for an installation of Private Terraform Enterprise. You will want to do before a license expiry, such as after a commercial renewal.
```
nfagerlund:
```suggestion
After renewing your commercial license for Private Terraform Enterprise (PTFE), follow the instructions below to update the license file for your installation.
```

kawsark:
Hello @petems , were you able to find the version number where the process for Airgapped license update changes?

petems:
Not sure, but it might've been long enough that we just KISS and only show the latest instructions? 

@koikonom Any thoughts?

koikonom:
Agreed! The expectation is that customers will regularly update the replicated installer in addition to the app itself.

petems:
Cool, I'll take out the older part.

rberlind:
Have we actually tested the airgap procedure @petems or @koikonom ?  A customer got an error when running `replicatedctl license-load` which said "Error: response from daemon: Online installation is not possible when running in airgapped mode.".  I hope that is because they specified the license file with `cat *.rli  |replicatedctl license-load`.  I told him to use `cat <license>.rli | replicatedctl license-load`. 

I looked at https://help.replicated.com/api/replicatedctl/replicatedctl_license-load/ and see that the replicatedctl license-load has an --airgap-package argument.  Is it possible that this has to be used with airgapped installations to update the license?

rberlind:
Somehow, I can't find your most recent comment in the conversation in the
ticket reporting that syncing the license for online installation complains
about the license being expired, possibly because the PTFE containers are
not running, Kawsar. I presume that you did first extend the license on the
Replicated portal. Right?

An interesting test would be to stop the PTFE app with a license that is
not expired, extend the license on the replicated portal, then try to sync
license. If it fails, that would prove that the PTFE app does have to be
running in order to sync the license. Of course, that would be impossible
if the license had expired since we usually configure PTFE to stop the app
when the license expires.

But it does not make sense that the PTFE app would need to be running. When
we stop the PTFE app, the Replicated containers should still be running and
they are the ones that I would think are needed for syncing a license.

Roger

rberlind:
This still needs to be merged, @petems and @koikonom

petems:
Ok, I've finally cut out the version specific stuff

rberlind:
So, I just tested the airgapped process and found it quite wrong.  In particular, I could not use the `replicatedctl license-load` command.  When I ran `cat <new_license>.rli | replicatedctl license-load`, I got the error "Error: response from daemon: Online installation is not possible when running in airgapped mode."  When I ran `cat <new_license>.rli | replicatedctl license-load --airgap-package Terraform-Enterprise-350.airgap`, I got "Error: response from daemon: Invalid package archive".

I'm fairly sure my license and airgap bundle are fine.  In particular, I tested my license with `file <new_license>.rli` and saw the expected "POSIX tar archive". 

Also, I tried running as ec2-user and as root on a RHEL EC2 instance.

So, I do not believe that this command can be used to update the license for an airgapped installation.

However, I was able to upgrade my license as follows, based on this article: https://help.replicated.com/community/t/how-can-a-customer-update-an-airgap-license/182

I did the following after copying the new license to the server:
1. I visited the console/settings page of the PTFE admin console and looked at the path for the license.
2. I then ran `mv <new_license>.rli <old_license>.rli` to overwrite the old license.
3. On the license page of the PTFE admin console, I then clicked the Sync License button.

My license was successfully updated.

I also validated another approach which was to change the path to the license file on the console/settings page of the admin console and then click the Sync License button on the license page.

So, the second of these approaches might be better since everything is done in the admin console after copying the new license file to the server.

I should note that I generally append the expiration date to the name of the license files.  So, for me, a new license would have a different name than the old license.  But the new licenses generated by Replicated would have the same name as the old license if one did not change the names.

rberlind:
I tried again today without renaming the license file either when downloading or when writing to disk in my user-data script.  I still got the same errors.  So, the fact that I renamed the license does not seem to be a factor. @kawsark reported success with the `replicatedctl license-load` command but said he was using an older license file.  I think I saw something in the #team-tfe-support channel that Replicated might have changed their license format recently.  Perhaps the command does not work with the new format?

nfagerlund:
It sounds like this needs more research and a new set of instructions, so I'm going to close this PR.

PR #507, opened 2018-10-05
Change some language about updating sensitive variables

beekus:
We no longer allow changing the `hcl` flag nor the `key` for an already persisted sensitive variable.

[TFE Changes PR](https://github.com/hashicorp/atlas/pull/6953)

### DIFF
--- a/content/source/docs/enterprise/workspaces/variables.html.md
+++ b/content/source/docs/enterprise/workspaces/variables.html.md
@@ -68,7 +68,7 @@ To protect these secrets, you can mark any any Terraform or environment variable
 
 Marking a variable as sensitive prevents anybody (including you) from viewing its value in TFE's UI or API.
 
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/workspaces/variables.html.md
```diff
@@ -68,7 +68,7 @@ To protect these secrets, you can mark any any Terraform or environment variable
 
 Marking a variable as sensitive prevents anybody (including you) from viewing its value in TFE's UI or API.
 
-Users with edit permissions can set new values for sensitive variables. However, no one can change a sensitive variable into a non-sensitive one. (If a variable was accidentally marked sensitive, you can delete it and create a new variable to replace it.)
+Users with edit permissions can set new values for sensitive variables. However, no one can change a sensitive variable's key, hcl flag, nor change a senstive variable into a non-sensitive one. (If a variable was accidentally marked sensitive, you can delete it and create a new variable to replace it.)
```
armchairlinguist: @beekus I had the impression that we were changing this so no updating would be possible?

beekus: we are allowing values to be changed, but nothing else.

beekus: That is so much clearer. Thanks!

PR #510, opened 2018-10-09
Update link to installer bootstrapper to use install.terraform.io

blalor:
We now have a lambda function fronted by api gateway that translates the URL and redirects to the latest Replicated installer.

ref: https://app.asana.com/0/331272348117180/720172589230144/f

We don't have a section for linking to the version-specific installer, however; should we include that, or communicate it to customers privately?

### DIFF
--- a/content/source/docs/enterprise/private/install-installer.html.md
+++ b/content/source/docs/enterprise/private/install-installer.html.md
@@ -137,7 +137,7 @@ If the instance cannot reach the Internet, follow these steps to begin an Airgap
 1. Download the `.airgap` file using the information given to you in your setup email and place that file somewhere on the the instance.
     * If you use are using `wget` to download the file, be sure to use `wget --content-disposition "<url>"` so the downloaded file gets the correct extension.
     * The url generated for the .airgap file is only valid for a short time, so you may wish to download the file and upload it to your own artifacts repository.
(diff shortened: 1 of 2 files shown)


armchairlinguist:
This is a nice change! Feels very official. :)

If it is possible to link to a specific version, I think we could add a Note (info-level, blue, `->`) below the last step in the "Prepare the instance" section explaining how to do that. I'm not opposed to leaving it out, but I think I'd rather put it in if we can.

nfagerlund:
I second Alexis' suggestion for direct-to-version links in an info box. 

I'm glad y'all got that proxy working, that's great!

blalor:
Right now we don't really track available versions for anyone, and the list of versions would come from https://release-notes.replicated.com.  I can add a note along the lines of:

> To download a specific version, use a URL like `https://install.terraform.io/airgap/<major>.<minor>.<point>.tar.gz` …

But that will invite the question "ok, so where do I find supported/released versions?" and it doesn't feel correct to send customers to an external site.  So that's why I didn't add it.  Suggestions on how to proceed?

armchairlinguist:
@blalor I think we usually suggest contacting support when people have doubts or questions, I'd be fine with that here.

nfagerlund:
Same

PR #513, opened 2018-10-10
fixed URL used to download providers

rberlind:
I changed registry.terraform.io which is the public module registry to releases.hashicorp.com which is where providers are currently downloaded from.  There are plans to add providers to the public registry, so this could change back to registry.terraform.io in the future.


rberlind:
Fixed

rberlind:
See https://github.com/hashicorp/go-checkpoint/blob/master/checkpoint.go for verification that the checkpoint-api.hashicorp.com URL is correct @armchairlinguist.  That was where I figured that out.

The other change was obvious and consistent with other doc at https://www.terraform.io/docs/extend/how-terraform-works.html#selecting-plugins.

PR #515, opened 2018-10-12
Change Name Format to Basic, note boolean in text

armchairlinguist:
Okta doesn't have Boolean as an actual format, so this should just be Basic.


armchairlinguist:
@nfagerlund it does!

PR #525, opened 2018-10-18
docs/sentinel: Update limitation docs for all mocks

vancluever:
We've discovered that we overlooked the fact that mock count indexes are
represented as strings in the JSON, even though the indexes are integers
in practice.

This can't be necessarily worked around, unfortunately; removing the
indexes and moving the resource objects to an array, for example, will
create edge cases in testing if someone is using a singular-context for
or any/all loop. Hence, we are keeping the mocks the way that they are,
and just noting the limitations.

I've also revised the limitations so that they are not info boxes, but
but rather simple bullet points. tfconfig, even though it's not affected
by this, has been re-formatted as well to stay consistent.

### DIFF
--- a/content/source/docs/enterprise/sentinel/import/mock-tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/mock-tfconfig.html.md
@@ -22,9 +22,11 @@ documentation][ref-official-sentinel-documentation].
 [ref-mocking-imports]: https://docs.hashicorp.com/sentinel/writing/imports#mocking-imports
 [ref-official-sentinel-documentation]: https://docs.hashicorp.com/sentinel/
 
--> **NOTE:** As functions cannot be mocked in the current Sentinel testing
-framework, the [module()][ref-module] function is not available. As a result,
-only root module data can be mocked at this time.
+#### Current mock limitations
+
+As functions cannot be mocked in the current Sentinel testing framework, the
+[module()][ref-module] function is not available. As a result, only root module
+data can be mocked at this time.
 
 [ref-module]: /docs/enterprise/sentinel/import/tfconfig.html#function-module-
 
--- a/content/source/docs/enterprise/sentinel/import/mock-tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/mock-tfconfig.html.md
@@ -24,7 +24,7 @@ documentation][ref-official-sentinel-documentation].
 
 #### Current mock limitations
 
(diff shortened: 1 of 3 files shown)


vancluever:
Thanks @davidcelis! I forgot they aren't a part of the governance team and yeah their feedback will be valuable :slightly_smiling_face:

content/source/docs/enterprise/sentinel/import/mock-tfplan.html.md
```diff
@@ -29,12 +29,25 @@ mock][ref-tfconfig-mock] and the [`tfstate` mock][ref-tfstate-mock].
...
+
+* As functions cannot be mocked in the current Sentinel testing framework, the
+  [module()][ref-module] function is not available. As a result, only root
+  module data can be mocked at this time.
 
 [ref-module]: /docs/enterprise/sentinel/import/tfplan.html#function-module-
 
+* [Resource and data source][resource-and-data-source] count keys (`NUMBER` in
+  `TYPE.NAME[NUMBER]`), which are actually represented as integers, cannot be
+  represented accurately JSON and as a result, mocks that depend on count keys
```
armchairlinguist:
```suggestion
  represented accurately in JSON, and as a result, mocks that depend on count keys
```

content/source/docs/enterprise/sentinel/import/mock-tfconfig.html.md
```diff
@@ -22,9 +22,11 @@ documentation][ref-official-sentinel-documentation].
 [ref-mocking-imports]: https://docs.hashicorp.com/sentinel/writing/imports#mocking-imports
 [ref-official-sentinel-documentation]: https://docs.hashicorp.com/sentinel/
 
--> **NOTE:** As functions cannot be mocked in the current Sentinel testing
-framework, the [module()][ref-module] function is not available. As a result,
-only root module data can be mocked at this time.
+#### Current mock limitations
+
+As functions cannot be mocked in the current Sentinel testing framework, the
```
armchairlinguist:
```suggestion
* As functions cannot be mocked in the current Sentinel testing framework, the
```

armchairlinguist: Changing from NOTEs here to headings makes it kind of look as if the mock itself has moved down a level. I think the list format helps counteract that, so even though there's only one bullet here, I'd suggest sticking with that convention.

PR #536, opened 2018-10-25
Document TFE_PARALLELISM, which support sometimes uses to solve weird…

nfagerlund:
… problems

This is also a good place to mention CONFIRM_DESTROY, which isn't documented
anywhere but the UI text at the moment.

Notably, I'm omitting ATLAS_TOKEN, since it should really only be used when
upgrading to v1 to v2 (which will stop being possible soon).


content/source/docs/enterprise/workspaces/variables.html.md
```diff
@@ -90,6 +90,22 @@ Do not commit a file named `terraform.tfvars` to version control, since TFE will
...
+
+The following variables are reserved for internal use:
+
+- `TFE_TOKEN` and `ATLAS_TOKEN` — Reserved for passing the ephemeral [per-run API token](../run/index.html#backend-details), which is used to read and write state and to access private modules.
+
+    ~> **Important:** Do not set a value for either of these variables except when [upgrading from the legacy version of TFE](../upgrade/index.html) or as instructed by HashiCorp support. Overriding the run token can interfere with TFE's operation.
+- `GIT_SSH_COMMAND`, `TF_APPEND_USER_AGENT`, `TF_FORCE_LOCAL_BACKEND`, `TERRAFORM_CONFIG`, `CHECKPOINT_DISABLE`, `TF_IN_AUTOMATION`, and `TF_INPUT` — Reserved for setting up Terraform's run environment. TFE will override any user-provided values.
```
svanharmelen: I have mixed feelings about making these details public like this. Are we very sure that we will always override and/or ignore any user configured values for any of these variables? I understand that we should, but did anybody check that? Because if not, we are asking for potential problems by putting this in the docs.

content/source/docs/enterprise/workspaces/variables.html.md
```diff
@@ -90,6 +90,22 @@ Do not commit a file named `terraform.tfvars` to version control, since TFE will
...
+
+The following variables can be set by a user to manage a workspace's run behavior:
+
+- `CONFIRM_DESTROY` — If this environment variable is set to `1` in a workspace, an admin user can destroy all of the infrastructure managed by the workspace using the "Queue destroy plan" button in the settings page. The UI text for the destroy plan button includes a reminder about this safety measure.
+- `TFE_PARALLELISM` — If present, TFE uses this to set `terraform apply`'s `-parallelism=<N>` flag ([more info](/docs/internals/graph.html#walking-the-graph)). Valid values are between 1 and 256, inclusive; the default is `10`. This is rarely necessary, but is available as a workaround for infrastructure providers that error on concurrent operations or implement rate limiting in a non-standard way. We recommend talking to HashiCorp support before using this.
+
+The following variables are reserved for internal use:
+
+- `TFE_TOKEN` and `ATLAS_TOKEN` — Reserved for passing the ephemeral [per-run API token](../run/index.html#backend-details), which is used to read and write state and to access private modules.
```
svanharmelen: This is both partially incorrect and could maybe be considered sensitive information that we don't want to put on the website? See here for more details about why this is partially incorrect: https://hashicorp.slack.com/archives/C6PE3FELQ/p1540549526006200

content/source/docs/enterprise/workspaces/variables.html.md
nfagerlund: I didn't run experiments in a real environment, but yes, I did follow the code paths:

- GIT_SSH_COMMAND, TF_APPEND_USER_AGENT, TF_FORCE_LOCAL_BACKEND, and TERRAFORM_CONFIG are all [overwritten by terraform-build-worker](https://github.com/hashicorp/terraform-build-worker/blob/master/worker.go), no matter what.
- CHECKPOINT_DISABLE, TF_IN_AUTOMATION, and TF_INPUT... well, I might have been wrong about these; the Atlas code is harder to follow than the build worker. They're set [here](https://github.com/hashicorp/atlas/blob/master/app/service-objects/default_var_generator.rb), but now that I'm looking at [models-new/run.rb](https://github.com/hashicorp/atlas/blob/master/app/models-new/run.rb#L432), it looks like a user might be able to override those. 

I can replace the "will override" comment with "don't set values for these." Would that address your concerns?

svanharmelen:
Very nice, thanks!

PR #538, opened 2018-10-29
update symlink for terraform provider

mildwonkey:


nfagerlund:
This must have come up in slack in a room I wasn't watching, but JUST IN CASE someone in the future is confused about this PR: it's because the standalone version of the terraform terraform provider got merged back into the main terraform repo as of 0.11. https://github.com/terraform-providers/terraform-provider-terraform

nfagerlund:
I think there's two more pieces of cleanup to do, though. PR incoming.

PR #543, opened 2018-11-02
[PTFE] Custom TBW image instructions

koikonom:


content/source/docs/enterprise/private/install-installer.html.md
```diff
@@ -106,6 +106,29 @@ The UI to upload these certificates looks like:
 
 ![ptfe-ca-ui](./assets/ptfe-ca-bundle.png)
 
+
+## Alternative Terraform worker image
+
+TFE runs terraform plan and apply operations in a disposable docker container. There are cases where users may require additional tools that are not available in the default docker image. To work around this limitation users can build their own image and configure TFE to use that instead. In order for this to happen the name of the alternative docker image must be set in the config by using the relevant field as shown below:
```
armchairlinguist:
```suggestion
TFE runs `terraform plan` and `terraform apply` operations in disposable Docker containers. There are cases where runs may make frequent use of additional tools that are not available in the default Docker image. To allow use of these tools for any plan or apply, users can build their own image and configure TFE to use that instead. 

In order to use an alternative Docker image, the image name must be set in the config by using the [Name] field as shown below:
```

armchairlinguist: `[Name]` here should be replaced by the name of the field (it's important that it be in the text, not only in an image).

content/source/docs/enterprise/private/install-installer.html.md
```diff
@@ -106,6 +106,29 @@ The UI to upload these certificates looks like:
...
+
+![ptfe-docker-image](./assets/ptfe-docker-image.png)
+
+During the first iteration of this feature it is important to impose certain limitations on what we can support. These limitations are:
```
armchairlinguist:
```suggestion
### Requirements
```

content/source/docs/enterprise/private/install-installer.html.md
```diff
@@ -106,6 +106,29 @@ The UI to upload these certificates looks like:
...
+ - The base image must be ubuntu:xenial
+ - The user is responsible for making sure the image exists on the PTFE host by running `docker pull` from their own registry, or any other similar method.
```
armchairlinguist:
```suggestion
 - The image must exist on the PTFE host. It can be added by running `docker pull` from a local registry or any other similar method.
```

content/source/docs/enterprise/private/install-installer.html.md
```diff
@@ -106,6 +106,29 @@ The UI to upload these certificates looks like:
...
 
+
+## Alternative Terraform worker image
+
+TFE runs terraform plan and apply operations in a disposable docker container. There are cases where users may require additional tools that are not available in the default docker image. To work around this limitation users can build their own image and configure TFE to use that instead. In order for this to happen the name of the alternative docker image must be set in the config by using the relevant field as shown below:
+
+![ptfe-docker-image](./assets/ptfe-docker-image.png)
+
+During the first iteration of this feature it is important to impose certain limitations on what we can support. These limitations are:
+ - The base image must be ubuntu:xenial
```
armchairlinguist:
```suggestion
 - The base image must be `ubuntu:xenial`.
```

content/source/docs/enterprise/private/install-installer.html.md
```diff
@@ -106,6 +106,29 @@ The UI to upload these certificates looks like:
...
+ - The user is responsible for making sure the image exists on the PTFE host by running `docker pull` from their own registry, or any other similar method.
+ - The user is responsible for making sure CA certificates are available when terraform runs.
```
armchairlinguist: Let's say a bit more here - even more than the example below. I'd suggest that we explain which certs are likely to be needed and how we recommend injecting them.

content/source/docs/enterprise/private/install-installer.html.md
```diff
@@ -106,6 +106,29 @@ The UI to upload these certificates looks like:
...
+## Alternative Terraform worker image
+
+TFE runs terraform plan and apply operations in a disposable docker container. There are cases where users may require additional tools that are not available in the default docker image. To work around this limitation users can build their own image and configure TFE to use that instead. In order for this to happen the name of the alternative docker image must be set in the config by using the relevant field as shown below:
+
+![ptfe-docker-image](./assets/ptfe-docker-image.png)
+
+### Requirements
+ - The base image must be `ubuntu:xenial`.
+ - The image must exist on the PTFE host. It can be added by running `docker pull` from a local registry or any other similar method.
+ - The user is responsible for making sure CA certificates are available when terraform runs.
```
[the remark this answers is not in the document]
maryelizbeth: after this i'd second alexis on expanding which certs and how to inject.

content/source/docs/enterprise/private/install-installer.html.md
```diff
@@ -106,6 +106,29 @@ The UI to upload these certificates looks like:
...
+ - Terraform must not be installed in the image. TFE will take care of that on runtime.
```
maryelizbeth:
```suggestion
 - Terraform must not be installed on the image. TFE will take care of that at runtime.
```

content/source/docs/enterprise/private/install-installer.html.md
koikonom: I added a small sentence explaining what we expect the file to contain. If we want to expand on what a CA cert is and which CAs we expect customers to include then this should be done on a different page, since this is also relevant to other configuration settings as well.

content/source/docs/enterprise/private/install-installer.html.md
```diff
@@ -106,6 +106,29 @@ The UI to upload these certificates looks like:
...
+## Alternative Terraform worker image
+
+TFE runs `terraform plan` and `terraform apply` operations in a disposable Docker containers. There are cases where runs may make frequent use of additional tools that are not available in the default Docker image. To allow use of these tools for any plan or apply, users can build their own image and configure TFE to use that instead. In order for this to happen the name of the alternative docker image must be set in the config by using the `Custom image tag` field as shown below:
+
+![ptfe-docker-image](./assets/ptfe-docker-image.png)
+
+### Requirements
+ - The base image must be `ubuntu:xenial`.
+ - The image must exist on the PTFE host. It can be added by running `docker pull` from a local registry or any other similar method.
+ - CA certificates must available when terraform runs. During image creation a file containing all necessary PEM encoded CA certificates must be placed in `/etc/ssl/certs/ca-certificates.crt`.
```
armchairlinguist:
```suggestion
 - CA certificates must available when terraform runs. During image creation, a file containing all necessary PEM encoded CA certificates must be placed in `/etc/ssl/certs/ca-certificates.crt`. 
```

armchairlinguist:
@koikonom For the sake of getting something out that covers the basic parameters, I feel like this documentation is probably sufficient - I only had one comma suggestion. I agree that we should probably factor out the SSL information both for this and in general (which coincidentally is right above this in the page) and make an entirely separate page(s) for SSL info and troubleshooting. I'll follow up on that internally.

content/source/docs/enterprise/private/install-installer.html.md
blalor: Building on Alexis':

```suggestion
 - CA certificates must be available when terraform runs. During image creation, a file containing all necessary PEM encoded CA certificates must be placed in `/etc/ssl/certs/ca-certificates.crt`. 
```

PR #548, opened 2018-11-07
Corrects spelling of "Enterprise" on home page banner

topfunky:
The small banner on the home page was missing an _r_. This corrects it.

screen shot 2018-11-07 at 1 13 30 pm


jlsuttles:
Nice catch!

PR #550, opened 2018-11-08
Update Bitbucket Server version support note w/ Atlassian EOL policy

nfagerlund:
(My memory might be flaking on me, but I think @rkst was the one who had the most input on this in the sync?) 

Context for this is that we have specific customer reasons for not eagerly moving that support bar up to 4.11 today (or 4.12 tomorrow, lol), but it's worth reminding people that you're taking a loan out from your future self whenever you continue using EOLed software. 

Anyway, how's that wording?


nfagerlund:
@armchairlinguist Yeah, word. :| Luckily, most people should only need to encounter this page once, so at least that doesn't feel too urgent to me.

PR #552, opened 2018-11-12
Mention upload limits on Sentinel policy uploads

ryanuber:
Adds mention of the new limitations on policy upload sizes. This is purposely not mentioned in the UI-level docs, because right now the UI is a text box that a user would have to copy/paste more than 10mb of text to in order to exceed the limit. If you think it should still be mentioned anyways just let me know!

### DIFF
--- a/content/source/docs/enterprise/api/policies.html.md
+++ b/content/source/docs/enterprise/api/policies.html.md
@@ -205,6 +205,8 @@ This endpoint uploads code to an existing Sentinel policy.
 
 -> **Note**: This endpoint does not use JSON-API's conventions for HTTP headers and body serialization.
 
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/api/policies.html.md
```diff
@@ -205,6 +205,8 @@ This endpoint uploads code to an existing Sentinel policy.
 
 -> **Note**: This endpoint does not use JSON-API's conventions for HTTP headers and body serialization.
 
+-> **Note**: This endpoint limits the size of uploaded policies to 10MB. If a larger payload is uploaded, an HTTP 413 error will be returned, and the policy will not be saved.
+
```
armchairlinguist: I wonder if it's worth providing a suggestion to refactor the policy into multiple smaller policies?
Otherwise, looks good.

ryanuber: Added!

PR #554, opened 2018-11-14
Move some content to learn.hashicorp.com

nfagerlund:
(To be merged on @tr0njavolta's instructions after the redirects are active.)

This is a PR to transition and redirect some getting started and guides content to https://learn.hashicorp.com/terraform.

Companion PR in terraform: https://github.com/hashicorp/terraform/pull/19365

### DIFF
--- a/content/source/layouts/layout.erb
+++ b/content/source/layouts/layout.erb
@@ -69,6 +69,7 @@
               <nav class="navigation-links" role="navigation">
                 <ul class="main-links nav navbar-nav navbar-right">
                   <li><a href="/intro/index.html">Intro</a></li>
+                  <li><a href="https://learn.hashicorp.com/terraform/">Learn</a></li>
                   <li><a href="/docs/index.html">Docs</a></li>
                   <li><a href="/guides/index.html">Guides</a></li>
                   <li><a href="/docs/extend/index.html">Extend</a></li>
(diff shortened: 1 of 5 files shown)


alvin-huang:
Since the Writing Custom Providers Guide is moving to learn, can we remove https://github.com/hashicorp/terraform-website/blob/bd09c5b4667bca78e6611643438ca08540eb8d5a/content/redirects.txt#L169?

nfagerlund:
@alvin-huang Rachel might have newer info, but the last I heard was that it was going to live in both places for the time being, to keep an eye on where people are finding it most easily.

content/source/layouts/layout.erb
```diff
@@ -69,6 +69,7 @@
               <nav class="navigation-links" role="navigation">
                 <ul class="main-links nav navbar-nav navbar-right">
                   <li><a href="/intro/index.html">Intro</a></li>
+                  <li><a href="https://learn.hashicorp.com/terraform/">Learn</a></li>
```
nfagerlund: @tr0njavolta It's annoying, but you need to make an additional update in _sidebar.erb... because for some reason there's a separate version of this menu that gets shown as the hamburger menu at mobile widths. (You can see the discrepancy in the local preview if you shrink your browser window narrow enough.)

tr0njavolta: I've updated - thanks!

PR #555, opened 2018-11-14
Remove stateless reference in external Vault docs

glasner:
In response to 
2018-11-14 at 12 58 pm

https://hashicorp.slack.com/archives/C1UJDFJEP/p1541693037878100?thread_ts=1541692236.873100&cid=C1UJDFJEP

Note: EA team is currently updating reference architectures to remove external Vault as well (ex: https://www.terraform.io/docs/enterprise/private/gcp-setup-guide.html)

### DIFF
--- a/content/source/docs/enterprise/private/vault.html.md
+++ b/content/source/docs/enterprise/private/vault.html.md
@@ -13,10 +13,6 @@ encrypt sensitive information such as variables and states.
 An external Vault cluster allows the customer to have full control over how
 the Vault cluster is managed, for example how it is sealed and unsealed, replicated, etc.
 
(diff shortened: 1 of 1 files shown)


KFishner:
Given that we will be removing external Vault as the recommended option, I think we should put a comment on this vault page saying that this is not a recommended option and should only be used if a user wants control over vault operations (sealing, etc) and should not be used for improved availability. 

We also need to update the reliability & availability page to state that even in external services mode, Vault is internal. That needs to be updated in the tables and the external services section
https://www.terraform.io/docs/enterprise/private/reliability-availability.html#external-services

glasner:
Superseded by https://github.com/hashicorp/terraform-website/pull/614

PR #557, opened 2018-11-15
add note about internal monitoring and how to disable

armchairlinguist:
I didn't find an ideal place to put this (it might be nice to at some point have a configuration page that included details of post-install config adjustments?) but this seemed reasonably related, and I want to make sure that people have the information they need on this functionality.

This can be updated if we adjust the port we're running on.

### DIFF
--- a/content/source/docs/enterprise/private/monitoring.html.md
+++ b/content/source/docs/enterprise/private/monitoring.html.md
@@ -28,6 +28,11 @@ Beginning in version 201811-1, PTFE includes internal monitoring of critical app
 Error starting userland proxy: listen udp 0.0.0.0:8125: bind: address already in use
 ```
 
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/private/monitoring.html.md
```diff
@@ -18,4 +18,16 @@ In addition to health-check monitoring, we recommend monitoring standard server
 
 - I/O
 - CPU
-- Disk
\ No newline at end of file
+- Disk
+
+## Internal Monitoring
+
+Beginning in version 201811-1, PTFE includes internal monitoring of critical application metrics using a statsd collector on port 8125. The resulting metrics are included with the [diagnostic bundles provided to HashiCorp Support](./diagnostics.html). If the PTFE instance is already running a collector on this port, PTFE may not start up correctly due to the conflict, and logs will indicate:
```
nfagerlund: Question! Why would the instance already be running a collector on that port? Would it be from a zombie copy of the normal collector, or something else?

armchairlinguist: Some of our customers run their own stats collectors!

nfagerlund: Oh huh, yeah, I guess the ptfe instance can run other software without having to cram it into our docker images. I always forget that. 😅

PR #559, opened 2018-11-15
Successful POST to `run/` returns 201 not 200

vwordsworth:
If merged, will fix documentation for the `Create run` endpoint success response status.

image


nfagerlund:
Good catch, thank you!

PR #565, opened 2018-11-20
Fix plugins path

nfagerlund:
This will have a companion PR in hashicorp/terraform, because there are affected pages in there.


nfagerlund:
Martin approved the exact same content in the companion PR, so I'm going to stretch that approval a bit to cover this one.

PR #566, opened 2018-11-26
Add note about 23000-23100 port range

evanphx:
We're starting to put things in this port range so we should reserve it and start to standardize on it more.

### DIFF
--- a/content/source/docs/enterprise/private/faq.html.md
+++ b/content/source/docs/enterprise/private/faq.html.md
@@ -167,7 +167,7 @@ Additionally these hostnames are accessed by the Installer product in online mod
 
 ## Inbound Access
 
-* Access to the product is done on port 443
+* To access Private TFE, ports 80 and 443 must be accessible. Port 80 will redirect to port 443.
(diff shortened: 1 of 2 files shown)


content/source/docs/enterprise/private/faq.html.md
```diff
@@ -183,6 +183,9 @@ from the ELB is routed. Many other internal Private Terraform Enterprise service
 but they do not require external traffic. The AWS security group for the
 instance as well as software firewall rules within the runtime enforce this.
 
+Additionally, the port ranges **9870-9880** and **23000-23100** must be accessible to the host and
+the subnet it is on. These ports are used for internal communication between internal services.
+
```
armchairlinguist: This has been added in the AMI section, but I think we want it up in the general section, if anywhere, as this change won't impact the AMI, right?

evanphx: Oh ack. yes. Thank you.

maryelizbeth: 👍

PR #570, opened 2018-11-27
Document the oauth client update endpoint

zhssh:
screenshot

### DIFF
--- a/content/source/docs/enterprise/api/oauth-clients.html.md
+++ b/content/source/docs/enterprise/api/oauth-clients.html.md
@@ -111,6 +111,100 @@ curl \
 }
 ```
 
+## Update an OAuth Client
+
+`PATCH /oauth-clients/:id`
+
+Parameter            | Description
+---------------------|------------
+`:id`                | The ID of the OAuth Client to update.
+
+Result  | Status and response
+--------|------------------------
+Success | HTTP 200 and a JSON API document (`type: "oauth-clients"`).
+Failure | HTTP 4XX and a JSON API list of error messages.
+
+### Request Body
+
+This PATCH endpoint requires a JSON object with the following properties as a request payload.
+
+Key path                             | Type   | Default | Description
+-------------------------------------|--------|---------|------------
+`data.type`                          | string |         | Must be `"oauth-clients"`.
+`data.attributes.service-provider`   | string |         | The VCS provider being connected with. Valid options are `"github"`, `"github_enterprise"`, `"bitbucket_hosted"`, `"gitlab_hosted"`, `"gitlab_community_edition"`, or `"gitlab_enterprise_edition"`.
+`data.attributes.http-url`           | string |         | The homepage of your VCS provider (e.g. `"https://github.com"` or `"https://ghe.example.com"`)
+`data.attributes.api-url`            | string |         | The base URL of your VCS provider's API (e.g. `https://api.github.com` or `"https://ghe.example.com/api/v3"`)
+`data.attributes.key`                | string |         | The OAuth client key.
+`data.attributes.secret`             | string |         | The OAuth client secret.
+
+### Sample Payload
+
+```json
+{
+  "data": {
+    "id": "oc-XKFwG6ggfA9n7t1K",
+    "type": "oauth-clients",
+    "attributes": {
+      "service-provider": "github",
+      "http-url": "https://github.com",
+      "api-url": "https://api.github.com",
+      "key": "key"
+      "secret": "secret"
+    }
+  }
+}
+```
+
+### Sample Request
+
+```shell
+curl \
+  --header "Authorization: Bearer $TOKEN" \
+  --header "Content-Type: application/vnd.api+json" \
+  --request PATCH \
+  --data @payload.json \
+  https://app.terraform.io/api/v2/oauth-clients/oc-XKFwG6ggfA9n7t1K
+```
+
+### Sample Response
+
+```json
+{
+  "data": {
+    "id": "oc-XKFwG6ggfA9n7t1K",
+    "type": "oauth-clients",
+    "attributes": {
+      "created-at": "2018-04-16T20:42:53.771Z",
+      "callback-url": "https://app.terraform.io/auth/35936d44-842c-4ddd-b4d4-7c741383dc3a/callback",
+      "connect-path": "/auth/35936d44-842c-4ddd-b4d4-7c741383dc3a?organization_id=1",
+      "service-provider": "github",
+      "service-provider-display-name": "GitHub",
+      "http-url": "https://github.com",
+      "api-url": "https://api.github.com",
+      "key": null,
+      "rsa-public-key": null
+    },
+    "relationships": {
+      "organization": {
+        "data": {
+          "id": "my-organization",
+          "type": "organizations"
+        },
+        "links": {
+          "related": "/api/v2/organizations/my-organization"
+        }
+      },
+      "oauth-tokens": {
+        "data": [],
+        "links": {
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/api/oauth-clients.html.md
```diff
@@ -111,6 +111,100 @@ curl \
...
+`PATCH /oauth-clients/:id`
+
+Parameter            | Description
+---------------------|------------
+`:id`                | The ID of the OAuth Client to update.
+
+Result  | Status and response
+--------|------------------------
+Success | HTTP 200 and a JSON API document (`type: "oauth-clients"`).
+Failure | HTTP 4XX and a JSON API list of error messages.
```
armchairlinguist: To be consistent, we should list all the specific 400s (or at least all the common ones).

content/source/docs/enterprise/api/oauth-clients.html.md
```diff
@@ -111,6 +111,100 @@ curl \
...
+```json
+{
+  "data": {
+    "id": "oc-XKFwG6ggfA9n7t1K",
+    "type": "oauth-clients",
+    "attributes": {
+      "service-provider": "github",
+      "http-url": "https://github.com",
+      "api-url": "https://api.github.com",
+      "key": "key"
```
armchairlinguist:
```suggestion
      "key": "key",
```

content/source/docs/enterprise/api/oauth-clients.html.md
```diff
@@ -111,6 +111,100 @@ curl \
...
 ```
 
+## Update an OAuth Client
+
+`PATCH /oauth-clients/:id`
+
+`:id`                | The ID of the OAuth Client to update.
+
```
nfagerlund:
```suggestion

When a VCS service provider changes their API URL, use this endpoint to update the value of `data.attributes.api-url`. Use caution when changing other attributes with this endpoint; editing an OAuth client that workspaces are currently using can have unexpected effects.

```

...or something like that? I feel like I remember from our Slack conversation that there were some possible hazards when pulling the tablecloth out from under the dishes like that.

content/source/docs/enterprise/api/oauth-clients.html.md
```diff
@@ -111,6 +111,100 @@ curl \
...
+This PATCH endpoint requires a JSON object with the following properties as a request payload.
+
+Key path                             | Type   | Default | Description
+-------------------------------------|--------|---------|------------
+`data.type`                          | string |         | Must be `"oauth-clients"`.
+`data.attributes.service-provider`   | string |         | The VCS provider being connected with. Valid options are `"github"`, `"github_enterprise"`, `"bitbucket_hosted"`, `"gitlab_hosted"`, `"gitlab_community_edition"`, or `"gitlab_enterprise_edition"`.
+`data.attributes.http-url`           | string |         | The homepage of your VCS provider (e.g. `"https://github.com"` or `"https://ghe.example.com"`)
+`data.attributes.api-url`            | string |         | The base URL of your VCS provider's API (e.g. `https://api.github.com` or `"https://ghe.example.com/api/v3"`)
+`data.attributes.key`                | string |         | The OAuth client key.
+`data.attributes.secret`             | string |         | The OAuth client secret.
```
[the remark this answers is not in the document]
zhssh: The attributes can be omitted. I've updated the doc :)

content/source/docs/enterprise/api/oauth-clients.html.md
zhssh: The Status/Response section has been replaced on all endpoints in this doc.

content/source/docs/enterprise/api/oauth-clients.html.md
```diff
@@ -26,10 +26,12 @@ This endpoint allows you to create a VCS connection between an organization and
...
 ~> **Note:** This endpoint does not currently support creation of a Bitbucket Server OAuth Client.
 
-Result  | Status and response
---------|------------------------
-Success | HTTP 201 and a JSON API document (`type: "oauth-clients"`).
-Failure | HTTP 4XX and a JSON API list of error messages.
+Status  | Response                                | Reason
+--------|-----------------------------------------|----------
+[201][] | [JSON API document][] (`type: "oauth-clients"`) | The request was successful
+[404][] | [JSON API error object][]               | Organization not found or user unauthorized to perform action
```
armchairlinguist: I think this should probably say "entity not found" or "OAuth client not found"

content/source/docs/enterprise/api/oauth-clients.html.md
```diff
@@ -26,10 +26,12 @@ This endpoint allows you to create a VCS connection between an organization and
...
+[422][] | [JSON API error object][]               | Malformed request body (missing attributes, wrong types, etc.)
+[500][] | [JSON API error object][]               | Failure during OAuth Client creation
```
armchairlinguist: Hm...we typically don't document 500s as return codes, because we don't want 500s to ever be user-facing ideally (although that's aspirational :). What I was suggesting here was that if different codes represent distinguishable issues in the creation (which is the case for e.g. the Admin SMTP settings) then those should all be listed. Otherwise, if all the cases where it would fail due to user error are covered by 422, then we can just omit 500 and stick with 404 and 422.

content/source/docs/enterprise/api/oauth-clients.html.md
```diff
@@ -111,6 +113,104 @@ curl \
...
 
+## Update an OAuth Client
+
+`PATCH /oauth-clients/:id`
+
+Parameter            | Description
+---------------------|------------
+`:id`                | The ID of the OAuth Client to update.
+
+When a VCS service provider changes their API URL, use this endpoint to update the value of `data.attributes.api-url`. Use caution when changing other attributes with this endpoint; editing an OAuth client that workspaces are currently using can have unexpected effects.
```
armchairlinguist: This is perfect!

content/source/docs/enterprise/api/oauth-clients.html.md
```diff
@@ -111,6 +113,104 @@ curl \
...
+
+Status  | Response                                | Reason
+--------|-----------------------------------------|----------
+[200][] | [JSON API document][] (`type: "oauth-clients"`) | The request was successful
+[404][] | [JSON API error object][]               | OAuth Client not found or user unauthorized to perform action
+[422][] | [JSON API error object][]               | Malformed request body (missing attributes, wrong types, etc.)
+[500][] | [JSON API error object][]               | Failure during OAuth Client creation
```
armchairlinguist: Same note here as above.

content/source/docs/enterprise/api/oauth-clients.html.md
nfagerlund: Oh, hmm, I was assuming it was specific to the org since the org is in the URL path... but you're probably right that it 404s for any not found.

zhssh: It was "Organization" because of the URL: `POST /organizations/:organization_name/oauth-clients`. Should we change it to "entity"?

content/source/docs/enterprise/api/oauth-clients.html.md
zhssh: :+1: 404 and 422 should cover all missing or invalid attributes.

The 500 error has been removed. It was copied from the template (`content/source/docs/enterprise/api/_template.md`), so we may want to update that document.

content/source/docs/enterprise/api/oauth-clients.html.md
armchairlinguist: Oh, this is for creation specifically - no, you're right, I'm confused. :)

PR #572, opened 2018-11-27
Adds note about disk performance choices

petems:
* This has come up with a few customers, where 
even though the disk size is fine, performance has
caused issues with the PTFE’s every day workings


roooms:
There are some typos in this PR so it shouldn't be merged as-is. I have incorporated the changes into a new PR that also updates the GCP doc for the new PTFE deployment model.

PR #581, opened 2018-11-30
Add note for replicatedctl location

petems:
* it’s happened enough times that we might
as well put it in the docs :dancer:


armchairlinguist:
This doesn't seem like the appropriate location to put this information really, since it applies to a number of commands we recommend. Perhaps the FAQ page would be better?

petems:
@armchairlinguist Any thoughts about having it on both pages now? I think an FAQ makes more structural sense, but in terms of debugging I think having it as close as possible to the times we're referencing running the commands makes sense in terms of maximum impact 😄

petems:
Made changes as requested, merging 👍

PR #584, opened 2018-11-30
Update some locking response reasons

beekus:
So many locking updates...

[This TFE PR](https://github.com/hashicorp/atlas/pull/7127) makes it so you can't call lock if the workspace is already locked and similarly with unlock/force-unlock.


nfagerlund:
Thanks so much for keeping these in sync!

PR #585, opened 2018-11-30
Delete any redirects from `/enterprise/*` to `/enterprise-legacy/*`

nfagerlund:
These redirects were put in to ease the transition for users of the legacy
version of TFE/Atlas. Now there aren't any of those users left.

We probably shouldn't delete the docs quite yet, but the redirects are starting
to interfere with adding pages to the modern TFE docs, so they should go.


nfagerlund:
Context: @zhssh was working on a PR and got attacked by one of these redirects.

PR #586, opened 2018-12-04
Add documentation for scrubbing upcoming PTFE support bundle commands

bnferguson:
[Opened the PR in the wrong place, moved to the right place, so I'm closing this.]


bnferguson:
[supportbundle.tar.gz](https://github.com/hashicorp/terraform-website/files/2649558/supportbundle.tar.gz)

Here's a copy of a support bundle with all of the current scripts, etc in it as outlined in this documentation. Just want to get it in front of ya'll for review (before https://github.com/hashicorp/ptfe-replicated/pull/185 hits) if ya'll get a chance! /cc @armchairlinguist @maryelizbeth @koikonom

PR #587, opened 2018-12-05
Adds note about validating JSON

petems:
Quick one with validation example

### DIFF
--- a/content/source/docs/enterprise/private/automating-the-installer.html.md
+++ b/content/source/docs/enterprise/private/automating-the-installer.html.md
@@ -41,6 +41,32 @@ The settings file is JSON formatted. All values must be strings.  The example be
 }
 ```
 
+> Note: The JSON file must be valid JSON for the install to work, so it's best to validate it before using for an install.
+
+The easiest way to check the application config is valid JSON would be with `python`, which will be present on most Linux installs:
+
+```
+$ python -m json.tool settings.json
+Expecting property name enclosed in double quotes: line 8 column 5 (char 171)
+```
+
+After fixing the JSON file, the command will return the valid JSON:
+
+```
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/private/automating-the-installer.html.md
```diff
@@ -41,6 +41,15 @@ The settings file is JSON formatted. All values must be strings.  The example be
 }
 ```
 
+> Note: The JSON file must be valid JSON for the install to work, so it's best to validate it before using for an install.
+
+The easiest way to check the application config is valid JSON would be with `python`, which will be present on most Linux installs:
+
+```
+$ python3 -m json.tool settings.json
```
evanphx: Probably just go with `python` here.

content/source/docs/enterprise/private/automating-the-installer.html.md
```diff
@@ -41,6 +41,15 @@ The settings file is JSON formatted. All values must be strings.  The example be
...
+Expecting property name enclosed in double quotes: line 8 column 5 (char 171)
```
evanphx: The example should show the output when it works probably not when there is an error. If you want to show an error output, show both.

content/source/docs/enterprise/private/automating-the-installer.html.md
petems: Fixed 👍

content/source/docs/enterprise/private/automating-the-installer.html.md
petems: Fixed 👍

petems:
@evanphx Can I get a 👍 now I've made those changes?

Issue #592, opened 2018-12-08
Add instructions for changing PATH in Downloads page


jdbohrman:
I feel like adding the link showing developers how to edit the PATH variable to the downloads page would be a very minor change but would also save some people a lot of trouble if they didn't see it beforehand. 

I was thinking something like: 

```
<p>
        Below are the available downloads for the latest version of Terraform
        (<%= latest_version %>). Please download the proper package for your
        operating system and architecture and <a href="https://stackoverflow.com/questions/1618280/where-can-i-set-path-to-make-exe-on-window">add the exracted binary to you PATH </a>
      </p>
```

PR #593, opened 2018-12-08
Update downloads.html.erb

jdbohrman:
Per #592


nfagerlund:
@jdbohrman Hi, and thanks for the suggestion!

The link this PR uses is a little too indirect and too Windows-centric, and the suggestion doesn't quite match the usual install method (most people move the binary to somewhere that's already in the PATH). But it's a good idea, and I think I've found a more general-purpose explanation of the PATH over at https://superuser.com/questions/284342/what-are-path-and-other-environment-variables-and-how-can-i-set-or-use-them. I've filed PR #709 to bring that in.

PR #594, opened 2018-12-10
Adds a general troubleshooting page

petems:
* Common issues I’ve seen occuring

Still a WIP: Need to link it from the sidebar, and add a link from the pre-installer page.

But these are networking gotchas that have come up several times, so it'd be good to catch these earlier and ease the burden on support 😄


armchairlinguist:
Yes - general troubleshooting is more appropriate in the wiki or support KB. Please feel free to chat with me if you aren't sure how to add these, or if they already exist somewhere. (I believe some do.)

petems:
After some discussion we've decided to split the difference. I think the `/etc/docker/daemon.json` is a good idea for pre-flight, the rest we can move into a potential blogpost on troubleshooting and KB articles! 🤝

PR #599, opened 2018-12-13
Add the new queue-all-runs attribute and order examples

svanharmelen:


beekus:
This looks good, but I always like to get reviews from Nick and Alexis as well.

content/source/docs/enterprise/api/workspaces.html.md
```diff
@@ -44,8 +44,9 @@ Key path                                      | Type    | Default   | Descriptio
 `data.type`                                   | string  |           | Must be `"workspaces"`.
 `data.attributes.name`                        | string  |           | The name of the workspace, which can only include letters, numbers, `-`, and `_`. This will be used as an identifier and must be unique in the organization.
 `data.attributes.auto-apply`                  | boolean | false     | Whether to automatically apply changes when a Terraform plan is successful.
-`data.attributes.terraform-version`           | string  | (nothing) | The version of Terraform to use for this workspace. Upon creating a workspace, the latest version is selected unless otherwise specified (e.g. `"0.11.1"`).
 `data.attributes.migration-environment`       | string  | (nothing) | The legacy TFE environment to use as the source of the migration, in the form `organization/environment`. Omit this unless you are migrating a legacy environment.
+`data.attributes.queue-all-runs`              | boolean | false     | Whether all runs should be queued. When set to false, runs triggered by a VCS change will not be queued until at least one run is manually queued.
```
armchairlinguist:
```suggestion
`data.attributes.queue-all-runs`              | boolean | false     | Whether runs should be queued immediately after workspace creation. When set to false, runs triggered by a VCS change will not be queued until at least one run is manually queued.
```

content/source/docs/enterprise/api/workspaces.html.md
```diff
@@ -44,8 +44,9 @@ Key path                                      | Type    | Default   | Descriptio
...
+`data.attributes.terraform-version`           | string  | (nothing) | The version of Terraform to use for this workspace. Upon creating a workspace, the latest version is selected unless otherwise specified (e.g. `"0.11.1"`).
```
armchairlinguist: Is there a need to move this line?

content/source/docs/enterprise/api/workspaces.html.md
```diff
@@ -297,6 +301,7 @@ Key path                                      | Type           | Default
 `data.type`                                   | string         |                  | Must be `"workspaces"`.
 `data.attributes.name`                        | string         | (previous value) | A new name for the workspace, which can only include letters, numbers, `-`, and `_`. This will be used as an identifier and must be unique in the organization. **Warning:** Changing a workspace's name changes its URL in the API and UI.
 `data.attributes.auto-apply`                  | boolean        | (previous value) | Whether to automatically apply changes when a Terraform plan is successful.
+`data.attributes.queue-all-runs`              | boolean        | (previous value) | Whether all runs should be queued. When set to false, runs triggered by a VCS change will not be queued until at least one run is manually queued.
```
armchairlinguist:
```suggestion
`data.attributes.queue-all-runs`              | boolean        | (previous value) | Whether runs should be queued immediately after workspace creation. When set to false, runs triggered by a VCS change will not be queued until at least one run is manually queued.
```

content/source/docs/enterprise/api/workspaces.html.md
svanharmelen: All the reordering is only done to make it alphabetical which is inline with how the API returns these fields. Functionally nothing changed and the order is not in any specific order for any specific reason other then being readable (the API will accepts whatever order).

armchairlinguist: I don't think we really do this in general? And it seems a bit weird to try since vcs-repo would precede working-directory, but it needs to go with its subattributes instead.

I'll let Nick comment, but in any case it doesn't matter much either way.

svanharmelen: @nfagerlund it seems your voice is needed to get this merged or not. I'm fine with swapping these lines again if that is now blocking this PR.

nfagerlund: @svanharmelen Making that alphabetical wasn't necessary, but it's also fine.

IMO the best order for keys of the request body goes something like this: 

- Ideally in order of importance. (Open to interpretation, but IMO type and name should always go first.) 
- In some cases, the _complexity_ of a value means it should go later, so a reader can clear the simple keys out of their head before attempting to understand the complex one. This is the case for the VCS object and its subkeys here — putting that last makes the table easier to understand.  
- Beyond that I don't really care, but I think it's fine to put keys of _equal relative importance and complexity_ in alphabetical order. In cases where there are a LOT of keys, that would make the table easier to use.

svanharmelen: That's useful to know, thanks! Will take that into account the next time I make doc updates 👍

armchairlinguist:
Yep with the changes in wording this is fine. I had a hard time telling if the JSON was no-op, but if that's the general impression then 👍

PR #601, opened 2018-12-13
Update docs for state uploads with locked workspaces

beekus:
State uploads via the v2 api now require that the workspace be locked by the user requesting state upload.

See: https://github.com/hashicorp/atlas/pull/7155

### DIFF
--- a/content/source/docs/enterprise/api/state-versions.html.md
+++ b/content/source/docs/enterprise/api/state-versions.html.md
@@ -16,9 +16,7 @@ sidebar_current: "docs-enterprise2-api-state-versions"
 | --------------- | --------------------------------------------------------- |
 | `:workspace_id` | The workspace ID to create the new state version in. Obtain this from the [workspace settings](../workspaces/settings.html) or the [Show Workspace](./workspaces.html#show-workspace) endpoint. |
 
-Creates a state version and sets it as the current state version for the given
-workspace. This is most useful for migrating existing state from open source
-Terraform into a new TFE workspace.
+Creates a state version and sets it as the current state version for the given workspace. The workspace must be locked by the user creating a state version. This is most useful for migrating existing state from open source Terraform into a new TFE workspace.
 
 !> **Warning:** Use caution when uploading state to workspaces that have already performed Terraform runs. Replacing state improperly can result in orphaned or duplicated infrastructure resources.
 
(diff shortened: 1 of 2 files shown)


content/source/docs/enterprise/api/state-versions.html.md
```diff
@@ -16,9 +16,7 @@ sidebar_current: "docs-enterprise2-api-state-versions"
 | --------------- | --------------------------------------------------------- |
 | `:workspace_id` | The workspace ID to create the new state version in. Obtain this from the [workspace settings](../workspaces/settings.html) or the [Show Workspace](./workspaces.html#show-workspace) endpoint. |
 
-Creates a state version and sets it as the current state version for the given
-workspace. This is most useful for migrating existing state from open source
-Terraform into a new TFE workspace.
+Creates a state version and sets it as the current state version for the given workspace. The workspace must be locked by the user creating a state version. This is most useful for migrating existing state from open source Terraform into a new TFE workspace.
```
armchairlinguist: Can we provide some info here about how to lock the workspace?

content/source/docs/enterprise/workspaces/settings.html.md
```diff
@@ -73,11 +73,13 @@ After changing this setting, you must click the "Update SSH key" button below it
...
 
 Locks are managed with a single "Lock/Unlock/Force unlock `<WORKSPACE NAME>`" button. TFE asks for confirmation when unlocking.
 
-~> **Important:** Locking a workspace prevents runs within TFE, but it **does not** prevent state from being updated. This means a user with write access can still modify the workspace's resources by running Terraform outside TFE with [the `atlas` remote backend](/docs/backends/types/terraform-enterprise.html). To prevent confusion and accidents, avoid using the `atlas` backend in normal workflows; to perform runs from the command line, see [TFE's CLI-driven workflow](../run/cli.html).
+~> **Important:** Locking a workspace prevents runs within TFE, but it **does not** prevent state from being updated when using the atlas backend. This means a user with write access can still modify the workspace's resources by running Terraform outside TFE with [the `atlas` remote backend](/docs/backends/types/terraform-enterprise.html). To prevent confusion and accidents, avoid using the `atlas` backend in normal workflows; to perform runs from the command line, see [TFE's CLI-driven workflow](../run/cli.html).
```
armchairlinguist: This seems like a duplicate of the comment above, so I'd suggest we remove the comment above or explicitly refer to this note.

beekus: agree. I think this comment has a lot more in it than the thing i added above, so maybe we extract bits of this comment to the paragraph I added. I do like that this is called out in an "important" box though.

content/source/docs/enterprise/api/state-versions.html.md
beekus: done

content/source/docs/enterprise/workspaces/settings.html.md
beekus: @nfagerlund what do you recommend here? Happy to jump on a zoom to explain the subtleties here.

PR #598, opened 2018-12-12
GitHub Actions: Update docker image and args for filter action

joakimhellum-in:
Fixes #597

PR #603, opened 2018-12-14
Switch to recommended filter action.

lkysow:
The superbbears/filter action was created before GitHub actions launched
and that Docker image is no longer published. Also change the arguments
to add single quotes as that is now required.


lkysow:
Closed by #598 which was created first (but I didn't see it and created this pull request instead).

PR #605, opened 2018-12-17
Add entitlement set docs

beekus:
Adds entitlement set information to the docs according to RFC TFE-096. I expect the descriptions in the first screenshot to need some help.

entitlements

Also, adds the entitlement-set endpoint to the organization api details page:
entitlement api


content/source/docs/enterprise/api/organizations.html.md
```diff
@@ -339,3 +339,52 @@ curl \
 ### Sample Response
 
 The response body will be empty if successful.
+
+## Show the Entitlement Set
+This endpoint shows the [entitlements](../users-teams-organizations/entitlements.html) for an organization.
```
armchairlinguist: Can these also just be included with the org? We usually have an Available Related Resources that points to a separate page in those cases, so it might be better to do this one that way?

content/source/docs/enterprise/users-teams-organizations/entitlements.html.md
```diff
@@ -0,0 +1,40 @@
+---
+layout: enterprise2
+page_title: "entitlements - Terraform Enterprise"
+sidebar_current: "docs-enterprise2-users-teams-organizations-entitlements"
+---
+
+# Entitlements
```
armchairlinguist: I'm not convinced that it makes sense for this page to exist, since there's no non-API way for orgs to "see" their entitlements. If I were going to write a page like this, I'd probably make it more "friendly" and describe a brief list of the entitlements and link to the API and the marketing plans page.

nfagerlund:
I have thoughts. I don't think there should be a separate page about entitlements! 

Are we planning to expose entitlements somehow in the organization settings UI? I think we probably should, _unless_ we believe that the entitlement situation is self-evident from the state of the app or from the user's memory of which paid plan their organization is on. My preference would be to have a badge on the org's profile page (https://app.terraform.io/app/<ORG>/settings/profile) that shows their billing plan, with a short list of the available features from that plan below. 

- If we decide to do that, we can document the entitlements as a bulleted list at the appropriate spot in the `users-teams-organizations/organizations.html` page.
- If not, then the specific entitlements are really only relevant to API clients. We should make them more useful for API clients by doing the following:
    1. List the possible entitlements inline under the entitlement set endpoint.
    2. Add a brief note to `api/index.html` stating that access to certain endpoints is limited by an organization's entitlements, and that if you're writing a general-purpose API client, it should check the entitlements to determine what its capabilities are for a given org. (And link to the entitlement set endpoint.) 
    3. Add brief comments to any endpoint whose access is limited by entitlements. (uhhhhhh which is _most_ of them, I think... so maybe we'd need to rethink the approach there. We could reverse that instead, and list the affected endpoints for each entitlement.)

y/n?

beekus:
Thanks for the feedback y'all.

While I don't think entitlements will be a part of the UI right now, I do expect us to expose that information at some point when we get deeper into plan levels. I originally intended this PR to just give us the places to add more information as we grow the entitlements. 

Right now, this is only useful for folks writing API clients, but it should be useful for users of the app within the next 6 months or so. What is the best way to organize this information given all that? Or should we just not document it until it should be present in the UI?

nfagerlund:
@beekus I just pushed a commit with my suggestions for how to arrange this today! It's got a couple of holes, which I figure you have the knowledge to fill in. 

What do you think of this approach? It keeps the language about "entitlements" tied to the spot where it's useful today, and we can shuffle things around in six months if needed.

content/source/docs/enterprise/api/index.html.md
```diff
@@ -28,6 +28,47 @@ There are three kinds of token available:
...
 
+## Feature Entitlements
+
+TFE is available at multiple pricing tiers (including free), which offer different subsets of TFE's features.
+
+Each organization has a set of _entitlements_ that corresponds to its pricing tier. These entitlements determine which TFE features the organization can use.
+
+If an organization doesn't have the necessary entitlement to use a given feature:
+
+- TFE rejects API requests to any endpoints devoted to that feature, returning a (SPECIFIC ERROR CODE GOES HERE) error.
```
beekus:
```suggestion
- TFE rejects API requests to any endpoints devoted to that feature, returning a *404 NOT FOUND*  error.
```

content/source/docs/enterprise/api/index.html.md
```diff
@@ -28,6 +28,47 @@ There are three kinds of token available:
...
+- (Anything else? Do some endpoints change their response, e.g. do workspace objects have some properties missing if the org can't perform runs?)
```
beekus: Nope! JSON response objects remain the same for now (I don't think we will change that, but I could be wrong).

beekus:
I like it.

PR #606, opened 2018-12-17
Sentinel mgmt: remote backend doesn't support import yet

nfagerlund:
I learned this during a demo. 😑 The atlas backend works fine tho.


nfagerlund:
@svanharmelen I think you're the one I should ask for review on this — it's just as a temp patch until 0.12 is out and/or we have that 0.11-with-improved-backend release.

nfagerlund:
@svanharmelen Ah, that makes sense. Sounds good to me!

PR #616, opened 2018-12-21
API template: Add tips on sorting properties for the request body

nfagerlund:
This came up in another PR, so I'm taking a stab at recording this for the next
person.


armchairlinguist:
This makes sense to me!

PR #617, opened 2018-12-21
Improve section-to-section navigation via sidebar navs

nfagerlund:
This is a companion PR to https://github.com/hashicorp/terraform/pull/19736, which has the rationale etc. 

For follow-up after the new year!


nfagerlund:
Approvals are in https://github.com/hashicorp/terraform/pull/19736.

PR #618, opened 2018-12-21
Add `terraform-init` action to credential configuration

aharden:
My initial tests of the TF GitHub Actions against AWS required the credentials to be present in both the `terraform-init` and `terraform-plan` actions. Also removed "only" from the statement about adding Secrets because they can be added in the repository's settings.


lkysow:
Build is failing for unrelated reasons so I'm going to merge. Thanks for this Alex.

PR #633, opened 2019-01-09
Add CE to note about local network hooks

armchairlinguist:
per customer feedback


armchairlinguist:
@nfagerlund yep - this was reported recently by a customer to affect CE as well.

PR #634, opened 2019-01-09
Addressing apparent failure to configure

armchairlinguist:
Adds quick troubleshooting for most common issues when install succeeds but isn't configured.

### DIFF
--- a/content/source/docs/enterprise/private/automating-the-installer.html.md
+++ b/content/source/docs/enterprise/private/automating-the-installer.html.md
@@ -294,9 +294,9 @@ done
 
 ## If the installation does not appear to be configured correctly
 
-If the installation is successful, but prompts for additional configuration rather than skipping that step, it's likely that the supplied configuration files were not applied during installation.
+If the installation script (`install.sh`) exits successfully, but the Replicated web UI prompts for additional configuration rather than skipping that step, it's likely that the supplied configuration files were not applied during installation.
 
-* Verify the locations of the files: `/etc/replicated.conf` for the Replicated configuration file, and the path specified in the Replicated configuration file for the application settings. 
+* Verify the locations and permissions of the files: `/etc/replicated.conf` for the Replicated configuration file, and the path specified in the Replicated configuration file for the application settings. 
 * Run a JSON validator on the files to check their validity.
 
 ## References
--- a/content/source/docs/enterprise/private/automating-the-installer.html.md
+++ b/content/source/docs/enterprise/private/automating-the-installer.html.md
@@ -292,6 +292,13 @@ while ! curl -ksfS --connect-timeout 5 https://tfe.example.com/_health_check; do
 done
 ```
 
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/private/automating-the-installer.html.md
```diff
@@ -292,6 +292,13 @@ while ! curl -ksfS --connect-timeout 5 https://tfe.example.com/_health_check; do
 done
 ```
 
+## If the installation does not appear to be configured correctly
+
+If the installation is successful, but prompts for additional configuration rather than skipping that step, it's likely that the supplied configuration files were not applied during installation.
```
koikonom:
```suggestion
If the installation script (`install.sh`) exits successfully, but the Replicated web UI prompts for additional configuration rather than skipping that step, it's likely that the supplied configuration files were not applied during installation.
```

content/source/docs/enterprise/private/automating-the-installer.html.md
```diff
@@ -292,6 +292,13 @@ while ! curl -ksfS --connect-timeout 5 https://tfe.example.com/_health_check; do
...
+
+* Verify the locations of the files: `/etc/replicated.conf` for the Replicated configuration file, and the path specified in the Replicated configuration file for the application settings.
```
koikonom:
```suggestion
* Verify the locations and permissions of the files: `/etc/replicated.conf` for the Replicated configuration file, and the path specified in the Replicated configuration file for the application settings. 
```

armchairlinguist: Great suggestion - what are the correct permissions? 600? 400? 644?

koikonom: 644 will work fine. If we want tighten permissions for security puproses then it's 600 and owned by the replicated user

content/source/docs/enterprise/private/automating-the-installer.html.md
```diff
@@ -292,6 +292,13 @@ while ! curl -ksfS --connect-timeout 5 https://tfe.example.com/_health_check; do
 done
 ```
 
+## If the installation does not appear to be configured correctly
+
+If the installation script (`install.sh`) exits successfully, but the Replicated web UI prompts for additional configuration rather than skipping that step, it's likely that the supplied configuration files were not applied during installation.
+
+* Verify the locations and permissions of the files: `/etc/replicated.conf` for the Replicated configuration file, and the path specified in the Replicated configuration file for the application settings. 
+* Run a JSON validator on the files to check their validity.
```
nfagerlund: After fixing what was wrong with the files, do you run the install script again? Or just go to the web UI?

armchairlinguist: Ah yeah, that's a good call as something to add. To verify that they work, you'd need to re-run it. I'll update.

PR #640, opened 2019-01-10
advise users that GitLab subgroups are not supported in the registry yet

armchairlinguist:
Adds a note to the note and to the requirements.


armchairlinguist:
@nfagerlund yep - I tested workspaces recently and they appear to work fine, I think because there's no required naming convention?

PR #644, opened 2019-01-14
add 'plan' as a valid option for TFE team access to workspace

hagurganus:
Update TFE API doc to include 'plan' access level as a valid option for adding team access. The [TFE Permissions](https://www.terraform.io/docs/enterprise/users-teams-organizations/permissions.html) documentation includes this option, and I've successfully tested an API call to grant a team this permission on a workspace (pTFE v201812-1).


nfagerlund:
Oh, good catch! Thank you.

hagurganus:
You got it—thanks for the merge!

PR #645, opened 2019-01-15
add base software used in PTFE worker containers

armchairlinguist:
This adds the software we install to the sample Dockerfile, and should help people avoid non-starter runs on custom images.

I know from testing that the first four are definitely required for even the most basic runs; if there's some of the others that aren't absolutely necessary, I can remove those.


armchairlinguist:
Failures don't seem to have anything to do with this change.

PR #646, opened 2019-01-16
update automated recovery process for airgap

armchairlinguist:
And some small style updates.


content/source/docs/enterprise/private/automated-recovery.html.md
```diff
@@ -222,3 +222,12 @@ done
 echo
 echo "Application booted!"
 ```
+
+## Airgap recovery considerations
+
+The instructions above are tailored for the online install method. When restoring on an airgap instance, there are several additional considerations:
+
+1. The minimum version of Replicated is 2.31.0, rather than 2.17.0.
+1. The license file and airgap package must be in place on the new instance prior to restore.
```
koikonom: where?

PR #648, opened 2019-01-16
[DO NOT MERGE] Add temporary redirects for 0.12 pages

paultyng:


paultyng:
I don't think this should be merged until we see how difficult it is to delete the redirs per:

> Deleting items is not supported at this time. To delete an item, contact the release engineering team and they will delete the dictionary item.

PR #650, opened 2019-01-18
sentinel: Update mocks to new >= v0.8.x format

vancluever:
This updates the mocks for the three TFE imports to use Sentinel code.

Note that this requires version 0.8.0 or higher of the simulator. For
instructions on how to use them, see:
https://docs.hashicorp.com/sentinel/commands/config#mocking-with-sentinel-code

The runtime docs referenced in the individual pages will be updated
with this information as well.

### DIFF
--- a/content/source/docs/enterprise/sentinel/import/mock-tfstate.html.md
+++ b/content/source/docs/enterprise/sentinel/import/mock-tfstate.html.md
@@ -30,26 +30,26 @@ version of the simulator, see the [downloads page][ref-downloads-page].
 
 [ref-downloads-page]: https://docs.hashicorp.com/sentinel/downloads
 
-These mocks are configured differently than JSON mocks: 
+These mocks are configured differently than JSON mocks:
 
-* Save the contents below to a `.sentinel` file within your project directory.
-   The example below uses `mock-tfstate.sentinel`.
+* Save the mock file contents below to a `.sentinel` file within your project
+  directory. The example below uses `mock-tfstate.sentinel`.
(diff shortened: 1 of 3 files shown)


vancluever:
Note: There is a bit of a wordy section on configuration in each section as I kind of want that to be front and center, at least until we have a chance to update the runtime documentation. Even with that I think having this data immediately available has value as there's no ambiguity on what needs to be done to configure things.

content/source/docs/enterprise/sentinel/import/mock-tfstate.html.md
```diff
@@ -22,127 +22,195 @@ documentation][ref-official-sentinel-documentation].
...
+  ],
+]
+
+terraform_version = "0.11.7"
+
+module = func(path) {
+  if length(path) == 0 {
+    return _root
+  }
+  if length(path) == 1 and path[0] is "foo" {
```
nfagerlund: This seems REAL squirrelly. As a user, I want something more resilient that I can copy-paste... can we give people a general purpose `module()` function for mocks? Like, can we save the module data as a map whose keys are module paths, and just index into that?

vancluever: These mocks are auto-generated so readability might not be 100% optimal.

The function could probably be optimized in the way that you are proposing, but I still would like it pushed for now so that we have at least communicated that mocking all parts of the Terraform imports are now possible. These can be updated at a later time.

PR #656, opened 2019-01-29
Add additional example for tfconfig import

rberlind:
I added an additional example for the config value within the resource namespace.
I thought it would be useful to include an example in which a single top-level attribute of a resource is evaluated by a policy.


rberlind:
Hi @nfagerlund and @vancluever :  Could I go ahead and merge this since @nfagerlund approved it?  Or did @vancluever also want to review first?

nfagerlund:
Go ahead and merge.

rberlind:
Thanks

PR #657, opened 2019-01-31
Auto-apply API - clarification on token requirement

firedot:
User token is required to use auto-apply over API. However this is not cristal clear on the documentation.

We have a few support tickets from people trying auto-apply with organisational tokens, which prevents auto-apply from running; this note might help people that are trying to use auto-apply over API to use the right token.

### DIFF
--- a/content/source/docs/enterprise/run/api.html.md
+++ b/content/source/docs/enterprise/run/api.html.md
@@ -95,7 +95,9 @@ UPLOAD_URL=($(curl \
 
 Next, upload the configuration version `tar.gz` file to the upload URL extracted from the previous step. If a file is not uploaded, the configuration version will not be usable, since it will have no Terraform configuration files.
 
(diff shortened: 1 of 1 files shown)


nfagerlund:
@firedot Hey, thanks for this! 

So, this is actually a bug in TFE, and one of our engineers is currently working on it. Team tokens are supposed to work normally with auto-apply. 

We don't want to merge documentation of a soon-to-be-fixed bug! But if you'd like, we can leave this PR open until the fix is shipped; that way you can get a ping once it's taken care of.

firedot:
@nfagerlund Thank you for the information! Let's keep the PR open until the issue is resolved.

nfagerlund:
@firedot Just checked on this, and it looks like the fix was merged in and deployed to app.terraform.io; it'll roll out to private instances in the next release.

PR #664, opened 2019-02-06
Sentinel VCS workflow: suggest using data source and clarify other options

nfagerlund:
image

### DIFF
--- a/content/source/docs/enterprise/sentinel/integrate-vcs.html.md
+++ b/content/source/docs/enterprise/sentinel/integrate-vcs.html.md
@@ -91,37 +91,6 @@ variable "tfe_organization" {
 }
 ```
 
-### Optional: Define a Workspace IDs Variable
-
-[inpage_ids]: #optional-define-a-workspace-ids-variable
-
-
-The `tfe_policy_set` resource uses workspace IDs, which can be found on a workspace's [settings page](../workspaces/settings.html#id). You can use these IDs directly, but the configuration will be more readable if you provide a map of names to IDs and refer to workspaces by name throughout the configuration. Use a Terraform variable for this map, so you can update it in TFE without changing the configuration:
-
-```hcl
-variable "tfe_workspace_ids" {
-  description = "Mapping of workspace names to IDs, for easier use in policy sets."
-  type        = "map"
-
-  default = {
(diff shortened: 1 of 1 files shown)


nfagerlund:
Shout-out to @svanharmelen for clarifying that security trade-off for me. 😄

content/source/docs/enterprise/sentinel/integrate-vcs.html.md
```diff
@@ -91,49 +91,47 @@ variable "tfe_organization" {
...
-Configure the `tfe` provider (version 0.3 or higher) with your API token and hostname variables.
+In our examples, we want to clearly show what our policy sets are doing, which is much easier if we can refer to workspaces by name. The [`tfe_workspace_ids` data source](/docs/providers/tfe/d/workspace_ids.html) (in `tfe` provider versions ≥ 0.6) makes it easy to use workspace names:
 
 ```hcl
-provider "tfe" {
-  hostname = "${var.tfe_hostname}"
-  token    = "${var.tfe_token}"
-  version  = "~> 0.4"
+data "tfe_workspace_ids" "all" {
+  names        = ["*"]
```
svanharmelen: Instead of using the wildcard in the example here, I suggest to use a list of actual names that are used. Using exact names seems to indicate the expected behavior better then when using a big "catch all" hammer to just get everything by default.

armchairlinguist: The number of names could be really large...is there some intermediate option like partial wildcard?

svanharmelen: No that isn't implemented at the moment.

nfagerlund: Haha, I remember talking about this when you were implementing the data source. 😄 

I'm using the wildcard because I'm _already_ specifying all of the workspace names I want, down in the policy set resources. Specifying them _again,_ up in the data source, seems like a waste of my time that doesn't provide me any benefit. 

Unless it somehow _does_ provide me benefit? But the only thing I can think of is if using the wildcard incurs some kind of huge multi-minute performance penalty once I hit a couple hundred workspaces. Otherwise... you say "big catch-all hammer," I hear "easy way to avoid pointless busywork." 🙂

PR #668, opened 2019-02-09
Bump image versions

paultyng:
These versions probably should not be patches, and maybe we should sync them to the TF version itself?


paultyng:
We should probably source this copy from the GH actions repo so a single commit can get the docker tags and the website. Similar to the other git submodules for TF core and the providers.

PR #669, opened 2019-02-12
Update several pages and screenshots to account for the rearranged settings menu

nfagerlund:
...I _think_ I got most of these. There are a few other changes in here as well,
since I noticed some out-of-date stuff in the Getting Started guide. I didn't go
deep on the remaining out-of-date screenshots in there, because they weren't
_meaningfully_ out of date yet and because I'm still hoping to use Mirage to
partially automate some of that pain.


nfagerlund:
Thanks for catching that!

PR #672, opened 2019-02-13
Clarify the exceptions to auto-apply

nfagerlund:
Pinging David for review bc I saw his prints on the section of code that does this. :]


nfagerlund:
Ugh yes, I was debating that. 

If the definition gets much more complicated, then yes. But for now, I think it's one of those things where we should just suck up a worse update experience in order to give a less annoying reading experience.

PR #673, opened 2019-02-13
Update vault key info

nfagerlund:
I'm pretty sure about the stuff I've got in here now (Thanks @armchairlinguist!), but have two questions about possible updates in other spots:

- In [the last section of the install page](https://terraform.io/docs/enterprise/private/install-installer.html#continue-installation-in-browser), we have this:

    > 1. _Optional:_ Adjust the path used to store the vault files that are used to encrypt
    >    sensitive data. This is a path on the host system, which allows you
    >    to store these files outside of the product to enhance security. Additionally,
    >    you can configure the system not to store the vault files within any snapshots,
    >    giving you full custody of these files. These files will need to be provided before
    >    any snapshot restore process is performed, and should be placed into the path configured.

    Does that need an update too? 
- Do we need to update any info about the restore process? My expectation would be that snapshot restores wouldn't be affected, but that this changes the procedure for rebuilding an instance from scratch (which I don't think we have any specific info about, since it's basically just installation?). Is that guess right, or is there something that needs adjustment?

### DIFF
--- a/content/source/docs/enterprise/private/reliability-availability.html.md
+++ b/content/source/docs/enterprise/private/reliability-availability.html.md
@@ -136,19 +136,21 @@ of the user.
 
 |  | Configuration | Vault | PostgreSQL | Blob Storage |
 |-------------------|--------------------------------------|----------------------------------------------------------------------------------------------|--------------------------------------|--------------------------------------|
-| Demo | Stored in Docker volumes on instance | Key material is stored in Docker volumes on instance, storage backend is internal PostgreSQL | Stored in Docker volumes on instance | Stored in Docker volumes on instance |
-| Mounted Disk | Stored in Docker volumes on instance | Key material on host in `/var/lib/tfe-vault`, storage backend is mounted disk PostgreSQL | Stored in mounted disks | Stored in mounted disks |
-| External Services | Stored in Docker volumes on instance | Key material on host in `/var/lib/tfe-vault`, storage backend is external PostgreSQL | Stored in external service | Stored in external service |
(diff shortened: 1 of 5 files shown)


content/source/docs/enterprise/private/reliability-availability.html.md
```diff
@@ -136,19 +136,21 @@ of the user.
...
+
 *Backup and Restore Responsibility*
 
-|                   | Configuration | Vault | PostgreSQL | Blob Storage |
-|-------------------|---------------|-------|------------|--------------|
-| Demo              | PTFE          | PTFE  | PTFE       | PTFE         |
-| Mounted Disk      | PTFE          | PTFE  | User       | User         |
-| External Services | PTFE          | PTFE  | User       | User         |
-| External Vault    | -             | User  | -          | -            |
+|                   | Configuration | Vault | Password | PostgreSQL | Blob Storage |
```
nfagerlund: Reasonable to treat the encryption password as its own part of your backup regime? Or not necessary?

armchairlinguist: No, I don't think it needs a separate entry. As is noted in one of the files you edited:

> The encrypted Vault credentials are stored in PostgreSQL, so they are included in normal backups and can be recovered with the password.

armchairlinguist:
@nfagerlund I think you're right that paragraph also needs a language update. I don't think restores/fresh instances change at all, other than needing to supply the encryption password, which I think we already added in the automated install/recovery pages.

I'll add @evanphx as a reviewer since he's the authority here :)

nfagerlund:
This went stale, and I'm going to close it.

armchairlinguist:
😭 Guess I didn't tie up all my loose ends. Miss y'all!

nfagerlund:
@armchairlinguist 💖

PR #680, opened 2019-02-22
Update OAuth Clients API docs with new `name` attribute

davidcelis:
OAuth Clients can now have an optional display name; this updates the API docs with support for that attribute.

### DIFF
--- a/content/source/docs/enterprise/api/oauth-clients.html.md
+++ b/content/source/docs/enterprise/api/oauth-clients.html.md
@@ -49,6 +49,7 @@ curl \
         "connect-path": "/auth/35936d44-842c-4ddd-b4d4-7c741383dc3a?organization_id=1",
         "service-provider": "github",
         "service-provider-display-name": "GitHub",
+        "name": null,
         "http-url": "https://github.com",
         "api-url": "https://api.github.com",
         "key": null,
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/api/oauth-clients.html.md
```diff
@@ -170,6 +172,7 @@ Key path                             | Type   | Default | Description
 -------------------------------------|--------|---------|------------
 `data.type`                          | string |         | Must be `"oauth-clients"`.
 `data.attributes.service-provider`   | string |         | The VCS provider being connected with. Valid options are `"github"`, `"github_enterprise"`, `"bitbucket_hosted"`, `"gitlab_hosted"`, `"gitlab_community_edition"`, or `"gitlab_enterprise_edition"`.
+`data.attributes.name`               | string | `null`  | An optional display name for the OAuth Client. If left `null`, the Terraform Enterprise UI will default to displaying the value of `service-provider-display-name` as the OAuth Client's name.
```
armchairlinguist: Is `service-provider-display-name` defined anywhere in our docs? It might be more straightforward to say "default to the display name of the VCS provider".

davidcelis: Yeah, it's listed as an attribute in OAuth Client responses

armchairlinguist: I'm not sure that anyone reads the whole JSON in these responses (I certainly never have :)). I think it's more straightforward to describe it the way I did, but up to you.

PR #681, opened 2019-02-22
Update null default for owners-team-saml-role-id

thrashr888:
A blank default value means it's required. `null` shows it's optional and empty.

A follow-up for #679


ryanuber:
Have we used this pattern elsewhere? Just wondering if `null` is the right word. This totally makes sense to me, but since it's user documentation and not a languageI wonder if `<none>` makes more sense, or something like that. Seems fine either way, just a thought 🤷‍♂️

thrashr888:
I just dug around and we use `null` or `(nothing)`. I've missed adding "**Optional**" though, so I'll add that and just use `(nothing)` since that seems to make more sense than actually sending `null`.

PR #682, opened 2019-02-22
Remove and note Sentinel data in policy check API

thrashr888:
> The /policy-checks API endpoint returns the JSON object which is returned from Sentinel. As such, if we update Sentinel in TFE and the output of that object changes, it will cause the API response on /policy-check to change as well.

https://app.asana.com/0/390268369299573/493393102787398/f


content/source/docs/enterprise/api/policy-checks.html.md
```diff
@@ -12,6 +12,8 @@ sidebar_current: "docs-enterprise2-api-policy-checks"
 
 This endpoint lists the policy checks in a run.
 
+-> **Note**: The `sentinel` value is passed through directly from Sentinel and is likely to be absent or change.
```
ryanuber: I would word this more like:

> The `sentinel` hash in the `result` attribute structure represents low-level Sentinel details generated by the policy engine. The keys or structure may change over time. Use the data in this hash at your own risk.

PR #688, opened 2019-02-27
API: Add some info about user objects

nfagerlund:


thrashr888:
We should probably also expose the user id for the user to copy, no? Would this help?

terraform_enterprise___account___profile

PR #702, opened 2019-03-11
Document extensions required by PTFE

evanphx:
Customer that want to use a more restrictive user will need to create the postgresql extensions before hand. This documents what those extensions are.

### DIFF
--- a/content/source/docs/enterprise/private/preflight-installer.html.md
+++ b/content/source/docs/enterprise/private/preflight-installer.html.md
@@ -155,14 +155,15 @@ When Terraform Enterprise uses an external PostgreSQL database, the
 following must be present on it:
 
 * PostgreSQL version 9.4 or greater
-* User with all privileges granted on the schemas created and the ability to run "CREATE EXTENSION" on the database
+* User with the ability to create/modify/read tables and indices on all schemas created
+  * If the user can not create extensions (ie is not a superuser), then the extensions listed below must be created first
 * The following PostgreSQL schemas must be installed into the database: `rails`, `vault`, `registry`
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/private/preflight-installer.html.md
```diff
@@ -155,14 +155,15 @@ When Terraform Enterprise uses an external PostgreSQL database, the
 following must be present on it:
 
 * PostgreSQL version 9.4 or greater
-* User with all privileges granted on the schemas created and the ability to run "CREATE EXTENSION" on the database
```
armchairlinguist: I might rather we stick with this wording, and only add the sub-bullet with the slightly stronger wording suggested

content/source/docs/enterprise/private/preflight-installer.html.md
```diff
@@ -155,14 +155,15 @@ When Terraform Enterprise uses an external PostgreSQL database, the
...
+* User with the ability to create/modify/read tables and indices on all schemas created
+  * If the user can not create extensions (ie is not a superuser), then the extensions listed below must be created first
```
armchairlinguist:
```suggestion
  * If it's not feasible to have a user with "CREATE EXTENSION", then create the [extensions](#extensions) below before installation
```

content/source/docs/enterprise/private/preflight-installer.html.md
evanphx: But it's not really true given the other statements. I don't want to contradict this statement just a few lines later.

PR #703, opened 2019-03-12
Update external link list

paultyng:
Updated external link list from Google Search Console.


paultyng:
`ACCEPT_PATH` is blank for websites it looks like this, so this should help exclude those.

PR #709, opened 2019-03-14
Add installation note to Terraform download page

### LINKED ISSUE #592: Add instructions for changing PATH in Downloads page

nfagerlund:
Supersedes #593, resolves #592.

### DIFF
--- a/content/source/downloads.html.erb
+++ b/content/source/downloads.html.erb
@@ -17,6 +17,13 @@ description: |-
         operating system and architecture.
       </p>
       <p>
+        Terraform is distributed as a single binary. Install Terraform by
+        unzipping it and moving it to a directory included in your system's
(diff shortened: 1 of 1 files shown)


nfagerlund:
@armchairlinguist UGH, me too!!! I looked for quite a while trying to find something similarly comprehensive but more authoritative, but came up bupkis — the wikipedia article was kind of unhelpful tbh.

PR #710, opened 2019-03-14
Replace /docs/index.html with a map to all of the docs!

nfagerlund:
> **Related PRs:** This can't be merged until after https://github.com/hashicorp/terraform/pull/20693 is merged and cherry-picked to stable-master. 

During a conversation in the Terraform docs slack channel, it occurred to several of us that the current /docs/index.html page isn't very helpful at all! I think this should improve it. 

Notably, changing this page means we need a dedicated landing page for the Terraform CLI docs now, which is what https://github.com/hashicorp/terraform/pull/20693 is about. That's going to be /docs/cli-index.html. 

image

### DIFF
--- a/content/source/docs/index.html.markdown
+++ b/content/source/docs/index.html.markdown
@@ -15,6 +15,10 @@ There are a lot of things a user of Terraform might need to know about. To help 
 
 You can navigate the docs via this index page, or you can jump from section to section using the "Other Docs" area of the navigation sidebar, available in most areas of this site.
 
+<div class="container-fluid"><div class="row">
+
+<div class="col-md-6 col-sm-12">
+
 ### [Terraform Glossary](/docs/glossary.html)
 
 Definitions (and helpful links) for technical terms used throughout Terraform's documentation, help text, and UI. Visit the glossary whenever you get lost.
@@ -49,6 +53,15 @@ A broad overview of what Terraform is and why people use it.
 
 Interactive guides to teach you how to use Terraform's features. Begin with the [Getting Started guide](https://learn.hashicorp.com/terraform/getting-started/install.html), then continue with task-specific advanced guides or go directly to the [Terraform CLI docs](/docs/cli-index.html).
 
+</div>
+
+
+
+
+
+
+<div class="col-md-6 col-sm-12">
+
 ### [Guides and Whitepapers](/guides/index.html)
 
 -> Intermediate users can go here for a deeper understanding of what's possible with Terraform.
@@ -77,3 +90,6 @@ Documentation about developing Terraform providers, with extensive information a
 
 Terraform relies on provider plugins to manage infrastructure resources across a wide variety of infrastructure services. Anyone can make and distribute a Terraform provider for their own service.
 
+</div>
+
+</div></div>
--- a/content/source/docs/index.html.markdown
+++ b/content/source/docs/index.html.markdown
@@ -1,7 +1,7 @@
 ---
 layout: "docs"
 page_title: "Documentation"
-sidebar_current: "docs-home"
+sidebar_current: "docs-frontpage"
 description: |-
   Welcome to the Terraform documentation! This documentation is more of a reference guide for all available features and options of Terraform. If you're just getting started with Terraform, please start with the introduction and getting started guide instead.
 ---
--- a/content/source/docs/index.html.markdown
+++ b/content/source/docs/index.html.markdown
@@ -1,14 +1,79 @@
 ---
-layout: "docs"
+layout: "docs-frontpage"
 page_title: "Documentation"
 sidebar_current: "docs-frontpage"
 description: |-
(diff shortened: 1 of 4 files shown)


nfagerlund:
nb that if your viewport is too small for the columns to be useful, they just stack like it's a normal page. 

image

content/source/docs/index.html.markdown
```diff
@@ -1,14 +1,95 @@
...
+
+...and much more.
+
+### [Terraform Enterprise](/docs/enterprise/index.html)
+
+Documentation for Terraform Enterprise.
+
+Terraform Enterprise is a supplementary paid product that makes it easier for teams to manage infrastructure together. It provides a way to share access to sensitive data, a consistent and reliable environment for Terraform runs, access controls for approving changes to infrastructure, a private registry for sharing Terraform modules, detailed policy controls for governing the contents of Terraform configurations, and more.
+
+### [Introduction to Terraform](/intro/index.html)
```
apparentlymart: Given that this section of the page is visually relatively small and aimed at less-familiar users, I wonder about putting it at the top of this column instead of the bottom. That way we can redirect new users to where they need to be without detracting _too_ much from the navigation experience for those who already saw it.

content/source/docs/index.html.markdown
```diff
@@ -1,14 +1,95 @@
...
+
+There are a lot of things a user of Terraform might need to know about. To help you manage this information, we've divided the Terraform docs into several sections.
+
+You can navigate the docs via this index page, or you can jump from section to section using the "Other Docs" area of the navigation sidebar, available in most areas of this site.
+
+<div class="container-fluid"><div class="row">
+
+<div class="col-md-6 col-sm-12">
+
+### [Terraform Glossary](/docs/glossary.html)
```
armchairlinguist: I can see a rationale for having this up top, but I think it won't be most people's first stop, and would suggest shifting it down/over.

apparentlymart:
Somehow the other comment I wrote here was lost before submitting. :thinking: Fortunately I still have most of it in my text editor due to it including "ASCII art"!

---

After thinking the positioning of the introduction section in the list I started to think about the positions of the others too. At first I thought similarly to Alexis that the Glossary felt weird to be at the top there since I expect people would visit it only when they encounter a term somewhere else that isn't defined well, but with the introduction moved to the top it would probably be approximately the same height visually as the "Glossary" box, which led met to think about the following arrangement:

```
| Introduction To Terraform       | Terraform Glossary                  |
| ...                             | ...                                 |
|                                 |                                     |
| Terraform CLI                   | Terraform Enterprise                |
| ...                             | ...                                 |
|                                 |                                     |
|                                 |                                     |
|                                 |                                     |
|                                 |                                     |
|                                 |                                     |
| Terraform Registry              |  Terraform GitHub Actions           |
| ...                             |  ...                                |
|                                 |                                     |
|                                 |                                     |
|                                 |                                     |
| Guides and Whitepapers          |                                     |
| ...                             |  Extending Terraform                |
|                                 |  ...                                |
|                                 |                                     |
|                                 |                                     |
|                                 |                                     |
|                                 |                                     |
|                                 |                                     |
```

This layout has a different visual balance and places Terraform CLI and Terraform Enterprise side-by-side, and thus gives them both similar visual prominence on the page, while still slightly highlighting CLI since it's content useful for Enterprise users too.

I'm not really sure that "Learn Terraform" necessarily needs to be its own section here since the "Introduction to Terraform" should presumably lead people over there anyway, and it's also already got a prominent spot on the very top navigation. I think that "Introduction to Terraform" and "Learn Terraform" sound like the same thing to someone who isn't familiar with the fact that the learning guides are managed and published via a separate mechanism.

content/source/docs/index.html.markdown
nfagerlund: Yeah, might as well!

nfagerlund:
@apparentlymart I think I'd like to leave "Learn" in the mix for now, mostly because it's going to continue to expand with task-specific content (in the "advanced tracks") that doesn't exist elsewhere and isn't necessarily directly linked from elsewhere. Definitely willing to reconsider later, tho. Or we could just make sure that every Learn guide is always linked from the front page of the Guides section, or something.

nfagerlund:
🙌🏼 heck yeah green travis build

apparentlymart:
With this now deployed and clickable, I wanted to note that when I first went to this new page I initially was unsure about what to click on for "Introduction to Terraform". It took me a few moment to realize that the heading itself was a link, since there isn't much precedent on our site for headings to be links.

While I expect having secondary links under each heading would make the page feel quite cluttered, perhaps including an arrow dingbat (`➞`, `➤`, `⯈`, etc) after each heading text would give a stronger visual hint that it's a link.

I see that it would conflict with the use of ➡︎ to indicate "external site" in the Learn Terraform heading, but it isn't really presented as a separate site in our information architecture anyway -- it has similar top-level navigation links, and in future `terraform.io` will adopt the same top navigation visual styling anyway to bring the two back together -- so maybe it's not _too_ important for us to flag that one as being special?

PR #716, opened 2019-03-20
Front page: Decorate clickable headers with arrows

nfagerlund:
@apparentlymart suggested doing this because we don't have a strong tradition of clickable headers on terraform.io, so I went ahead and tried it out. 

I think I like it. 

image

### DIFF
--- a/content/source/docs/index.html.markdown
+++ b/content/source/docs/index.html.markdown
@@ -19,13 +19,13 @@ You can navigate the docs via this index page, or you can jump from section to s
 
 <div class="col-md-6 col-sm-12">
 
-### [Introduction to Terraform ➡︎](/intro/index.html)
+### [Introduction to Terraform ➜](/intro/index.html)
 
 -> Visitors who are curious about Terraform can start here.
 
 A broad overview of what Terraform is and why people use it.
 
-### [Terraform CLI ➡︎](/docs/cli-index.html)
+### [Terraform CLI ➜](/docs/cli-index.html)
 
 -> Intermediate and advanced users spend most of their time here.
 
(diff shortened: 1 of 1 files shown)


armchairlinguist:
@nfagerlund I like the idea, maybe slightly thinner arrow shafts though?

apparentlymart:
Now I can see how it looks in context, I agree with @armchairlinguist that these _particular_ arrows feel out of place since they have much more visual weight than the text adjacent to them.

Sadly I don't have a full website development environment ready right now to experiment with, but I wonder if either using arrows without shafts at all (⯈, ➤) or using an arrow whose weight is more similar to the text (➞) would help with the visual balance here.

nfagerlund:
@apparentlymart @armchairlinguist Ok... 

🔥CHOOSE YOUR FIGHTER🔥

image

thrashr888:
I'm into the simpler "Black Rightwards Arrowhead" or http://www.rsaquo.net/ but, why not use underlines? Is that possible? I'm guessing the answer is "it's hard to change the styles in this CMS." :)

nfagerlund:
Man, I'm all ABOUT underlining links!! IDEK why that ever fell out of fashion. 

I kind of don't want to get into it in this PR tho, because 1, it'd affect more of the site than just this page, and 2, it requires more evaluation of how things look in general - for example, check out the underline (I guess it's more of a `border-bottom`, but still) under the non-clickable headers on [this page](https://www.terraform.io/docs/backends/types/s3.html).

armchairlinguist:
I'd go Rightwards Arrow or Heavy Round-Tipped.

apparentlymart:
I noticed that there's already [precedent for a link arrow style on "learn"](https://learn.hashicorp.com/terraform/), but that is an inline SVG image rather than a unicode arrow so we won't be able to match it exactly, but maybe we can pick the Unicode arrow that most closely matches that in our current typeface and then when the Terraform site gets respun to be more like learn maybe that'll be an opportunity to rework this page's content as a full HTML+CSS creation following similar visual cues as learn, rather than just trying to force it with Markdown.

nfagerlund:
Learn's arrow is very close to Heavy Round-Tipped, so Heavy Round-Tipped it is!

image

y/n?

PR #720, opened 2019-03-21
WIP: Add note that an organization token will not work for the oauth-clients API endpoint

justinretzolk:
Copied the note from the Runs API endpoint to make sure it was stylistically consistent


justinretzolk:
Closing this PR as this is no longer accurate.

PR #726, opened 2019-03-25
WIP: Corrected 200 response links

justinretzolk:
Corrected broken links for the `200` response for "List Runs in a Workspace" and "Get run details"

**Note:** I noticed when doing this that defining the link once allows it to be reused throughout the document, yet there are other links that are defined several times. Should we consider defining all API status/response links once at the top of the document so as to not unnecessarily repeat ourselves?


nfagerlund:
Debating a longer-term fix in #728

PR #727, opened 2019-03-25
Use replicated admin instead, as ptfe-admin is a bash alias

evanphx:


evanphx:
I added `--no-tty` to the default because it's just a single string anyway.

armchairlinguist:
I don't think I understand this change well enough to review, alas, but it looks like better minds agree that it's a good plan. :)

PR #728, opened 2019-03-26
Make API status code links easier

nfagerlund:
Previously, when people add or update an API method, we've been having them add their own link anchors for any HTTP status codes they need to reference. 

**Advantages:**

- Well, at first we didn't really even know which status codes we were going to need. So that was a good lightweight way to find out. 

**Disadvantages:**

- Too easy to forget one. 
- Too easy to link to the [wrong](https://github.com/hashicorp/terraform-website/pull/728/files#diff-cb7de1a2065460d6b09fc222f25af2b4L561) status code! 
- Pain in the butt! Lots of duplicated effort.

**Point:**

So, let's just put all the anchors we've ever used so far at the top of every API page, and then we can just reference them by number. 

**Counterpoint:**

Or! should we put an anchor for _every_ status code and never have to update it again? Seems heavyweight; the API is mature enough that if we haven't used a status by now, are we _really_ going to start using it anytime soon?

### DIFF
--- a/content/source/docs/enterprise/api/_template.md
+++ b/content/source/docs/enterprise/api/_template.md
@@ -1,5 +1,25 @@
 Follow this template to format each API method. There are usually multiple sections like this on a given API endpoint page.
 
+<!-- Boilerplate link references: This entire list should be included at the top of every API page, so that the tables can use short links freely. This SHOULD be all of the status codes we use in TFE's API; if we need to add more, update the list on every page. -->
+
+[200]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200
+[201]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201
+[202]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202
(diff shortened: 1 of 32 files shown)


davidcelis:
Part of me feels like we should create a helper method like `status_code_url(status)` that we can use when necessary as opposed to including all anchors on every page… Though this would require changing all of the API doc pages to `*.html.md.erb` since the helpers require ERB i think…

If that's a bit much, I think the boilerplate section you put in the template is fine and we can add to it if we encounter a new status

nfagerlund:
@davidcelis That occurred to me too! And yeah, turning them all to .erb... argh. Honestly that's a great candidate for a reusable component in the ✨ world of the future ✨  the marketing-webdev team is planning, since we should be able to use components in content at that point. 

Although, if we _were_ to turn the api pages into .erb, maybe we can come up with a few other API doc things that would be easier to work with programmatically. Any thoughts?

armchairlinguist:
This seems like a good "for now" solution to me; I agree that in The Future ™️ it would be great to improve, but includes, etc. are quite messy right now and this is straightforward.

I think listing just the ones we use is correct. If for some weird reason we added a new one, it could be included just for that route.

PR #732, opened 2019-03-28
Sentinel doc updates for Terraform 0.12

chrisarcand:
Contains doc updates for the following changes in 0.12:

* Module-level outputs no longer available in `tfstate`
* Unknown values no longer show up as magic UUIDs in `tfplan`'s `applied`
* Raw interpolation strings are no longer available within `tfconfig`
* Consequently, new `references` config key docs in `tfconfig`
* `tfconfig` no longer expresses config blocks of the same type as arrays

### DIFF
--- a/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
@@ -10,19 +10,34 @@ description: |-
 
 The `tfconfig` import provides access to a Terraform configuration.
 
-The Terraform configuration is the set of `*.tf` files that are used
-to describe the desired infrastructure state. This import alone doesn't
-give you access to the state of the infrastructure. To view the state
-of the infrastructure, see the [`tfstate`][import-tfstate] import.
+The Terraform configuration is the set of `*.tf` files that are used to
+describe the desired infrastructure state. Policies using the `tfconfig`
+import can access all aspects of the configuration: providers, resources,
+data sources, modules, and variables.
 
-[import-tfstate]: /docs/enterprise/sentinel/import/tfstate.html
+Some use cases for `tfconfig` include:
 
-Policies using the `tfconfig` import can access all aspects of the
-configuration: providers, resources, data sources, modules, and variables. Note
-that since this is the configuration and not an invocation of Terraform, you
-can't see values for variables, the state, or the diff for a pending plan.
+* **Organizational naming conventions**: requiring that configuration elements
+  are named in a way that conforms to some organization-wide standard.
+* **Required inputs and outputs**: organizations may require a particular set
+  of input variable names across all workspaces or may require a particular
+  set of outputs for asset management purposes.
+* **Enforcing particular modules**: organizations may provide a number of
+  "building block" modules and require that each workspace be built only from
+  combinations of these modules.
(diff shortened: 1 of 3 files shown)


chrisarcand:
Some screenshots of the larger changes, for a better visual:

image

image

content/source/docs/enterprise/sentinel/import/tfplan.html.md
```diff
@@ -415,6 +407,15 @@ import "tfplan"
 main = rule { tfplan.resources.null_resource.foo[0].applied.triggers.foo is "bar" }
 ```
 
+-> Note that some values will not be available in the `applied` state because
+they cannot be known until the plan is actually applied. In Terraform 0.11 or
+earlier, these values are represented by a placeholder (the UUID value
+`74D93920-ED26-11E3-AC10-0800200C9A66`) and in Terraform 0.12 or later they
+are simply omitted (a null value). **In either case**, you should instead use the
```
vancluever: Just ahead of a larger review - remember that it's not a _null_ value, but an _undefined_ value.

Null itself is an explicit value in Sentinel (see [the spec](https://docs.hashicorp.com/sentinel/language/spec#null)).

chrisarcand: Gotcha, fixed!

content/source/docs/enterprise/sentinel/import/tfconfig.html.md
```diff
@@ -588,14 +552,43 @@ main = rule { subject.outputs.id.sensitive }
...
+[references](#value-references) documentation below.
+
+### Value: `references`
+
+* **Value Type:**. List of strings.
+
+-> Note: This value is only present when using Terraform 0.12 or later
+
+The `references` value within the [output namespace](#namespace-outputs)
+represents the identifier within an expression found in [`value`](#value-value).
```
nfagerlund: I think this needs a little more elaboration; what's here doesn't quite tell me what I need to know to use it. 

My GUESS, from extrapolating from what I'm seeing here:

- Either:
    - If the value is any expression other than a literal value, `value` is always empty. This includes concatenating two literals together with a function, or something. 
    - If the value is an expression that _references one or more named values,_ `value` is always empty. Other kinds of transformations, like functions, would still end up in the `value`.
- And either:
    - `references` is a list of every named value that was referenced in the value's expression. This is extracted using the same sort of logic Terraform uses to infer dependencies. 
    - `references` is a list of every... actually, IDK what it would be otherwise, but if `value` works like the first option above, I'd expect something odd.

vancluever: Hey Nick, Chris and I talked about this offline today and it it is being updated to elaborate and also update the other fields where references exists. This should help address this.

nfagerlund: OK, great, thanks!

chrisarcand:
Revamped the whole thing as expanding `references` on every single namespace was basically taking up 80% of the space on the page. (which is ironic, given the the most useful utility of this import is for object names and enforcing rules around resources, providers, etc. Given this, I've also added more to the introduction of why you'd want to use this import.)

Some screenshots:

image

image

image

image

image

PR #733, opened 2019-03-29
Remove Preview Release disclaimer

mocofound:
As per conversation in product-engineering Slack channel on 03/28/2019,  "The Terraform `remote` backend is production-ready as of v0.11.13." Therefore, we can remove the preview release disclaimer.  On the first link to the terraform remote backend doc, we specify that v0.11.13 or newer versions are recommended so I opted not to include that note again on this page for simplicity.

Old Disclaimer:
**Preview Release:** As of Terraform 0.11.8, the remote backend is a preview release, and we do not recommend using it with production workloads. Please continue to use the existing Terraform Enterprise (`atlas`) backend or the [TFE CLI tool][tfe-cli] for production workspaces.


nfagerlund:
Ah thanks! I just saw that in the page the other day and was like "didn't we..?" but was in the middle of something I couldn't drop.

PR #734, opened 2019-03-29
End constant, error-prone fussing with `sidebar_current`

nfagerlund:
It's time to stop using `sidebar_current` in any capacity and just auto-detect which sidebar nav sections should be activated based on the current URL. This PR does so, by adding a class to any sidebar links pointed at the current page during the Middleman build and then using JQuery to ensure the current page's parent sections are activated. 

The code in this PR works everywhere on the site with better accuracy than the current status quo, both on initial load and on subsequent click-through navigation. It respects the `.nav-visible` hack used for sections that always remain open, as used extensively in the provider docs (among other places). In the event that a user has JavaScript disabled, it displays the entire sidebar navigation in expanded mode to ensure the site stays usable. In my testing it does not appear to cause any additional jank.

### Why

`sidebar_current` functions primarily as a cruel prank on the people I rely on to help keep the docs updated.

- **`sidebar_current` is pointless.** After a year and a half working in the Terraform docs, I can confidently say there is _no_ situation in which we purposely use `sidebar_current` to activate anything other than the current page and all of its parent lists. The double-bookkeeping buys us nothing.
- **`sidebar_current` is error-prone.** Just this week we shipped new provider docs for JDCloud, and they helpfully illustrate several of the ways sidebar_current constantly goes wrong: 
    - [Non-activation](https://www.terraform.io/docs/providers/jdcloud/jdcloud_network_acl.html): Make a typo, and your sidebar breaks. 
    - Non-activation with children (doesn't happen in this provider thankfully): If the page you typoed is an index page with a sub-list of children, all those child pages become completely inaccessible.
    - [Double-activation](https://www.terraform.io/docs/providers/jdcloud/jdcloud_network_interface_attachment.html): Since the slug check uses a naïve `.start_with?()` check, it's easy to accidentally activate siblings. 
- **Fixing these errors wastes disproportionate time.** Provider docs are the place where people are most likely to screw this up (since provider maintainers spend less time editing the website than I do, and thus haven't internalized all these unnecessary nonsense rules). Fixing provider docs involves cloning and PR-ing a new repo and getting approval from a new set of maintainers, then cherry-picking the fix onto stable-website once it's merged. Acceptable for a meaningful edit, but maddening for problems that shouldn't have been possible in the first place. 

### Review

Here are the types of PR feedback I'm most interested in:

- "Here is a patch to do the DOM manipulation part of this in the static build instead of client-side, along with proof that it doesn't cause an unacceptable spike in build/deploy times."
    - (I know I could do this with Nokogiri, but prior experience told me to Don't. If you have a better way, I 100% want to hear it.)
- "Here is where to put that JS snippet such that `application.js` stays clean and our other sites that are still on old-school middleman can trivially use it."
- "Here is video evidence of new visible jank caused by this patch."

I'd like to keep discussion here focused. Suggestions for other theoretical improvements (auto-generating navs or portions of navs, creating a DSL for expressing navs more succinctly, etc.) are appreciated, but should go in follow-up issues or PRs.

### DIFF
--- a/content/middleman_helpers.rb
+++ b/content/middleman_helpers.rb
@@ -48,12 +48,7 @@ module Helpers
   # This helps by setting the "active" class for sidebar nav elements
   # if the YAML frontmatter matches the expected value.
   def sidebar_current(expected)
-    current = current_page.data.sidebar_current || ""
-    if current.start_with?(expected)
-      return " class=\"active\""
-    else
-      return ""
-    end
+    return ""
   end
 
   # Returns the id for this page.
--- a/content/source/assets/javascripts/application.js
+++ b/content/source/assets/javascripts/application.js
@@ -10,6 +10,6 @@
 document.addEventListener("turbolinks:load", function() {
     "use strict";
     var docsSidebar = $("#docs-sidebar ul.nav.docs-sidenav");
-    var activeListItems = docsSidebar.find("li").has("a.current-page");
-    activeListItems.addClass("active");
+    docsSidebar.find("ul.nav").addClass("nav-hidden");
+    docsSidebar.find("li").has("a.current-page").addClass("active");
 });
--- a/content/source/assets/javascripts/application.js
+++ b/content/source/assets/javascripts/application.js
@@ -6,3 +6,10 @@
 //= require hashicorp/analytics
 
 //= require analytics
+
+document.addEventListener("turbolinks:load", function() {
+    "use strict";
+    var docsSidebar = $("#docs-sidebar ul.nav.docs-sidenav");
+    var activeListItems = docsSidebar.find("li").has("a.current-page");
+    activeListItems.addClass("active");
+});
--- a/content/source/assets/stylesheets/_docs.scss
+++ b/content/source/assets/stylesheets/_docs.scss
@@ -76,8 +76,10 @@
     }
 
     // subnav
-    ul.nav {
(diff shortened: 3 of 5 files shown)


nfagerlund:
Nice nice nice, security.html.erb uses inner.erb as its layout but doesn't set content for `:sidebar`, and that makes things blow up. `undefined method 'gsub!' for nil:NilClass`. Easy fix tho.

jescalan:
I feel your frustration here -- this is 100% valid. Just so you're aware, we do have [a new version of the sidebar](https://hashi-web-components.netlify.com/?component=docsSidenav) implemented that does not suffer from this issue and will roll out slowly alongside the new docs designs.

I think probably you knew this already and knowing that it will be a while before the new designs roll out did this to fix in the meantime, but wanted to make extra sure you (or any observers) were also aware 😁

nfagerlund:
@jescalan Thanks for the note! Looking at https://github.com/hashicorp/vault/blob/master/website/source/layouts/docs.erb, it seems like it needs directory names and filenames without extensions, and then it just extracts everything else from the page metadata? If so, good, that's pretty much what I would have asked you for. 👍🏼

nfagerlund:
Jeff confirmed that that was an approval, and no one else who saw this noticed anything wrong with the syntax. 🚢 toot toot

PR #735, opened 2019-03-29
Show correct command for reloading Replicated settings

glasner:
... because rerunning installer does not in fact reload the settings. More info:

https://app.asana.com/0/331272348117180/1112824705991543

### DIFF
--- a/content/source/docs/enterprise/private/automating-the-installer.html.md
+++ b/content/source/docs/enterprise/private/automating-the-installer.html.md
@@ -301,11 +301,19 @@ If the installation script (`install.sh`) exits successfully, but the Replicated
 * Verify the locations and permissions of the files. The Replicated configuration file should be placed in `/etc/replicated.conf`, and the application settings file should be placed in the path specified in the Replicated configuration file. The permissions of both files should be `600` if owned by the `replicated` user, or `644`.
 * Run a JSON validator on the files to check their validity.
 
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/private/automating-the-installer.html.md
```diff
@@ -301,11 +301,17 @@ If the installation script (`install.sh`) exits successfully, but the Replicated
 * Verify the locations and permissions of the files. The Replicated configuration file should be placed in `/etc/replicated.conf`, and the application settings file should be placed in the path specified in the Replicated configuration file. The permissions of both files should be `600` if owned by the `replicated` user, or `644`.
 * Run a JSON validator on the files to check their validity.
 
-After resolving any issues with file validity or permissions, re-run the installer to verify that the configuration is applied.
+After resolving any issues with file validity or permissions, run the following command to reload your configuration:
+
+``` bash
+replicatedctl app-config import < /etc/replicated.conf
```
armchairlinguist: Hm...this doesn't seem correct to me. app-config export exports app settings (the application settings file referred to above), not the replicated.conf file itself.

Can we confirm which file(s) this should import? If it is the application settings, how do they reload replicated.conf? Or is it not necessary?

glasner: Following up w/ Replicated to be sure

PR #740, opened 2019-04-02
Document shallow workspace endpoints

chrisarcand:
The following routes are being added and should be documented:

```
    GET    /workspaces/:workspace_id
    PATCH  /workspaces/:workspace_id
    DELETE /workspaces/:workspace_id
```

Note that these are a bit different as this is the first time we're specifying two endpoints to return the same result.

### DIFF
--- a/content/source/docs/enterprise/api/workspaces.html.md
+++ b/content/source/docs/enterprise/api/workspaces.html.md
@@ -298,16 +298,24 @@ _Migrating a legacy environment_
 
 ## Update a Workspace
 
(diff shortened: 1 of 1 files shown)


dylanegan:
Is this the first time in the documentation that we're specifying two endpoints to return the same result?

chrisarcand:
Yup. I kept it very simple for now but should we add the other shallow actions, an entirely separate section at the bottom (with references to it in the nested docs) should probably be made for clarity.

chrisarcand:
@nfagerlund I agree, I had them inlined before and I don't really know why I decided to split them off. Rebased and made some tweaks, ready for a final glance 👀

PR #741, opened 2019-04-02
In-page quick-nav menu

nfagerlund:
This adds a small drop-down menu at the top of each page. When clicked,
it reveals an in-page hierarchical TOC which you can use to jump directly to any
of the H2, H3, or H4 headers in the primary content. The TOC can be dismissed by
clicking outside it or by re-clicking the trigger.

**gif:** 

quicknav

**desktop:** 

Screenshot_2019-04-02 Workspaces - API Docs - Terraform Enterprise - Terraform by HashiCorp(1)

**mobile:**

This re-uses the SCSS from the top-nav dropdown, and uses a little bit of JQuery
to do the rest of the lifting.

### DIFF
--- a/content/source/assets/javascripts/application.js
+++ b/content/source/assets/javascripts/application.js
@@ -23,27 +23,32 @@ document.addEventListener("turbolinks:load", function() {
     // H2/H3/H4 headers.
     var headers = $('#inner').find('h2, h3, h4');
     if (headers.length > 0) {
+        // Build the quick-nav HTML:
         $("#inner #inner-quicknav").html(
             '<span id="inner-quicknav-trigger">Page Quick Nav<svg width="9" height="5" xmlns="http://www.w3.org/2000/svg"><path d="M8.811 1.067a.612.612 0 0 0 0-.884.655.655 0 0 0-.908 0L4.5 3.491 1.097.183a.655.655 0 0 0-.909 0 .615.615 0 0 0 0 .884l3.857 3.75a.655.655 0 0 0 .91 0l3.856-3.75z" fill-rule="evenodd"/></svg></span><ul class="dropdown"></ul>'
         );
-        var quickNavTrigger = $('#inner-quicknav #inner-quicknav-trigger');
         var quickNav = $('#inner-quicknav > ul.dropdown');
         headers.each(function(index, element) {
             var level = element.nodeName.toLowerCase();
             var header_text = $(element).text();
             var header_id = $(element).attr('id');
-            quickNav.append(`<li class="level-${level}"><a href="#${header_id}">${header_text}</a></li>`);
+            quickNav.append('<li class="level-' + level + '"><a href="#' + header_id + '">' + header_text + '</a></li>');
         });
-        quickNavTrigger.on('click', function(e) {
+        // Attach event listeners:
+        // Trigger opens and closes.
+        $('#inner-quicknav #inner-quicknav-trigger').on('click', function(e) {
             $(this).siblings('ul').toggleClass('active');
             e.stopPropagation();
         });
+        // Clicking inside the quick-nav doesn't close it.
         quickNav.on('click', function(e) {
             e.stopPropagation();
         });
+        // Jumping to a section means you're done with the quick-nav.
         quickNav.find('li a').on('click', function() {
             quickNav.removeClass('active');
         });
+        // Clicking outside the quick-nav closes it.
         $('body').on('click', function() {
             quickNav.removeClass('active');
         });
--- a/content/source/assets/javascripts/application.js
+++ b/content/source/assets/javascripts/application.js
@@ -7,20 +7,20 @@
 
 //= require analytics
 
-// In navigation sidebars, hide most sub-lists and reveal any that contain the
-// current page. The a.current-page class is added during build by
-// layouts/inner.erb.
+// Set up terraform.io UI helpers
 document.addEventListener("turbolinks:load", function() {
     "use strict";
+
+    // In navigation sidebars, hide most sub-lists and reveal any that contain the
+    // current page. The a.current-page class is added during build by
+    // layouts/inner.erb.
     var docsSidebar = $("#docs-sidebar ul.nav.docs-sidenav");
     docsSidebar.find("ul.nav").addClass("nav-hidden");
     docsSidebar.find("li").has("a.current-page").addClass("active");
-});
 
-// On docs/content pages, add a hierarchical quick nav menu if there are any
-// H2/H3/H4 headers.
-document.addEventListener("turbolinks:load", function() {
-    "use strict";
+
+    // On docs/content pages, add a hierarchical quick nav menu if there are any
+    // H2/H3/H4 headers.
     var headers = $('#inner').find('h2, h3, h4');
     if (headers.length > 0) {
         $("#inner #inner-quicknav").html(
@@ -28,7 +28,7 @@ document.addEventListener("turbolinks:load", function() {
         );
         var quickNavTrigger = $('#inner-quicknav #inner-quicknav-trigger');
         var quickNav = $('#inner-quicknav > ul.dropdown');
-        headers.each( function(index, element) {
+        headers.each(function(index, element) {
             var level = element.nodeName.toLowerCase();
             var header_text = $(element).text();
             var header_id = $(element).attr('id');
--- a/content/source/assets/javascripts/application.js
+++ b/content/source/assets/javascripts/application.js
@@ -13,3 +13,32 @@ document.addEventListener("turbolinks:load", function() {
     docsSidebar.find("ul.nav").addClass("nav-hidden");
     docsSidebar.find("li").has("a.current-page").addClass("active");
 });
+
+document.addEventListener("turbolinks:load", function() {
+    "use strict";
+    $("#inner #inner-quicknav").html(
+        '<span id="inner-quicknav-trigger">Quick Nav<svg width="9" height="5" xmlns="http://www.w3.org/2000/svg"><path d="M8.811 1.067a.612.612 0 0 0 0-.884.655.655 0 0 0-.908 0L4.5 3.491 1.097.183a.655.655 0 0 0-.909 0 .615.615 0 0 0 0 .884l3.857 3.75a.655.655 0 0 0 .91 0l3.856-3.75z" fill-rule="evenodd"/></svg></span><ul class="dropdown"></ul>'
+    );
+    var quickNavTrigger = $('#inner-quicknav #inner-quicknav-trigger');
+    var quickNav = $('#inner-quicknav > ul.dropdown');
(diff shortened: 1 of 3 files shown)


armchairlinguist:
This is SO EXTREMELY GOOD. :terraform-da: 💜
What do you think about calling it "Page Quick Nav" for a bit more clarity?

nfagerlund:
"Page Quick Nav" sounds fine!

content/source/assets/javascripts/application.js
```diff
@@ -13,3 +13,34 @@ document.addEventListener("turbolinks:load", function() {
     docsSidebar.find("ul.nav").addClass("nav-hidden");
     docsSidebar.find("li").has("a.current-page").addClass("active");
 });
+
+document.addEventListener("turbolinks:load", function() {
+    "use strict";
+    var headers = $('#inner').find('h2, h3, h4');
+    if (headers.length > 0) {
+        $("#inner #inner-quicknav").html(
+            '<span id="inner-quicknav-trigger">Quick Nav<svg width="9" height="5" xmlns="http://www.w3.org/2000/svg"><path d="M8.811 1.067a.612.612 0 0 0 0-.884.655.655 0 0 0-.908 0L4.5 3.491 1.097.183a.655.655 0 0 0-.909 0 .615.615 0 0 0 0 .884l3.857 3.75a.655.655 0 0 0 .91 0l3.856-3.75z" fill-rule="evenodd"/></svg></span><ul class="dropdown"></ul>'
```
nfagerlund:
```suggestion
            '<span id="inner-quicknav-trigger">Page Quick Nav<svg width="9" height="5" xmlns="http://www.w3.org/2000/svg"><path d="M8.811 1.067a.612.612 0 0 0 0-.884.655.655 0 0 0-.908 0L4.5 3.491 1.097.183a.655.655 0 0 0-.909 0 .615.615 0 0 0 0 .884l3.857 3.75a.655.655 0 0 0 .91 0l3.856-3.75z" fill-rule="evenodd"/></svg></span><ul class="dropdown"></ul>'
```

content/source/assets/javascripts/application.js
```diff
@@ -7,9 +7,45 @@
...
 document.addEventListener("turbolinks:load", function() {
     "use strict";
     var docsSidebar = $("#docs-sidebar ul.nav.docs-sidenav");
     docsSidebar.find("ul.nav").addClass("nav-hidden");
     docsSidebar.find("li").has("a.current-page").addClass("active");
 });
+
+// On docs/content pages, add a hierarchical quick nav menu if there are any
+// H2/H3/H4 headers.
+document.addEventListener("turbolinks:load", function() {
```
jescalan: Why put this in a different load function?

content/source/assets/javascripts/application.js
```diff
@@ -7,9 +7,45 @@
...
+    "use strict";
```
jescalan: any reason why this is only in this function rather than at the top level?

content/source/assets/javascripts/application.js
```diff
@@ -7,9 +7,45 @@
...
+    var headers = $('#inner').find('h2, h3, h4');
+    if (headers.length > 0) {
+        $("#inner #inner-quicknav").html(
+            '<span id="inner-quicknav-trigger">Page Quick Nav<svg width="9" height="5" xmlns="http://www.w3.org/2000/svg"><path d="M8.811 1.067a.612.612 0 0 0 0-.884.655.655 0 0 0-.908 0L4.5 3.491 1.097.183a.655.655 0 0 0-.909 0 .615.615 0 0 0 0 .884l3.857 3.75a.655.655 0 0 0 .91 0l3.856-3.75z" fill-rule="evenodd"/></svg></span><ul class="dropdown"></ul>'
+        );
+        var quickNavTrigger = $('#inner-quicknav #inner-quicknav-trigger');
+        var quickNav = $('#inner-quicknav > ul.dropdown');
+        headers.each( function(index, element) {
```
jescalan: Is the space before the function on purpose?

content/source/assets/javascripts/application.js
```diff
@@ -7,9 +7,45 @@
...
+        });
+        quickNav.on('click', function(e) {
+            e.stopPropagation();
+        });
+        quickNav.find('li a').on('click', function() {
+            quickNav.removeClass('active');
+        });
+        $('body').on('click', function() {
+            quickNav.removeClass('active');
+        });
```
jescalan: Just a general question - since turbolinks prevents a full reload, does every page navigation transition continue stacking more and more of these event listeners? If so, maybe should we remove the event listeners somehow - didn't see a "before route change" event [here](https://github.com/turbolinks/turbolinks#full-list-of-events)

content/source/assets/javascripts/application.js
nfagerlund: No good reason! I'll have it move in with the other one.

content/source/assets/javascripts/application.js
nfagerlund: - I thought it was only function scope? 
- If not, would it mess with anything being pulled in by the `//= require` statements?

Anyway, I'll defer to you!

content/source/assets/javascripts/application.js
nfagerlund: JS is one of my more recently-acquired tongues, and sometimes I go overboard on trying to keep it readable for my future self. Probably not necessary.

content/source/assets/javascripts/application.js
nfagerlund: That's a great question. Lemme see if I can find out using the inspector tools.

I can't seem to do anything that causes doubled event listeners. After reading more, it looks like: 

- [It at least doesn't cache event listeners.](https://github.com/turbolinks/turbolinks#understanding-caching) So they shouldn't get duplicated when someone navigates history w/ back and forward.
- Turbolinks prevents a full reload, but it replaces the whole `<body>` element, right? That _should_ cause it to discard any elements that have event listeners attached.

So, it seems like we should be good. But it's also possible I'm not seeing prod-like behavior in my preview... I'll build a complete copy of the site and test with that, as well.

nfagerlund: Sure enough, even with a pre-built copy nothing seems to cause duplicated event listeners. The replacement of the entire `<body>` element and the fact that it doesn't cache event listeners for history navigation takes care of it, as expected.

nfagerlund: Hmm, thinking more about this: Turbolinks WOULD probably cache the `<span>` and `<ul>` that we add to the div, since those are new DOM elements. But since we're adding them with `.html("etc")`, it's actually idempotent -- `.html()` replaces the current contents of that div. 

Which means that on back/forward navigation, we aren't messing anything up, but we ARE doing a small amount of extra work by rebuilding the quick-nav. 

I could throw that into some `if` statements so reattach the event listeners but don't rebuild the elements. But that might be premature optimization, since it doesn't really feel bad as-is.

content/source/assets/javascripts/application.js
```diff
@@ -7,9 +7,51 @@
...
+    // layouts/inner.erb.
     var docsSidebar = $("#docs-sidebar ul.nav.docs-sidenav");
     docsSidebar.find("ul.nav").addClass("nav-hidden");
     docsSidebar.find("li").has("a.current-page").addClass("active");
+
+
+    // On docs/content pages, add a hierarchical quick nav menu if there are any
+    // H2/H3/H4 headers.
+    // Check whether turbolinks cached the quick-nav elements:
+    if ($("#inner #inner-quicknav ul").length == 0) {
```
alisdair: Is this ever non-zero, in your experience? I haven't tried it locally, but I thought turbolinks cached pages _before_ calling `turbolinks:load` handlers, so I don't understand how it would ever save the generated quick-nav.

content/source/assets/javascripts/application.js
```diff
@@ -7,9 +7,51 @@
...
+    });
+    quickNav.on('click', function(e) {
+        e.stopPropagation();
+    });
+    quickNav.find('li a').on('click', function() {
+        quickNav.removeClass('active');
+    });
+    $('body').on('click', function() {
+        quickNav.removeClass('active');
+    });
```
alisdair: I think this event handler is going to be added again and again for every page transition. You might want to namespace it, and remove it first? `$('body').off('click.quicknav'); $('body').on('click.quicknav', function() { …`

content/source/assets/javascripts/application.js
jescalan: It can be global as well, but that's a fair question about messing with other statements. Honestly I don't think you have any code here that violates strict mode anyway so it probably makes little difference, but usually i see people who use it putting it right at the top of files.

content/source/assets/javascripts/application.js
jescalan: Great investigation! It seems like, [so long as when the element is removed no references to it remain](https://stackoverflow.com/questions/12528049/if-a-dom-element-is-removed-are-its-listeners-also-removed-from-memory), we have no potential for a memory leak here. I assume that turbolinks isn't holding on to internal references in this manner, and if it did, someone would have called it out a while ago, so we're probably good here.

content/source/assets/javascripts/application.js
jescalan: Haha I had the same concern in my review above, but it seems as though if turbolinks replaces the entire body element and no references to the previous element remain, the event will be removed and garbage collected. It does still make me uncomfortable to see this though.

alisdair: I don't think turbolinks replaces the `body` element, just its contents. For event handlers on all other elements I think we're safe, but I think `body` is a special case. I could be mistaken though!

jescalan: Took a quick look in their docs and it says it does, although it might be worth verifying: https://github.com/turbolinks/turbolinks#navigating-with-turbolinks

nfagerlund: Ah, because it's on `<body>` instead of a sub-element? 

I tested this by having that handler just log something to the console, and navigated across a whole bunch of pages. But it just kept logging only once per click. So, I think that means it's not being re-added. 

But if you think it's good practice to do the off/on thing, I can do it anyway!

content/source/assets/javascripts/application.js
nfagerlund: Oh, I might have misunderstood the timing on the cache! Yeah, I've never managed to make it be non-zero. I guess I might as well not test for that after all.

content/source/assets/javascripts/application.js
nfagerlund: lol sorry, I commented without reloading the page first. Well, anyway, I verified it. 😅

nfagerlund:
Ok! Got functionality approval, got JS approval, responded to review comments (ended up removing the cache check — it seems like it'll never be a cache hit, which means it's a useless DOM query)... I'm doing this.

PR #744, opened 2019-04-03
Expandable sidebars

nfagerlund:
This PR makes the nav sidebars expandable/collapsable. 

- Chevrons only appear on nav items with subnavs. Normal nav items just have dots.
- Clicking the chevrons opens and closes the subnavs without navigating anywhere.
- Links work like they always have (navigate to destination AND expand subnav once there). Need this for backward compatibility.
    - Dummy subnav headers (links that don't go anywhere because their href starts with `#`) are a special case -- they have the same effect as clicking the chevron. (Previous behavior: probably jump you to the top of the page and lose your spot. See aws provider in the prod site for example.)

        This means our dummy headers now work the same way as subnav headers in the new sidebar component on vaultproject.io. (They use a pattern where subnavs are _only_ folders, and the index.html file goes in an "overview" item at the top of the sub-list.) That gives us the option of incrementally transitioning to that style of sidebar nav.

        (Why does vaultproject.io use that style? Better discoverability of the nav collapse/expand, more consistent overall behavior, and easier targeting on mobile.)
- Subnavs that are marked as always-open in their .erb file work like they always have, except now the user can temporarily close them if they're in the way (like that "data sources" section in the aws provider always is). 

### Questions for reviewers:

@phinze and @armchairlinguist and @rkst @apparentlymart and anyone else involved in Terraform docs: 

- Is this good to have? 
- Is the behavior what you'd expect?

@mwickett or anyone else with JavaScript chops:

- Is this code OK to check in? 
- Is there a better way to do any of this?

### Criteria

If I get consensus that this is good to have, plus a head nod from web platform, I'll treat that as approval to rebase and merge. 

### Gif

rotate2

### DIFF
--- a/content/source/assets/javascripts/application.js
+++ b/content/source/assets/javascripts/application.js
@@ -15,11 +15,10 @@ document.addEventListener("turbolinks:load", function() {
     // current page. The a.current-page class is added during build by
     // layouts/inner.erb.
     var docsSidebar = $("#docs-sidebar ul.nav.docs-sidenav");
-    docsSidebar.find("ul.nav").addClass("nav-hidden");
-    docsSidebar.find("li").has("a.current-page, ul.nav-visible").addClass("active");
+    docsSidebar.find("li").has(".current-page, .nav-visible").addClass("active");
+    var subNavs = docsSidebar.find("ul").addClass("nav-hidden").parent("li");
 
     // Make sidebar navs expandable
-    var subNavs = $("ul.nav.docs-sidenav li").has("ul");
     subNavs.addClass("has-subnav");
     subNavs.on("click", function(e) {
         if (e.target == this) {
--- a/content/source/assets/javascripts/application.js
+++ b/content/source/assets/javascripts/application.js
@@ -16,7 +16,17 @@ document.addEventListener("turbolinks:load", function() {
     // layouts/inner.erb.
     var docsSidebar = $("#docs-sidebar ul.nav.docs-sidenav");
     docsSidebar.find("ul.nav").addClass("nav-hidden");
-    docsSidebar.find("li").has("a.current-page").addClass("active");
+    docsSidebar.find("li").has("a.current-page, ul.nav-visible").addClass("active");
+
+    // Make sidebar navs expandable
+    var subNavs = $("ul.nav.docs-sidenav li").has("ul");
+    subNavs.addClass("has-subnav");
+    subNavs.on("click", function(e) {
+        if (e.target == this) {
+            $(this).toggleClass("active");
+        }
+        e.stopPropagation();
+    });
 
 
     // On docs/content pages, add a hierarchical quick nav menu if there are any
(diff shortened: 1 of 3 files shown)


nfagerlund:
(Base branch is set as the quick-nav one, since that's not merged yet and this all kind of builds on previous work. If this reaches mergeable status, I'll rebase and target master.)

armchairlinguist:
This looks like a really nice addition. You had me at "collapse the AWS data sources" :) I really like how compact the TFE sidebar looks with this, as well.

nfagerlund:
I've gotten Zach Shilton's review on the design of this out of band, and made some improvements he suggested. (Updated the gif accordingly!) He and Jeff have said they're cool with merging this approach, so at this point mostly I just want someone to double-check my Javascript.

bflad:
I apologize my JS skills are pretty much abandonware at this point so I can't be too useful there, but I will certainly 🥇 💯 👍 🚀 this for the Terraform AWS Provider documentation.

content/source/assets/javascripts/application.js
```diff
@@ -15,8 +15,35 @@ document.addEventListener("turbolinks:load", function() {
...
+    // Navs that don't want that can leave it out.
+    var globalExpand = true;
+    $("#docs-sidebar a.subnav-toggle").on("click", function(e) {
+        e.preventDefault();
+        if (globalExpand) {
+            subNavs.addClass("active");
+        } else {
+            subNavs.removeClass("active");
+        }
+        globalExpand = !globalExpand;
```
alisdair: I don't follow why we need a JavaScript state variable here instead of using `toggleClass("active")`. What am I missing?

content/source/assets/javascripts/application.js
```diff
@@ -15,8 +15,35 @@ document.addEventListener("turbolinks:load", function() {
...
+    // Make sidebar navs expandable
+    subNavs.addClass("has-subnav");
+    subNavs.on("click", function(e) {
+        if (e.target == this) {
+            $(this).toggleClass("active");
+        }
+        e.stopPropagation();
+    });
+    // For subnavs that don't link to a page, use the whole header as a toggle.
+    docsSidebar.find("a[href^='#']").on("click", function(e) {
```
alisdair: Are there ever any subnav links that are to real in-page anchors? Would this prevent them from working? Instead, maybe we want to use `a[href='#']` to match only empty anchor links.

content/source/assets/javascripts/application.js
nfagerlund: When you click that toggle, most likely there are a few subnavs open already! So if I just used `toggleClass`, it would _close_ the ones that weren't currently closed, and when that happens it feels like you're being pranked!

content/source/assets/javascripts/application.js
nfagerlund: Only as a workaround for bad behavior! They always just link to their own anchor, never to a spot in the content area.

For example, I added "real" anchors to the subnav headers in the [functions list](https://www.terraform.io/docs/configuration/functions.html), because otherwise when you clicked one by mistake, you'd rocket to the top of the page. The backend types sidebar has the same thing. The massive providers don't, which is probably frustrating. 

Since we only added those anchors to suppress a bug, I'm not concerned about preserving them.

PR #745, opened 2019-04-05
Fix Sass "extend compound selector" deprecation warnings on preview and build

nfagerlund:
Sass has been giving us sass about using deprecated syntax. (For details, see
https://github.com/sass/sass/issues/1599.) This commit just moves the affected
rules into placeholder classes and extends those, which silences the warnings
while giving the exact same behavior.

(We could probably just put up with it until we jump off this sinking ship, but
we _did_ get an unexpected Sass update recently while doing some security fixes,
and I can't guarantee that won't happen again before we're on the new platform.)

### DIFF
--- a/content/source/assets/stylesheets/_header.scss
+++ b/content/source/assets/stylesheets/_header.scss
@@ -26,7 +26,7 @@
 
       svg.logo {
         transition: opacity 0.15s ease-in-out;
-        @extend svg.logo.white;
+        @extend %logo-white;
 
         &:hover,
         &:focus,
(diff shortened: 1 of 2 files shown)


nfagerlund:
Reviewing this doesn't actually require any SCSS knowledge, btw. To review:

- Check out the branch and run a local preview server.
- Nagivate to the preview's front page, then to one other page (so you can see both the white-background header and the purple-background one). 

- [ ] Does the terminal where you ran the preview _not_ complain about the deprecation anymore?
- [ ] Does the logo look the same as on the prod side?
- [ ] Does the logo act the same when you hover over it as on the prod side?

PR #746, opened 2019-04-05
Clean up some fussy details of how the "other docs" sidebar section works

nfagerlund:
This had two problems:

- I wanted to replace the header on the docs homepage, and was doing it in a
  fairly dirty way to avoid causing havoc in other pages that need that snippet.
- The snippet would blow up if you invoked it without passing a value for
  `skip` (NilClass doesn't have `#downcase`), which meant you had to make up
  some nonsense if you didn't want to skip anything.

### DIFF
--- a/content/source/layouts/docs-frontpage.erb
+++ b/content/source/layouts/docs-frontpage.erb
@@ -1,10 +1,7 @@
 <% wrap_layout :inner do %>
   <% content_for :sidebar do %>
-    <h4><a href="/docs/index.html">Terraform Docs</a></h4>
 
-    <%= partial(
-                "layouts/otherdocs", :locals => { :skip => "none" }
-        ).sub(%r{^ *<h4>Other Docs</h4> *$}, '').gsub(/class="back" /, '') %>
+    <%= partial("layouts/otherdocs", locals: { header: "Terraform Docs" }) %>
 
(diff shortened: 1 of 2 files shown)


nfagerlund:
whoops, that wasn't supposed to have a terraform submodule update in it. Amended and re-pushed.

content/source/layouts/docs-frontpage.erb
```diff
@@ -1,10 +1,7 @@
 <% wrap_layout :inner do %>
   <% content_for :sidebar do %>
-    <h4><a href="/docs/index.html">Terraform Docs</a></h4>
 
-    <%= partial(
-                "layouts/otherdocs", :locals => { :skip => "none" }
-        ).sub(%r{^ *<h4>Other Docs</h4> *$}, '').gsub(/class="back" /, '') %>
```
alisdair: We used to remove `class="back"` and now we don't. Is that change intentional? I can't find a CSS definition for `.back` anywhere…

nfagerlund: Yeahhhh I kind of dropped .back from the styling when I made the nav sections collapsable/expandable, so it's inert now. Could bring it back sometime, but as a reader I always found it inscrutable.

PR #748, opened 2019-04-08
Add "jump to content" link on mobile

nfagerlund:
image

This link is only visible if the viewport is narrow enough that the nav sidebar
collapses. At that point, you need to scroll past the entire nav content before
you reach the page content, which makes the site very tiring on mobile.

Having to tap an extra link before reading still isn't the best of experiences,
but it's nicer than scrolling all that way.


nfagerlund:
This one doesn't even really do anything, so I don't need to hunt down any code review this time. 😅 It's more of just an "is this an OK idea" check.

nfagerlund:
This was popular in the product sync too. goin' for it.

PR #749, opened 2019-04-08
Increase text contrast

nfagerlund:
Users have sometimes pointed out that the text contrast on terraform.io
is low. More recent updates to our web properties use text colors in
the #3x3x3x sort of range (for example, vaultproject.io uses #373942),
which feels like a better fit for us as well.

**before/after:** 

contrast

### DIFF
--- a/content/source/assets/stylesheets/_variables.scss
+++ b/content/source/assets/stylesheets/_variables.scss
@@ -1,7 +1,7 @@
 // Colors
 $white: #FFFFFF;
 $black: #000000;
-$gray-darker: #555555;
+$gray-darker: #333333;
 
 $consul-pink: #D62783;
 $consul-pink-dark: #961D59;


content/source/assets/stylesheets/_variables.scss
```diff
@@ -25,7 +25,7 @@ $font-weight-reg:       400;
 $font-weight-bold:      600;
 
 // Body
-$body-font-color: $gray-darker;
+$body-font-color: #333;
```
apparentlymart: Peeking at the rest of the CSS, it looks like the only other place we use `$gray-darker` is the border of the "Feature Cards" on the home page:

https://github.com/hashicorp/terraform-website/blob/f4d46f7989a92efa5de2becf0bea917f411c7cdc/content/source/assets/stylesheets/_home.scss#L170-L172



I assume this matching of the text color inside the boxes with the border color was intentional. With that said, perhaps we should just change `$gray-darker` to be `#333` and let these both get darker together.

nfagerlund: Looks like we also use it for `$home-header-link-color`, but yeah, I think probably that should be darker too!

nfagerlund:
Got approvals from Zach Shilton and from people in the product sync, so it looks like this is super uncontroversial.

PR #750, opened 2019-04-10
Filterable sidebar navs

nfagerlund:
Full behavior of this should hopefully be completely intuitive, but it takes a moment to explain, so let's start with the customary gif. 

filter15

### ok, so,

- We now automatically add expand/filter controls to "large sidebars." Currently that's defined as sidebars with more than 30 links in them.
    - You can choose a place for the controls (like putting them below a header) or let them take their default place. (Probably best to let them default.)
- The expand all/collapse all control has changed its behavior slightly: 
    - "Collapse all" is replaced by "reset," which always returns the sidebar to its initial state (current page, default-open subnavs, and their parents are activated). 
    - It's a button.
    - It displays its current mode, so you know what it's going to do when you click.
- There's a ✨filter field✨.
    - Focus by clicking in it, or by typing slash (/) on the page while not inside a text field. 
        - Squatting on slash is somewhat aggressive and overrides a power-user Firefox feature, but it's also a well-trodden path that power-users generally understand by now. Try typing slash on GitHub, Gmail, most other Google properties, DuckDuckGo, Twitter, and more. (Or `less`.)
    - Type to filter. Filter text matches against _the text of the sidebar links,_ not deep page content or anything. 
    - While filtering, sidebar items default to hidden. 
    - Items that match the filter are shown, as are all of their parents. 
    - If a filter matches a subnav header (like "API" or EC2 in the example gif), that also reveals all the _direct_ children of that subnav even if they don't match the filter, since we assume all of those items are relevant to the filter that matched their parent. (That all sounds complicated, but it's just crossing the Ts and dotting the Is to make it act the way you were already expecting it to.)
    - To clear a filter, either click the "reset" button or press your "esc" key. (This triggers the button's "reset" action.) 
    - Using the filter form changes the button's mode to "reset."

### DIFF
--- a/content/source/assets/javascripts/application.js
+++ b/content/source/assets/javascripts/application.js
@@ -11,28 +11,36 @@
 document.addEventListener("turbolinks:load", function() {
     "use strict";
 
-    // In navigation sidebars, hide most sub-lists and reveal any that contain the
+    // SIDEBAR STUFF:
+    // - "subNavs" are <li> elements with a nested <ul> as a direct child.
+    // - The <a> child is the "header" of the subnav, and the <ul> is its "content."
+    // - Subnavs are collapsed (<ul> hidden) or expanded (<ul> visible).
+    // - Collapse/expand is managed by the "active" class on the <li>.
+
+    // Collapse most subnavs, but reveal any that contain the
     // current page. The a.current-page class is added during build by
     // layouts/inner.erb.
     var docsSidebar = $("#docs-sidebar ul.nav.docs-sidenav");
     var subNavs = docsSidebar.find("ul").addClass("nav-hidden").parent("li");
         // we leave the nav-hidden class alone after this.
-    var sidebarReset = function() {
+    var resetActiveSubnavs = function() {
         subNavs.removeClass("active");
         // Activate current page, locked-open navs, and all their parents:
         docsSidebar.find("li").has(".current-page, .nav-visible").addClass("active");
     };
-    sidebarReset();
+    resetActiveSubnavs();
 
-    // Make sidebar navs expandable
+    // CSS class that adds toggle controls:
     subNavs.addClass("has-subnav");
+    // Toggle subnav expansion when clicking an area that isn't claimed by the
+    // header or content (usually the :before pseudo-element)
     subNavs.on("click", function(e) {
         if (e.target == this) {
             $(this).toggleClass("active");
         }
         e.stopPropagation();
     });
-    // For subnavs that don't link to a page, use the whole header as a toggle.
+    // If the subnav header doesn't link to a different page, use it as a toggle.
     docsSidebar.find("a[href^='#']").on("click", function(e) {
         e.preventDefault();
         $(this).parent("li").trigger("click");
@@ -68,7 +76,7 @@ document.addEventListener("turbolinks:load", function() {
                 filterField.val("");
                 filterField.trigger("blur");
                 sidebarLinks.parent("li").show();
-                sidebarReset();
+                resetActiveSubnavs();
                 $(this).html("Expand all");
             },
             "click": function(e) {
@@ -82,7 +90,11 @@ document.addEventListener("turbolinks:load", function() {
             }
         });
 
-        // Filter as you type:
+        // Filter as you type. This alters three things:
+        // - "active" class on subnavs
+        // - direct show/hide of <li>s
+        // - state of subnavToggle button
+        // We rely on subnavToggle's "reset" event to clean up when done.
         filterField.on('keyup', function(e) {
             if (e.keyCode === 27) { // escape key
                 subnavToggle.trigger("reset");
--- a/content/source/assets/javascripts/application.js
+++ b/content/source/assets/javascripts/application.js
@@ -53,8 +53,8 @@ document.addEventListener("turbolinks:load", function() {
         if ($("#sidebar-controls").length === 0) { // then add it!
             var sidebarControlsHTML =
                 '<div id="sidebar-controls">' +
-                    '<label for="sidebar-filter-field">Filter page titles</label>' +
-                    '<input type="text" id="sidebar-filter-field" name="sidebar-filter-field" />' +
+                    '<span class="glyphicon glyphicon-search"></span>' +
+                    '<input type="search" id="sidebar-filter-field" class="form-control" name="sidebar-filter-field" role="search" placeholder="Filter page titles" />' +
                     '<a href="#" class="subnav-toggle btn btn-default" role="button">Expand all</a>' +
                 '</div>';
             if ($("#docs-sidebar a.subnav-toggle").length === 0) { // ...in default location
--- a/content/source/assets/javascripts/application.js
+++ b/content/source/assets/javascripts/application.js
@@ -46,10 +46,10 @@ document.addEventListener("turbolinks:load", function() {
     });
 
 
-    // On docs/content pages, add a hierarchical quick nav menu if there are any
-    // H2/H3/H4 headers.
+    // On docs/content pages, add a hierarchical quick nav menu if there are
+    // more than two H2/H3/H4 headers.
     var headers = $('#inner').find('h2, h3, h4');
-    if (headers.length > 0) {
+    if (headers.length > 2) {
         // Build the quick-nav HTML:
         $("#inner #inner-quicknav").html(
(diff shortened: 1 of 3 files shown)


armchairlinguist:
This looks pretty smooth, but I'm worried it's a little much (size-wise and regarding `/` appropriation. Is this intended to solve some of the issues raised for the AWS provider, mainly (besides the addon in general function)?

nfagerlund:
@armchairlinguist About the scope... I asked @bflad about that when I was giving him a demo, and he argued that all of the providers except the small ones would benefit from it. (Where small = datadog, middlin = vsphere, big = gcp/oracle/aws/et. al) 

And I was sort of tempted to leave it at providers, except then when I was testing it in the TFE docs I realized I was already finding it really helpful for getting around, especially in API stuff. 

About the size: Yeah, probably wouldn't hurt to make it smaller. 

About the slash: That's probably the most controversial part. Very willing to yank it depending on the consensus, but like I said, it's also everywhere these days.

nfagerlund:
How's this for size? Less obtrusive?

image

content/source/assets/javascripts/application.js
```diff
@@ -31,28 +37,94 @@ document.addEventListener("turbolinks:load", function() {
...
-            subNavs.removeClass("active");
+
+    // If this is a Very Large Sidebar, add extra controls to expand/collapse
+    // and filter it.
+    var sidebarLinks = docsSidebar.find("a");
+    if (sidebarLinks.length > 30) {
+        if ($("#sidebar-controls").length === 0) { // then add it!
+            var sidebarControlsHTML =
+                '<div id="sidebar-controls">' +
+                    '<label for="sidebar-filter-field">Filter page titles</label>' +
```
bflad: Should this be `Filter navigation` or `Filter sidebar` to be a little bit more clear what its doing? In my testing it seems to handle navigation headers just fine. 👍 

```suggestion
                    '<label for="sidebar-filter-field">Filter navigation</label>' +
```

bflad:
Debatable improvement 😄 : I'm not sure if its also worth having light gray text in the initial box along the lines of "Click here or press / to activate". Does help power users 👍 but make it look a less clean, especially being wordy. 👎 Could/should it be automatically focused when present on the page?

nfagerlund:
@bflad Automatic focus is a no-go; that'd get in the way of arrowkey/spacebar scrolling. Placeholder, I was considering but hadn't decided yet.

apparentlymart:
I was loathe to wade in to the details here given how subjective this all is and thus how much bikeshed potential there is -- I'm in favor of doing this regardless of the details, and don't intend this comment to be a blocker for it -- but since we've started:

I was teetering on suggesting just putting the "Filter page titles" text as the placeholder in the box, and discarding the label altogether, in the interests of compactness. I think if combined with a conventional magnifying-glass icon in the box (probably along with [`role="search"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Search_role) for accessibility) that would be a pretty clear affordance for "search" even once the user has typed text in the box, and thus the persistent text caption won't be needed.

I think with the button to the right of the text box it feels like it's a search action in terms of visual hierarchy, in spite of the caption text. The old treatment of having it as a link may make it clearer that it's only tangentially related to the text field next to it, and maybe removing the persistent "Filter page titles" recovers enough vertical space to justify putting that link back underneath the box to make it even clearer that the "Expand all" action is independent of the filtering.

I don't think it's _super_ important to make the `/` keyboard shortcut visible in the UI. As Nick pointed out, it's become quite a popular power-user convention so some people will discover it automatically while others can get their using pointer-navigation. The de-facto convention seems to be to also support '?' as a shortcut to show the shortcut help, though I don't think we have any other keyboard shortcuts right now so feels like overkill to implement any special UI for discovering something that is just a convenience rather than strictly necessary for usage.

As I say: I don't intend any of this to be blockers and don't feel strongly enough about these to argue them. Just wanted to raise them in case it's helpful. I'm in favor of this change regardless of these details.

content/source/assets/javascripts/application.js
```diff
@@ -11,48 +11,132 @@
...
+        // Type slash to focus sidebar filter:
+        $(document).keydown(function(e) {
+            var focusedElementType = $(document.activeElement).get(0).tagName.toLowerCase();
+            var inputting = focusedElementType === "textarea" || focusedElementType === "input";
+
+            // / (forward slash) key = search
+            if (!inputting && e.keyCode === 191) {
+                e.preventDefault();
+                filterField.focus();
+            }
```
jescalan: So we will be allocating a few new variables every keystroke with this logic, it might be worth adding an earlier bailout, like `if (e.keyCode !== 191) return` at the top to filter out the majority of these situations?

jescalan: Oh also another thing that might be worth doing is scoping this only to pages that contain the sidebar. I assume you will get a javascript error on pages that don't (like the homepage) in its current state.

content/source/assets/javascripts/application.js
```diff
@@ -11,48 +11,132 @@
...
     // layouts/inner.erb.
     var docsSidebar = $("#docs-sidebar ul.nav.docs-sidenav");
-    docsSidebar.find("li").has(".current-page, .nav-visible").addClass("active");
     var subNavs = docsSidebar.find("ul").addClass("nav-hidden").parent("li");
+        // we leave the nav-hidden class alone after this.
+    var resetActiveSubnavs = function() {
+        subNavs.removeClass("active");
+        // Activate current page, locked-open navs, and all their parents:
+        docsSidebar.find("li").has(".current-page, .nav-visible").addClass("active");
+    };
```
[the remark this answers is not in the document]
nfagerlund: Reflex/uncertainty! My grasp of local variable scope is excellent, but I always forget what the rules are for named function statements, and didn't feel like looking it up while I was in the thick of it. If you tell me it's fine I'll switch it.

content/source/assets/javascripts/application.js
nfagerlund: Oh hey, bailing early's an excellent idea, thank you. 

I think this ought to be scoped only to sidebar pages — the whole thing's wrapped in an `if` statement, and the condition expands to `$("#docs-sidebar ul.nav.docs-sidenav").find('a')` not being empty. That's definitely empty on the homepage, and going through `ul.nav.docs-sidenav` means it's also empty on pages that use `inner.erb` as their layout but don't provide sidebar content (like /security.html).

nfagerlund: OH, nope, you're right. Turbolinks. \*shakes fist\* It's a handler on `document`, which outlives `<body>`.

content/source/assets/javascripts/application.js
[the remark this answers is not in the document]
jescalan: Not a huge difference but I tend to prefer them for a couple reasons:

- In any error trace, a named function gets the name of the function in the trace
- Named functions are scope hoisted, so they do not need to be defined before being used in the code (utility functions can be at the bottom of the file)

nfagerlund:
@jescalan Thanks for looking this over! About that meta stuff:

### Telemetry

Pinging a GA event sounds like a great idea, but I haven't learned how to do that yet — can you suggest someone who could me the ropes? 

In terms of what we're hunting... how about something like, "state of the reset button when a nav link was clicked?" That would let us compare nav events where the user used the "pure" nav hierarchy to find what they needed (possibly with some manual expand/collapse) vs. events where the user resorted to either expand-all or filtering.

### User expectations and Terraform special-ness

I only have incomplete answers here. `¯\_(ツ)_/¯` 

I just went and counted some stuff:

- The AWS provider docs have 620 docs pages.
- "Next-tier" providers like Google/Azure/Oracle tend to have around 200.
- Consul, in full, has around 200. Vault has around 400.

So from that perspective, it'd be shocking if terraform.io _didn't_ have to resort to features no one else needs yet. But I don't know yet whether users will start expecting it elsewhere. (Once _I_ started using it I kind of wanted it everywhere, but I also use my address bar's memory instead of navigating like a normal person, so there's a personality issue there.)

TBH though, even though I'm enabling it across the site, I was really looking at this as a prototype for providers-in-the-registry rather than a prototype for our docs sites. Provider docs are legitimately different, because our users don't seem to go there to _acquire knowledge_ — instead, they're there to _access memories that aren't stored in their brains._ Intermediate Terraform users already "know" everything in the AWS docs, but they know it in the form of a fuzzy text-fragment that points to the set of resources they probably need; the goal is to get to those pages instantly, so they can "remember" what arguments they need.

I asked Zach whether he thought I was causing any trouble for the future platform's design. He said he was cool with this today, but that he'd rather use a broader search capability (once we have it) to satisfy the use-case I'm after. I agree with that, but I think the whole external memory thing (which hits provider docs hardest but is also salient in some other spots) has interesting implications for how search should behave; for example, prioritize exact page titles, prioritize titles within the current section even higher, return those matches _immediately_ while we're still trying to bring up the rest of the search system, and try to retain some indicator of the hierarchy map when we display those early results (since that spatial cue can help with remembering).

nfagerlund:
A few other variations on organizing the buttons:

Screenshot_2019-04-10 Documentation - Terraform by HashiCorp(1)

armchairlinguist:
I think with the smaller field/button combo, my concerns are outweighed by the benefits. I'm not too sure on over/under vs side-by-side. Your description of the length of some of the provider pages convinced me this is a good idea. :)

PR #752, opened 2019-04-12
Prevent mobile Webkit/Blink from auto-zooming sidebar filter

nfagerlund:
If a form field has a font-size of less than 16px, telephone browsers
will auto-zoom the screen to them on focus. So, that's obnoxious.
But the pointer media query seems like a solid good-enough way to guess
whether we need to guard against that.


nfagerlund:
btw my new name for this effect is "dramatic woodchuck zoom."

PR #756, opened 2019-04-17
Update the update oauth client endpoint attributes

beekus:
We removed the ability to edit the following attributes for an already existing oauth client:
* api-url
* http-url
* service-provider

See https://github.com/hashicorp/atlas/pull/7506


armchairlinguist:
@beekus Why? I thought the ability to edit those was intentionally added.

armchairlinguist:
I left a comment on the other PR for now. I think this change should wait a final conclusion there.

PR #758, opened 2019-04-19
Add speculative enabled attribute to workspaces API

beekus:
Docs for https://github.com/hashicorp/atlas/pull/7511

Adds a new attribute, `speculative-enabled` on workspaces.


nfagerlund:
@beekus Hey, I'm cleaning up older PRs and found this one. Can you fill me in about the use case for disabling speculative-enabled? If so, I can do the cleanup and merge this.

beekus:
Hey! There is a pretty good description in the linked Atlas PR, but from what I remember it is a security concern. Any user with the ability to create a PR could cause `terraform plan` to be executed on their behalf in Terraform cloud.

PR #766, opened 2019-05-01
CSS: Fix alignment of nav sidebar chevrons/bullets on Safari

nfagerlund:
Some wacky inherited behavior was making it really difficult to get these
bullets aligned right cross-browser, ensuring that they'd look wack on either
Safari or Firefox. I think I've found the weirdnesses we need to reset, so we
should be good now.

**Safari:** 

Screen Shot 2019-04-30 at 4 55 39 PM

**Firefox:**

Screen Shot 2019-04-30 at 4 56 06 PM


nfagerlund:
Closing in favor of #814, which cherry-picks these changes.

PR #773, opened 2019-05-08
Show correct commands for reloading Replicated settings

glasner:
Got myself into a mess with git, so had to start a new PR for https://github.com/hashicorp/terraform-website/pull/735 @armchairlinguist :) 

... because rerunning installer does not in fact reload the settings. More info:

https://app.asana.com/0/331272348117180/1112824705991543

### DIFF
--- a/content/source/docs/enterprise/private/automating-the-installer.html.md
+++ b/content/source/docs/enterprise/private/automating-the-installer.html.md
@@ -301,10 +301,19 @@ If the installation script (`install.sh`) exits successfully, but the Replicated
 * Verify the locations and permissions of the files. The Replicated configuration file should be placed in `/etc/replicated.conf`, and the application settings file should be placed in the path specified in the Replicated configuration file. The permissions of both files should be `600` if owned by the `replicated` user, or `644`.
 * Run a JSON validator on the files to check their validity.
 
-After resolving any issues with file validity or permissions, re-run the installer to verify that the configuration is applied.
+After resolving any issues with file validity or permissions, run the following commands to reload your configuration:
+
+``` bash
+sudo systemctl stop replicated replicated-ui replicated-operator 
+sudo rm -rf /var/lib/replicated 
+sudo systemctl start replicated replicated-ui replicated-operator
+```
+
 
 ## References
 
 - [Replicated installer flags](https://help.replicated.com/docs/distributing-an-application/installing-via-script/#flags)
 - [`/etc/replicated.conf`](https://help.replicated.com/docs/kb/developer-resources/automate-install/#configure-replicated-automatically)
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/private/automating-the-installer.html.md
```diff
@@ -301,10 +301,19 @@ If the installation script (`install.sh`) exits successfully, but the Replicated
...
+sudo systemctl start replicated replicated-ui replicated-operator
+```
+
 
 ## References
 
 - [Replicated installer flags](https://help.replicated.com/docs/distributing-an-application/installing-via-script/#flags)
 - [`/etc/replicated.conf`](https://help.replicated.com/docs/kb/developer-resources/automate-install/#configure-replicated-automatically)
 - [application settings](https://help.replicated.com/docs/kb/developer-resources/automate-install/#configure-app-settings-automatically)
+- [`app-config import` flags](https://help.replicated.com/api/replicatedctl/replicatedctl_app-config_import/)
```
armchairlinguist: It doesn't make much sense to have this if the command isn't being used, right?

content/source/docs/enterprise/private/automating-the-installer.html.md
```diff
@@ -301,10 +301,19 @@ If the installation script (`install.sh`) exits successfully, but the Replicated
...
 
-After resolving any issues with file validity or permissions, re-run the installer to verify that the configuration is applied.
+After resolving any issues with file validity or permissions, run the following commands to reload your configuration:
+
+``` bash
+sudo systemctl stop replicated replicated-ui replicated-operator 
+sudo rm -rf /var/lib/replicated 
+sudo systemctl start replicated replicated-ui replicated-operator
+```
+
```
armchairlinguist: Conceptual yikes, and can you confirm that we've tested this and verified it works?

I think the literal commands are fine for a first pass, but we've been looking at adding our own guide to restarting Replicated, because these commands don't work on every distro, so this is inclining me further to thinking we should make that a part of the public docs.

glasner: Yeah, if you're already writing it; I would agree on making it public. Hopefully, we can get a `ptfe restart` command in v5 to remedy this in the future

koikonom: The commands will work on any systemd-compatible distro (rhel/ubuntu/whatever). We should make it explicit though that this is ONLY for new installations.

Running `rm -rf /var/lib/replicated` on an existing installation that we care about is *not* a good idea.

armchairlinguist:
@glasner Any chance of getting this into the docs soon?

nfagerlund:
I'm tying up loose ends before the great tfe / cloud divide, so I added the suggested changes and am merging now.

PR #784, opened 2019-05-16
glossary: Add deposed (resource) definition

bflad:
Reference: https://github.com/hashicorp/terraform/issues/10753

### DIFF
--- a/content/source/docs/glossary.html.md
+++ b/content/source/docs/glossary.html.md
@@ -270,7 +270,7 @@ Data sources are implemented by [providers][].
 [deposed resource]: glossary.html#deposed
 [deposed resources]: glossary.html#deposed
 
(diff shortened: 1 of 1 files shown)


content/source/docs/glossary.html.md
```diff
@@ -263,6 +264,16 @@ Data sources are implemented by [providers][].
 
 - [Terraform docs: Data Sources](/docs/configuration/data-sources.html)
 
+## Deposed
+
+[deposed]: glossary.html#deposed
+[deposed resource]: glossary.html#deposed
+[deposed resources]: glossary.html#deposed
+
+This status tracks a [resource][] that was marked for deletion from a previous replacement and is no longer used in references, but still remains in the Terraform [state][] and infrastructure due an error from that previous [apply][]. Terraform expected to replace the existing resource by creating a new resource then destroying the existing resource, but the error occurred before the destruction. Terraform will destroy the existing (now previous) resource next apply. This only can occur in resource [configurations][] that include the `lifecycle` configuration block `create_before_destroy` [argument][] set to `true`.
```
armchairlinguist:
```suggestion
This status tracks a [resource][] that was marked for deletion, but still remains in the Terraform [state][] and infrastructure due an error from a previous [apply][]. Terraform expected to replace the existing resource by creating a new resource, then destroying the existing resource, but an error occurred in the apply before the destruction. Existing references to the resource refer to the new resource. Terraform will destroy the `deposed` resource on the next apply. This only can occur in resource [configurations][] that have the `lifecycle` configuration block `create_before_destroy` [argument][] set to `true`.
```

bflad: Great suggestion, applying! 😄

PR #797, opened 2019-05-23
document our max API page size

armchairlinguist:


armchairlinguist:
I did, but let's have someone review and double-check my work :)

PR #798, opened 2019-05-24
document the force execute endpoint in the API

armchairlinguist:


content/source/docs/enterprise/api/run.html.md
```diff
@@ -533,6 +533,51 @@ curl \
...
+
+Parameter | Description
+----------|------------
+`run_id`  | The run ID to execute
+
+The `force-execute` action is analogous to executing the latest queued run after discarding or canceling all prior runs that are not already complete. It initiates the same actions as the "Run this plan now" button at the top of the view of a pending run.
+
+This endpoint enforces the following prerequisites:
+
+* the target run is in the Pending state
```
ryanuber: Should each of these sentences be upper-cased? Actually not sure but that would feel right to me.

Also in casing land, `Pending` isn't the actual name of the state, and these are API docs, so I'd suggest using the real state name of `pending`.

content/source/docs/enterprise/api/run.html.md
```diff
@@ -533,6 +533,51 @@ curl \
...
+* the workspace is locked by another run
+* the run locking the workspace can be discarded 
+
+Runs can currently be discarded if they are in the Planned, Policy Checked, or Policy Override states.
```
ryanuber: Should we remove `currently`? I think that's the assumption when reading API docs. Saying "currently" explicitly makes me feel like this is about to change.

content/source/docs/enterprise/api/run.html.md
```diff
@@ -533,6 +533,51 @@ curl \
...
+
+This endpoint represents an action as opposed to a resource. As such, it does not return any object in the response body.
+
+-> **Note:** This endpoint cannot be accessed with [organization tokens](../users-teams-organizations/service-accounts.html#organization-service-accounts). You must access it with a [user token](../users-teams-organizations/users.html#api-tokens) or [team token](../users-teams-organizations/service-accounts.html#team-service-accounts).
+
+~> **Warning:** This endpoint, while safer than a force-cancel, is not intended for regular use. If you find yourself using it frequently, please reach out to HashiCorp Support for help in developing an alternative approach.
```
ryanuber: I'd probably omit the safety comparison to force-cancel. I'm not exactly sure how it is or _if_ it is safer. I think it depends a lot on context.

chrisarcand: I agree; while it's 'safer' in the sense that cancelled runs really are just static and discardable and haven't taken affect, whereas force cancelling an currently running run is....much more error prone and ⚠️, the context is inherently different. With force canceling you're pulling a potentially unsafe escape lever, where as here there's probably nothing unsafe about executing the run in this state, you're just choosing to flagrantly jump over dealing with all previous runs that are pending and ready, but discardable.

But as we chatted about earlier, a warning to not use this _as your normal workflow_ is definitely reasonable, if you want to. How about....

> While useful at times, force executing circumvents the typical workflow of applying runs using Terraform Enterprise. It is not intended for regular use. If you find yourself using it frequently, please reach out to HashiCorp Support.

Thoughts?

content/source/docs/enterprise/api/run.html.md
armchairlinguist: I went ahead and removed this - I definitely see your concern in keeping it updated. I do have some reservations that not explaining it leaves us open for the question "what is a discardable state?" But we can always add something about it later, if the question is frequent.

armchairlinguist:
@ryanuber @chrisarcand Thank you both for the great feedback! I've updated accordingly, and would love a second look when you have a moment.

PR #803, opened 2019-05-27
Doc sentinel module iteration

rberlind:
The original example functions that showed how to iterate over modules in the tfplan, tfconfig, and tfstate imports were inadequate in that they did not include the name of the resources or the modules in which they were found in the lists they returned.  I have defined generic, parameterized functions, `find_resources_from_plan`, `find_resources_from_config`, and `find_resources_from_state` that accept the resource type as an argument.

These functions can be used unchanged in all policies that use the respective TFE Sentinel imports. Instead of returning a list of maps, they return a map of maps in which the keys are the module path and the values are maps containing all resources of the specfied type in that module. This allows policies that call the functions to output the module name and full resource path in calls to the Sentinel `print()` function, making the outputs from TFE Sentinel policies much clearer to users when their plans violate Sentinel policies.

Additionally, I've defined an auxillary function, `get_address`, that returns the address of a resource using the form `module.A.module.B.<type>.<name>` that is used in plan and apply logs. I also show how to get the full address of a resource instance by appending something like `"[" + string(index) + "]"`. This function can be used with all 3 of the above functions.

You can see examples of the `find_resources_from_plan`  function being used in this [policy](https://github.com/rberlind/new-policies/blob/master/aws/restrict-ec2-instance-type.sentinel) and others in the same repository. The versions of the functions in that repository have additional comments, but I left them out of the text in the 3 updated pages since the surrounding text explains the functions, making the comments unnecessary in them.  In the near future, I plan to merge the policies in that repository into the official [hashicorp/teraform-guides/governance](https://github.com/hashicorp/terraform-guides/tree/master/governance) repository of official Sentinel examples.

### DIFF
--- a/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
@@ -297,7 +297,8 @@ resouce_maps = find_resources_from_config("azurerm_lb")
 ```
 You would then iterate over the `resource_maps` variable to get the maps of the
 specified resource in each module and iterate over the resources in those maps to
-access their attributes.
+access their attributes. Remember that this function will only locate modules (and
+hence resources) in the current configuration.
 
 You could define a similar `find_data_sources_from_config` function to find all data
 sources of a particular type from all modules by simply changing `resources[type]`
--- a/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
+++ b/content/source/docs/enterprise/sentinel/import/tfconfig.html.md
@@ -269,26 +269,74 @@ main = rule { tfconfig.module_paths contains ["foo"] }
 
 #### Iterating through modules
 
-Iterating through all modules to find particular resources can be useful. This
-example shows how to use `module_paths` with the [`module()`
-function](#root-function-module) to retrieve all resources of a particular type from
-all modules (in this case, the [`azurerm_virtual_machine`][ref-tf-azurerm-vm]
-resource). Note the use of `else []` in case some modules don't have any
-resources; this is necessary to avoid the function returning undefined.
-
-[ref-tf-azurerm-vm]: /docs/providers/azurerm/r/virtual_machine.html
+Iterating through all modules to find particular resources can be useful. The
+following example function shows how to use `module_paths` with the
+[`module()`](#root-function-module) function to retrieve all resources of a
+particular type from all modules.
 
 ```python
-import "tfconfig"
-
-get_vms = func() {
-	vms = []
-	for tfconfig.module_paths as path {
-		vms += values(tfconfig.module(path).resources.azurerm_virtual_machine) else []
-	}
-	return vms
+find_resources_from_config = func(type) {
+  resource_maps = {}
+  for tfconfig.module_paths else []  as path {
+    resource_maps[path] = tfconfig.module(path).resources[type] else {}
+  }
+  return resource_maps
 }
 ```
+
+This function creates a map for which the keys are the module paths and the values
+are maps of all resources of the specified type within the modules. Note the use of
+`else {}` in case some modules don't have any resources of the specified type; this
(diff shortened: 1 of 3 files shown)


rberlind:
I removed a lot of the extra details about iterating over modules, leaving in my new functions, but removing additional information on how to use them and the get_address function that I had added to show how to convert module path to standard resource address.

content/source/docs/enterprise/sentinel/import/tfconfig.html.md
```diff
@@ -132,8 +132,8 @@ import "tfconfig"
 filename_value = tfconfig.resources.local_file.accounts.config.filename
 
 main = rule {
-	filename_value contains "${var.domain}" and
-	filename_value contains "${var.subdomain}"
+  filename_value contains "${var.domain}" and
```
vancluever: Can you restore the tabs here? `sentinel fmt` currently uses tabs and that's the way examples should stay until this changes.

This should be the same way for all of the other examples as well.

content/source/docs/enterprise/sentinel/import/tfconfig.html.md
```diff
@@ -269,27 +269,32 @@ main = rule { tfconfig.module_paths contains ["foo"] }
...
-get_vms = func() {
-	vms = []
-	for tfconfig.module_paths as path {
-		vms += values(tfconfig.module(path).resources.azurerm_virtual_machine) else []
-	}
-	return vms
+find_resources_from_config = func(type) {
+  resource_maps = {}
+  for tfconfig.module_paths else []  as path {
+    resource_maps[path] = tfconfig.module(path).resources[type] else {}
```
vancluever: Recommended to use `strings.join` on the path here so that the key is a string and not a list.

content/source/docs/enterprise/sentinel/import/tfconfig.html.md
rberlind: I'll fix the formatting.

content/source/docs/enterprise/sentinel/import/tfconfig.html.md
rberlind: The facts that a module path like `path` is a list and that I do want to eventually print out the complete address of each resource or resource instance that violates a policy is why I originally included the second function for each page, either `get_instance_address` for tfstate and tfplan or `get_address` for tfconfig.

Note that while https://docs.hashicorp.com/sentinel/language/spec#map-literals does say that keys of maps should only be boolean, numeric, or string, my function has been working correctly.

Building the correct path is not as simple as using a single join since we need to treat the root module differently than other modules. We could build the module portion of resource addresses inside the find_resources_from_plan function like this, which I just tested:
```
find_resources_from_plan = func(type) {
  resource_maps = {}
  for tfplan.module_paths as path {
    if length(path) == 0 {
      joined_path = ""
    } else {
      joined_path = "module." + strings.join(path, ".module.") + "."
    }
    resource_maps[joined_path] = tfplan.module(path).resources[type] else {}
  }
```

This lets me write my get_instance_address function more simply than before as:
```
get_instance_address = func(module_path, type, name, index) {
  address = module_path + type + "." + name + "[" + string(index) + "]"
  return address
}
```
Note that I included the final "." at the end of joined_path in the first function so I would not need to use an if/else condition in the second function.  My functions adhere to the specification given in https://www.terraform.io/docs/internals/resource-addressing.html since this is what customers are used to seeing in plan and apply logs.  Were you suggesting that this is changing in Terraform 0.12? I don't understand your inclusion of `[data.]` in your suggested format.

If you prefer the above find_resources_from_plan over my original one, I will make the change and ensure it is correctly formatted.  Of course, this would all be easier if Sentinel just gave the joined path itself as you suggested it might in a future helper function.  I would hope that function would give resource addresses the same way that plan and apply logs do.

PR #809, opened 2019-06-04
Removing nonexistent link to fix CI tests

sudomateo:
Seems like this link doesn't exist in the official documentation as well: https://www.terraform.io/docs/providers/alicloud/r/forward.html

Removing it fix CI tests.


nfagerlund:
Oh hey, thank you, I wondered where that was coming from.

PR #812, opened 2019-06-06
Remove Atlas and legacy migration docs

alisdair:
The Atlas migration functionality has been removed for some time now, so these docs should no longer be needed.

- Remove out of date migration API docs
- Remove Terraform Enterprise v1 (Atlas) docs
- Remove references to the above deleted docs
- Fix some old API responses (ingress-trigger-attributes is no more)


alisdair:
Not yet…

alisdair:
Maybe this is okay after all? @rkst approves the idea at least.

alisdair:
[Here's a PR for Terraform](https://github.com/hashicorp/terraform/pull/21648) removing links to now-removed docs.

There's also [a broken link from terraform-provider-atlas](https://github.com/terraform-providers/terraform-provider-atlas/blob/9345fd78e9e0d177b7c121505bbebdfff7dec1d3/website/docs/index.html.markdown), but the repository is archived so I don't know what to do about that. Tests are failing. I tried adding a redirect in a71ca65 but it didn't seem to work.

@nfagerlund, any advice?

alisdair:
8c46ece attempts to remove the atlas provider submodule altogether, which seems like a reasonable idea at this point. We'll see if it helps with the test failure, at least…

nfagerlund:
Hmm, I think that'll probably put the build green. Let's see. 

If it goes green I'm pulling the trigger, btw.

alisdair:
Thank you! 😍

PR #814, opened 2019-06-08
Improve the look and design of the newer navigation controls

nfagerlund:
This PR fixes some of the gnarlier look-and-feel issues with a couple of recently added features on terraform.io: the jump-to-section-of-current-page dropdown, and the sidebar filter/expand controls. 

This redesign is based on a design @glasner whipped up. The new controls take up less space, are positioned in a more intuitive spot, and are less obtrusive and flashy. They work basically the same way as before; the only change is that you click the "filter" button to reveal and focus the text field, instead of the text field being present by default. 

I also went to some pains to get the text in the headers and controls to line up across the sidebar and the main content area.

Here are some screenshots, with some web inspector boxes showing how the stuff lines up a little better:

filter open

mismatch controls

empty align

headers align

controls align

### DIFF
--- a/content/source/assets/stylesheets/_docs.scss
+++ b/content/source/assets/stylesheets/_docs.scss
@@ -2,6 +2,19 @@
   margin-bottom: 30px;
   overflow: hidden;
 
+  // Re-using the .nav class name on lists means we inherit some
+  // unfortunate behavior from Bootstrap, so reset all that.
+  .nav {
+    ul, li, a {
+      float: none;
+      position: static;
+      padding: 0;
+    }
+    ul { display: block; }
+    li { display: list-item; }
+    a { display: inline; }
+  }
+
   // hide an outdated implementation of expand/collapse link
   a.subnav-toggle {
     display: none;
@@ -27,8 +40,6 @@
     color: $sidebar-link-color;
     font-size: $sidebar-font-size;
     overflow-wrap: break-word;
-    margin: 0 0 0 15px;
-    padding: 0 0 11px 0;
 
     &:focus,
     &:hover {
@@ -97,18 +108,25 @@
     padding-top: 10px;
 
     li {
+      padding: 0 0 11px 0;
+
       a {
         @extend %sidebar-link;
+        padding-left: 6px;
+        // Hanging indent for long sidebar links:
+        // * like   (<-yes)   * like
+        //   this   (no ->)   this
+        display: table-cell;
       }
 
       &:before {
         color: $sidebar-link-color-active;
         content: '\2022';
         font-size: 1em;
-        left: -3px;
-        top: -3px;
         opacity: 0.4;
-        position: absolute;
+        position: relative;
+        // Hanging indent for long sidebar links:
+        display: table-cell;
       }
 
       &.has-subnav {
@@ -119,8 +137,7 @@
         }
         &.active:before {
           transform: rotate(90deg);
-          left: -1px;
-          top: -1px;
+          top: 2px;
         }
       }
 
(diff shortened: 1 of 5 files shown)


mitchellh:
Nice yeah that looks good!

content/source/assets/stylesheets/_docs.scss
```diff
@@ -61,18 +108,21 @@
...
       &:before {
         color: $sidebar-link-color-active;
         content: '\2022';
         font-size: 1em;
-        left: -3px;
-        top: -3px;
         opacity: 0.4;
-        position: absolute;
+        position: relative;
+        display: table-cell;
```
alisdair: This took me a bit of inspector fiddling to understand. I think we're using `table-cell` here (and on the `a`) to maintain a hanging indent between the custom bullet and the text, for long sidebar page titles. Is that correct?

If so, maybe a comment would be helpful.

content/source/assets/stylesheets/_docs.scss
```diff
@@ -1,26 +1,45 @@
...
   h3,
-  h4,
-  h5,
-  h6 {
-    margin-top: 10px;
+  h4 {
+    min-height: 35px;
+    margin-top: 30px;
+    display: flex;
+    align-items: center;
```
alisdair: I can't see why we need flex here. Is there some page with multiple children in the `h4`?

content/source/assets/stylesheets/_docs.scss
nfagerlund: That's exactly right, and I definitely should have left a comment there. Thanks for fiddling, I did this one commit in April and it probably would have taken me just as much fiddling to remember why I did that.

content/source/assets/stylesheets/_docs.scss
nfagerlund: Nope — this is a workaround to get the H1 in the main area and the H4 in the sidebar aligned Just So. I'm open to doing something else, but if I'm remembering the eighty things I tried on Friday correctly, it goes something like:

- I need to line up those two headers _at their text baselines,_ and I also need to line up the controls beneath them. I can't use grid or anything, so I have to do px math and make sure it adds up to the same things.
- Lining up the content boxes of the controls works fine because their texts use the same font metrics... but the headers look way off if I do that.
- CSS doesn't give me any way to specify where the text baseline should be for a line box inside a block box... I think. 
- Aligning the header boxes slightly askew ends up making it harder to align the controls beneath. 
- Well, flexbox actually lets me center things vertically, or say "text starts at the bottom of the box, not the top". And if I put the boxes _like so,_ the baselines look like they line up if I center the text in the H4 box and flex-end the text in the H1 box. 

I recognize that's kind of a crime, but I was starting to run out of ways to align things. idk, any ideas?

nfagerlund:
@alisdair One-way transition for that thing is a fantastic idea! Should I also do that on the "docs" drop-down menu in the header?

content/source/assets/stylesheets/_docs.scss
nfagerlund: Oh, and a related question — I just remembered that we have some rare H1 headers that go to 2 lines. Should I leave enough room in the layout for those? That ends up looking like this: 

Screen Shot 2019-06-10 at 2 34 40 PM

Screen Shot 2019-06-10 at 2 35 15 PM

nfagerlund:
Rebased in some fixes for the stuff Alisdair pointed out. I'm gonna go ahead and merge this, and if we want to revisit how to align text baselines across unrelated columns, we can come back to that. I'm also going to hold off on leaving enough room for two-line H1s by default — that's really super unusual, so it's hard for me to justify pushing the page content down that far on the other 95% of pages. We can revisit that later if we end up getting annoyed by it.

PR #818, opened 2019-06-14
restore team membership checkbox content to match current behavior

armchairlinguist:
Most of this is taken from the previous version of the page, with some small adjustments.


armchairlinguist:
@nfagerlund I noticed when I was checking that the screenshot is still accurate that it might make sense to provide a shot of the actual setting in question instead/in addition to the overall screenshot. Thoughts?

PR #819, opened 2019-06-14
update remote apply language to use proper it's/its

maryelizbeth:
change its -> it's in the remote apply language (TF Remote Backend). 

old: its still pending
new: it's still pending 

there is also complementing change proposed for the Terraform CLI: [#21735](https://github.com/hashicorp/terraform/pull/21735)

### DIFF
--- a/content/source/docs/enterprise/run/cli.html.md
+++ b/content/source/docs/enterprise/run/cli.html.md
@@ -126,7 +126,7 @@ Remote applies use the configuration code from the local working directory, but 
 $ terraform apply
 
 Running apply in the remote backend. Output will stream here. Pressing Ctrl-C
-will cancel the remote apply if its still pending. If the apply started it
+will cancel the remote apply if it's still pending. If the apply started it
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/run/cli.html.md
```diff
@@ -126,7 +126,7 @@ Remote applies use the configuration code from the local working directory, but
 $ terraform apply
 
 Running apply in the remote backend. Output will stream here. Pressing Ctrl-C
-will cancel the remote apply if its still pending. If the apply started it
+will cancel the remote apply if it's still pending. If the apply started it
```
armchairlinguist:
```suggestion
will cancel the remote apply if it's still pending. If the apply has started, it
```

maryelizbeth: i'm going to _sadly_ reject the change because the TF PR was merged without this very good and correct suggestion.  BUT i encourage someone else (or me) to make this change soon!

PR #820, opened 2019-06-14
update iteration functions

rberlind:
Here is my new PR that updates functions used to iterate over modules.  In fact, I essentially rewrote much of the TFE Sentinel import/index.html page: (Defining Policies) at https://www.terraform.io/docs/enterprise/sentinel/import/index.html

This page now gives what I consider to be much better functions for iterating over modules when looking for resource instances of a specific type and for giving the complete addresses of the resource instances that violate policies.  It also adopts the approach I have recently adopted of doing most of the work of validating conditions in Sentinel functions intead of directly in rules since this allows the policies to report all violations of the policy.

While Sentinel's short-circuit logic is efficient at evaluating boolean expressions, it is not really useful for the consumer of a Sentinel policy violation, meaning the person who is informed while doing a run in TFE that their Terraform code violated a policy.  Sentinel's short-circuit logic causes rules to only report the first violation.  The person receiving the violation then has to fix the violation and try again.  If they have multiple resources that violate the policy, they have to go through this cycle multiple times.

It is much better to tell the user up front in a single shot about all their violations so they can fix them and then just try their plan once more.  That is what doing the work in functions accomplishes.

I look forward to feedback from the reviewers.

### DIFF
--- a/content/source/docs/enterprise/sentinel/import/index.html.md
+++ b/content/source/docs/enterprise/sentinel/import/index.html.md
@@ -24,100 +24,178 @@ see [Mocking Terraform Sentinel Data](../mock.html).
 
 -> **Note:** Terraform Enterprise does not currently support custom imports.
 
-## Useful Idioms for Terraform Sentinel Policies
+## Useful Functions and Idioms for Terraform Sentinel Policies
 
-Terraform's internal data formats are complex, which means basic Sentinel policies for Terraform are more verbose than basic policies that use simpler data sources.
-
-This will improve in future versions of Terraform and Sentinel; in the meantime, be aware of the following idioms as you start writing policies for Terraform.
+The following functions and idioms will be useful as you start writing Sentinel policies for Terraform.
 
 ### To Find Resources, Iterate over Modules
 
-The most basic Sentinel task for Terraform is to enforce a rule on all resources of a given type. Before you can do that, you need to get a collection of all the relevant resources.
-
-The easiest way to do that is to copy a function like the following into any policies that examine every resource in a configuration:
+The most basic Sentinel task for Terraform is to enforce a rule on all resources of a given type. Before you can do that, you need to get a collection of all the relevant resources. The easiest way to do that is to copy a function like the following into your policies:
 
 ```python
 import "tfplan"
+import "strings"
 
-# Get an array of all resources of the given type (or an empty array).
-get_resources = func(type) {
-	if length(tfplan.module_paths else []) > 0 { # always true in the real tfplan import
-		return get_resources_all_modules(type)
-	} else { # fallback for tests
-		return get_resources_root_only(type)
-	}
-}
-
-get_resources_root_only = func(type) {
-	resources = []
-	named_and_counted_resources = tfplan.resources[type] else {}
-	# Get resource bodies out of nested resource maps, from:
-	# {"name": {"0": {"applied": {...}, "diff": {...} }, "1": {...}}, "name": {...}}
-	# to:
-	# [{"applied": {...}, "diff": {...}}, {"applied": {...}, "diff": {...}}, ...]
-	for named_and_counted_resources as _, instances {
-		for instances as _, body {
-			append(resources, body)
-		}
-	}
-	return resources
-}
-
-get_resources_all_modules = func(type) {
-	resources = []
-	for tfplan.module_paths as path {
-		named_and_counted_resources = tfplan.module(path).resources[type] else {}
-		# Get resource bodies out of nested resource maps, from:
-		# {"name": {"0": {"applied": {...}, "diff": {...} }, "1": {...}}, "name": {...}}
-		# to:
-		# [{"applied": {...}, "diff": {...}}, {"applied": {...}, "diff": {...}}, ...]
-		for named_and_counted_resources as _, instances {
-			for instances as _, body {
-				append(resources, body)
-			}
-		}
-	}
-	return resources
+# Find all resources of a specific type from all modules using the tfplan import
+find_resources_from_plan = func(type) {
+        resource_maps = {}
+        for tfplan.module_paths as path {
+                # Compute joined_path from the module path
+                if length(path) == 0 {
+                        joined_path = ""
+                } else {
+                        joined_path = "module." + strings.join(path, ".module.") + "."
+                }
+                # Append all resources of the specified type from the current module
+                resource_maps[joined_path] = tfplan.module(path).resources[type] else {}
+        }
+        return resource_maps
 }
 ```
 
--> **Note:** This example uses the tfplan import. You can easily substitute tfconfig or tfstate, depending on your needs.
+-> **Note:** This example uses the `tfplan` import to find all resources of the specified type that have pending changes. You can easily substitute `tfconfig` or `tfstate` and change the name of the function, depending on your needs.
 
-Later, use the function to get a collection of resources:
+Later, you can call this function to get all resources of a desired type by passing the type as a string in quotes:
 
 ```python
-aws_instances = get_resources("aws_instance")
+aws_instances = find_resources_from_plan("aws_instance")
 ```
 
 This example function handles several things that are tricky about finding resources:
 
 - It checks every module for resources (including the root module) by looping over the `module_paths` namespace. The top-level `resources` namespace is more convenient, but it only reveals resources from the root module.
-- It unwraps the import's nested data structures, leaving only an array of resource bodies. The value of `tfplan.module(path).resources[type]` is a series of nested maps keyed by resource name and by [`count`](/docs/configuration/resources.html#count), but the name and count are almost never relevant to a policy. Removing them early makes the rest of the policy more readable.
-- It uses `else` expressions to recover from `undefined` values, for modules that don't have any resources of that type.
-- It falls back to the `resources` namespace if the real tfplan import isn't available, to support [testing](https://docs.hashicorp.com/sentinel/writing/testing). Since current versions of Sentinel don't allow you to mock tfplan's `module()` function, it isn't possible to test Sentinel code that accesses non-root modules. However, you can still test the rest of the policy by mocking resource data under the `resources` namespace.
+- It computes a `joined_path` variable from Sentinel's module path structure and uses this as the key for the module-specific maps added to the `resource_maps` map.
+- It sets the values of these module-level maps to `tfplan.module(path).resources[type]` which gives a series of nested maps keyed by resource name and instance [`count`](/docs/configuration/resources.html#count-multiple-resource-instances). When combined with the `joined_path` keys, the resource names and instance counts allow writers of Sentinel policies to print the full [addresses](/docs/internals/resource-addressing.html) of resource instances that violate policies. This makes it easier for users who see violation messages to know exactly which resources they need to modify. (See the `validate_instance_types` function below for an example.)
+- It uses the Sentinel `else` operator to recover from `undefined` values which would occur for modules that don't have any resources of the specified type.
 
--> **Note:** This example is checking whether it's in a test by looking for an empty `module_paths` namespace, which assumes that our organization is omitting that key in our mock data for Sentinel tests. For policies that need to test mocked `module_paths` data for other purposes, you might need to use a different method to check for the real Terraform imports.
+### A Function to Give the Full Addresses of Resources
 
-### To Test Resources, Use `all`/`any` Expressions
+Here is a function which can be used to give the full [address](/docs/internals/resource-addressing.html) of a resource instance that violates a policy.
 
-Once you have a collection of resources, you usually want to test some property of each resource in the collection — for example, the planned final value of a particular resource attribute.
+```python
+# Get the full address of a resource instance including modules, type,
+# name, and index in form module.<A>.module.<B>.<type>.<name>[<index>]
+get_instance_address = func(joined_path, type, name, index) {
+        address = joined_path + type + "." + name + "[" + string(index) + "]"
+        return address
+}
+```
 
-The most concise tool for this is Sentinel's [`all` and `any` expressions](https://docs.hashicorp.com/sentinel/language/boolexpr#any-all-expressions). For example:
+-> **Note:** This example is intended for use with the tfplan and tfstate imports. For printing resource addresses with the tfconfig import, you could use a function called `get_resource_address` which would leave out the `index` argument and set `address` to `joined_path + type + "." + name`. For all three imports, `joined_path` should be formatted as it is in the `find_resources_from_plan` function above.
+
+We show this function being called in the next section.
+
+### To Validate Resources, Use `for` Loops Inside Functions
+
+Once you have a collection of resources instances of a desired type and know how to give resource instance addresses, you usually want to validate that a particular resource attribute meets some condition.
+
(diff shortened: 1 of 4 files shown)


content/source/docs/enterprise/sentinel/import/index.html.md
```diff
@@ -24,100 +24,178 @@ see [Mocking Terraform Sentinel Data](../mock.html).
...
-			for instances as _, body {
-				append(resources, body)
-			}
-		}
-	}
-	return resources
+import "strings"
+
+# Find all resources of a specific type from all modules using the tfplan import
+find_resources_from_plan = func(type) {
```
vancluever: I still would really prefer the logic for `get_instance_address` be rolled into this function so that the pattern that we promote returns the full resource address instead of having to dip into it further.

I get that this differs between imports, so maybe if we provide some sort of "tabbed" code sample block that would be better. @nfagerlund do we have precedent of this happening in other places currently and can we do that here?

content/source/docs/enterprise/sentinel/import/index.html.md
```diff
@@ -24,100 +24,178 @@ see [Mocking Terraform Sentinel Data](../mock.html).
...
 instance_type_allowed = rule {
-	all get_resources("aws_instance") as r {
-		r.applied.instance_type in allowed_types
-	}
+        instance_types_validated is true
+}
+
+# Main rule
+main = rule {
+        instance_type_allowed is true
```
vancluever: `is true` is superfluous here and should be omitted.

content/source/docs/enterprise/sentinel/import/index.html.md
```diff
@@ -24,100 +24,178 @@ see [Mocking Terraform Sentinel Data](../mock.html).
...
 
+# Call the validation function and assign results
+instance_types_validated = validate_instance_types(allowed_types)
+
 # Rule to restrict instance types
 instance_type_allowed = rule {
-	all get_resources("aws_instance") as r {
-		r.applied.instance_type in allowed_types
-	}
+        instance_types_validated is true
```
vancluever: `is true` is superfluous here and should be omitted.

content/source/docs/enterprise/sentinel/import/index.html.md
```diff
@@ -24,100 +24,178 @@ see [Mocking Terraform Sentinel Data](../mock.html).
...
+}
+
+# Main rule
+main = rule {
+        instance_type_allowed is true
+}
+
+```
+
+Note that you could leave out the `instance_type_allowed` rule and place `instance_types_validated is true` directly inside your `main` rule. You could also leave out `is true` in both rules; we included it here to be more explicit.
```
vancluever: 2 issues here:

> Note that you could leave out the `instance_type_allowed` rule and place `instance_types_validated is true` directly inside your `main` rule.

If this is the case, the `instance_type_allowed` rule should just be removed from the example as it serves no extra purpose.

> You could also leave out `is true` in both rules; we included it here to be more explicit.

I get what you are saying here, but "more explicit" is a misnomer - there is no ambiguity in what a standalone unary boolean expression evaluates to.

I think that this paragraph could be removed if we correct the notes in the example:

* Remove the `instance_type_allowed` rule and just evaluate off of the `instance_types_validated` rule.
* Use the unary form of the boolean expression for `instance_types_validated`. This is readable enough and conveys the purpose just fine.

content/source/docs/enterprise/sentinel/import/index.html.md
```diff
@@ -24,100 +24,178 @@ see [Mocking Terraform Sentinel Data](../mock.html).
...
+
+### Validating Multiple Conditions in a Single Policy
+
+If you want a policy to validate multiple conditions against resources of a specific type, we recommend that you use a single function to iterate over all resource instances of that type and make this function return a list of boolean values, using one for each condition.  You can then use multiple Sentinel rules that evaluate those boolean values or evaluate them all in your `main` rule. Here is a partial example:
```
vancluever: I understand _why_ this is necessary in some instances - namely to reduce evaluation time on larger plans by removing the need to scan over all of the data multiple times. I wish it didn't need to be done this way as it circumvents some of Sentinel's tracing functionality.

What I would probably prefer in these instances - to help with readability, is that either we don't rely on a function here (as there is no parameterization) and that results are just set as globals. If this is only designed to be executed once, then let's just execute it once.

* `validate_private_acl_and_kms_encryption` should be inlined and executed as early as possible.
* The `private` and `encrypted_by_kms` global values should be set in the inlined replacement.
* Rules should be created for both conditions so that tracing will reflect the failure class. You can just add an `is_` prefix for each condition (example: `is_private = rule { private }`).

content/source/docs/enterprise/sentinel/import/tfconfig.html.md
```diff
@@ -267,28 +267,14 @@ import "tfconfig"
...
 
 Iterating through all modules to find particular resources can be useful. This
-example shows how to use `module_paths` with the [`module()`
-function](#root-function-module) to retrieve all resources of a particular type from
-all modules (in this case, the [`azurerm_virtual_machine`][ref-tf-azurerm-vm]
-resource). Note the use of `else []` in case some modules don't have any
-resources; this is necessary to avoid the function returning undefined.
+[example](iterate-over-modules) shows how to use `module_paths` with the
+[`module()` function](#function-module-) to find all resources of a
+particular type from all modules using the `tfplan` import. By changing `tfplan` in this function to `tfconfig`, you could make a similar function find all resources of a specific type in the Terraform configuration.
```
vancluever: Please wrap this line to 79 characters to match the rest of the document.

content/source/docs/enterprise/sentinel/import/tfstate.html.md
```diff
@@ -169,31 +169,14 @@ main = rule { tfstate.module_paths contains ["foo"] }
...
 
 Iterating through all modules to find particular resources can be useful. This
-example shows how to use `module_paths` with the [`module()`
-function](#function-module-) to retrieve all resources of a particular type from
-all modules (in this case, the [`azurerm_virtual_machine`][ref-tf-azurerm-vm]
-resource). Note the use of `else []` in case some modules don't have any
-resources; this is necessary to avoid the function returning undefined.
+[example](iterate-over-modules) shows how to use `module_paths` with the
+[`module()` function](#function-module-) to find all resources of a
+particular type from all modules using the `tfplan` import. By changing `tfplan` in this function to `tfstate`, you could make a similar function find all resources of a specific type in the current state.
```
vancluever: Please wrap this line to 79 characters to match the rest of the document.

content/source/docs/enterprise/sentinel/import/index.html.md
rberlind: Getting the full address of resource instances in find_resource_from_plan would require iterating over all the resources which is not necessary at this point.  What I am doing is returning a map of maps, one per module, for which each key is the joined_path of the module.  Iterating over all the resources and their instances inside the find_resources_from_plan function would make it much more complex than it needs to be.

You had previously objected to my using path as the keys of the resource_maps even though the code worked fine. I know that https://docs.hashicorp.com/sentinel/language/spec#map-literals says that keys of maps are supposed to be boolean, numeric, or string, but my code did work fine.  I actually liked the way the code looked better when I did set the key to path and computed the full resource instance address in the get_instance_address function.  While that required rederiving the joined path more often and is therefore less efficient, it did keep all the code for the derivation of the address in a single function.  I would be happy to revert back to that pattern.  I think efficiency of the Sentinel code is less important than clear organization.

If you eventually end up writing standard functions in go that do what my functions are doing, then I would write my policies to use those functions.  But since we currently have to paste the contents of these functions inside the policies, I don't want to make them more complicated than necessary.

content/source/docs/enterprise/sentinel/import/index.html.md
```diff
@@ -24,100 +24,178 @@ see [Mocking Terraform Sentinel Data](../mock.html).
...
+While you could use Sentinel's [`all` and `any` expressions](https://docs.hashicorp.com/sentinel/language/boolexpr#any-all-expressions) directly inside Sentinel rules, your rules would only report the first violation because Sentinel uses short-circuit logic. It is therefore usually preferred to use [`for` loops](https://docs.hashicorp.com/sentinel/language/loops) inside functions called from your rules so that you can report all violations that occur. Another benefit of calling a function from a rule is that the standard Sentinel output is streamlined allowing users to focus on the warnings your print in your policies.
 
-The most concise tool for this is Sentinel's [`all` and `any` expressions](https://docs.hashicorp.com/sentinel/language/boolexpr#any-all-expressions). For example:
+Here is a function that validates that the instance types of all EC2 instances being provisioned are in a given list:
+
+```python
+# Validate that all EC2 instances have instance_type
+# in the allowed_types list
+validate_instance_types = func(allowed_types) {
+
```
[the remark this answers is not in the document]
rberlind: OK

content/source/docs/enterprise/sentinel/import/index.html.md
```diff
@@ -24,100 +24,178 @@ see [Mocking Terraform Sentinel Data](../mock.html).
...
+        validated = true
+
+        aws_instances = find_resources_from_plan("aws_instance")
+
+        # Loop through the resource maps, resources, and instances
+        for aws_instances as joined_path, resource_map {
+                for resource_map as name, instances {
+                        for instances as index, r {
+
```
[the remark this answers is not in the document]
rberlind: OK

content/source/docs/enterprise/sentinel/import/index.html.md
rberlind: I know it is superfluous from a code point of view, but it makes things clearer to people who are not already familiar with Sentinel or might not themselves be programmers.  Some of the people being asked to create Sentinel policies at our customers are actually not programmers.  Including `is true` does not hurt.

content/source/docs/enterprise/sentinel/import/index.html.md
rberlind: Same response as previous comment.

content/source/docs/enterprise/sentinel/import/index.html.md
rberlind: I'll remove the rule and just use a main rule.  I actually mostly do that in most of my policies now anyways to reduce verbose Sentinel logging.

content/source/docs/enterprise/sentinel/import/index.html.md
rberlind: I agree that the function in this case and in some of my new second-generation policies under https://github.com/hashicorp/terraform-guides/tree/master/governance/second-generation are not parameterized and are only invoked once. However, some of those policies do have functions that are parameterized, particularly when they accept a list of allowed values. It is true that in those cases, the functions could just read the global list and that the functions could be inlined to the main policy. Also, I have written other policies in which the functions have more parameters and are invoked more than once.

So, I prefer to do all the processing inside functions so that all the policies follow the same pattern.  I don't see any harm in doing this.

Additionally, in so far as users will copy code from one policy to another, using functions makes it clearer which portion of the source policy should be copied.

Frankly, I find the way Sentinel reports evaluation of rules useless to the consumer of the policies by which I mean the person who ran some Terraform code in a TFE workspace and encountered violations.  One ends up with many lines of useless output and line numbers.  Those might be useful to the writer of the policy, but not to the consumer of its violations.  Using a single main rule instead of multiple rules reduces Sentinel's verbose rule processing output.

content/source/docs/enterprise/sentinel/import/tfconfig.html.md
rberlind: ok

content/source/docs/enterprise/sentinel/import/tfstate.html.md
rberlind: ok

content/source/docs/enterprise/sentinel/import/index.html.md
rberlind: but I'll remove to make you happy.

rberlind:
I left comments inline about the various points you raised, @vancluever.
I also made the various formatting changes and pushed a new commit to my branch.

content/source/docs/enterprise/sentinel/import/index.html.md
vancluever: > Iterating over all the resources and their instances inside the find_resources_from_plan function would make it much more complex than it needs to be.

Aside from the extra two loops and a bit of extra details on the `join`, what is more complex about the below example?

```
find_resources_by_type = func(type) {
	result = {}

	for tfplan.module_paths as path {
		for tfplan.module(path).resources[type] else {} as name, by_index {
			for by_index as index, r {
				result[strings.join([module_address(path), type, name, index], ".")] = r
			}
		}
	}

	return result
}
```

(Where `module_address` is a helper function that returns a properly joined module address.)

> I actually liked the way the code looked better when I did set the key to path and computed the full resource instance address in the get_instance_address function. While that required rederiving the joined path more often and is therefore less efficient, it did keep all the code for the derivation of the address in a single function.

I still don't see how this can be useful aside from niche uses.

Returning a map of resources keyed on the full resource path allows a very simple path to reporting a resource in violation - printing the key. It also allows for much easier iteration as you have already done the work of looping on the two inner maps.

As far as I can see, the rest of the example doesn't demonstrate any value in providing less granularity than the final set of actual resources keyed on address - nor do I think it should. In fact, `validate_instance_types` could be greatly simplified if it was operating on a flat set of resources versus having to deeply iterate and combine the resource address itself.

content/source/docs/enterprise/sentinel/import/index.html.md
vancluever: I think I'd be okay with this if the "result generator" function was not returning an ambiguous list. A map would be better:

```
validate_private_acl_and_kms_encryption = func() {
	result = {
		"private":          true,
		"encrypted_by_kms": true,
	}

	s3_buckets = find_resources_from_plan("aws_s3_bucket")
	for s3_buckets as joined_path, resource_map {
		# Set "private" or "encrypted_by_kms" result map values to
		# false here if there is a violation.
	}

	return result
}

validations = validate_private_acl_and_kms_encryption()
is_private = rule { validations["private"] }
is_encrypted_by_kms = rule { validations["encrypted_by_kms"] }

main = rule {
	is_private and is_encrypted_by_kms
}
```



> Frankly, I find the way Sentinel reports evaluation of rules useless to the consumer of the policies by which I mean the person who ran some Terraform code in a TFE workspace and encountered violations. One ends up with many lines of useless output and line numbers. Those might be useful to the writer of the policy, but not to the consumer of its violations. Using a single main rule instead of multiple rules reduces Sentinel's verbose rule processing output.

We are working on correcting the way trace output is shown within a specific TFE run.

Having **clearly defined rules** is a core design pattern for Sentinel, and our documentation should be promoting these patterns wherever possible, regardless of opinions on the current state of their usefulness.

content/source/docs/enterprise/sentinel/import/index.html.md
```diff
@@ -24,100 +24,161 @@ see [Mocking Terraform Sentinel Data](../mock.html).
...
+# Find all resources of a specific type from all modules using the tfplan import
+find_resources_from_plan = func(type) {
+        resource_maps = {}
+        for tfplan.module_paths as path {
+                # Compute joined_path from the module path
+                if length(path) == 0 {
+                        joined_path = ""
+                } else {
+                        joined_path = "module." + strings.join(path, ".module.") + "."
```
vancluever: This line will not print a correct resource address as `path` may have multiple entries, producing something like `module.foo.bar.module.`

A helper would probably be better here after all:

```
module_address = func(path) {
	if length(path) < 1 {
		return ""
	}

	result = []
	for path as p {
		append(result, ["module", "p"]) // Can also be multiple append calls
	}

	return strings.join(result, ".")
}
```

vancluever: Yeah I see that now - sorry about that!

vancluever:
@rberlind I've just updated the line wrapping - with that this PR is good to go and can be merged. I'll let you do the honors!

PR #821, opened 2019-06-17
Update links to terraform-guides/governance

rberlind:
The governance section of the terraform-guides repository was reorganized into first-generation and second-generation policies.  All of the links on this page had to be updated.  Some point to older first-generation policies while others point to new secong-generation policies.


rberlind:
After merging into master, it would be good to also merge into stable-webiste (as described here: https://github.com/hashicorp/terraform-website#more-about-stable-website) so that the broken links are fixed as quickly as possible.

PR #823, opened 2019-06-19
Two small CSS tweaks

nfagerlund:
- Keep the dropdown menus from lying to people.
- Keep the "Expand all" button from wrapping to two lines (while still keeping it from jumping around when it transforms).

### DIFF
--- a/content/source/assets/stylesheets/_docs.scss
+++ b/content/source/assets/stylesheets/_docs.scss
@@ -100,9 +100,12 @@
       font-variant-caps: all-small-caps;
       color: #666;
       display: inline-block;
-      width: 40%;
       border: none;
       background: transparent;
+
+      &:first-child {
+        min-width: 6em;
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


nfagerlund:
Before: Wrapped expand

Screenshot_2019-06-26 Home - Terraform Enterprise - Terraform by HashiCorp

After: Unwrapped expand

Screenshot_2019-06-26 Home - Terraform Enterprise - Terraform by HashiCorp(1)

Before: One of these cursors only tells lies (clickable `<a>` elements are highlighted with web inspector)

IMG_7631

After: One of these cursors only tells the truth

IMG_7632

nfagerlund:
@alisdair That was my first thought as well; these should be bigger targets. I got it sort of working in https://github.com/nfagerlund/terraform-website/commit/9c8db263cd2ad7aa91e3c46d7475b12c524c6cb0, but wow it was brittle -- it relied on that line-height: 44px bit for making sure the text was centered in the click target, and that would blow up if a link ever extended to two lines long (which we're maybe not 100% in control of, because of browser zoom implementations etc.). So I decided not to push it for the moment. 😬 

Anyway, thanks for the review! 🙌🏼  Thank you too @dylanegan!

PR #827, opened 2019-06-25
fixed iterate-over-modules links

rberlind:
When I changed a section header in index.html.md under sentine/import, I broke the iterate-over-modules links in the tfconfig.html.md, tfstate.html.md, and tfplan.html.md files.  This PR fixes those links.


rberlind:
I'm not sure why, but the links are still broken, at least in the Travis builds. I'm currently referencing like this:

`This [example](iterate-over-modules)` with the link being set to:
`[iterate-over-modules]: /docs/enterprise/sentinel/import/index.html#to-find-resources-iterate-over-modules`.

index.html.md has section header: `### Iterate over Modules and Find Resources`

Any ideas on how to fix?

rberlind:
There is also a broken link for the Akamai provider in the push check, but that is not from my PR.

rberlind:
I see the problem. I have to use [] instead of () for the reference

PR #828, opened 2019-06-25
Update and clarify usage of team tokens

nfagerlund:
Team tokens are the recommended way to do remote runs from CI/CD scripts.
In cases where SAML is enabled, they're the only viable option, since user
tokens go inactive if the user doesn't log in on schedule.


nfagerlund:
Confirmed with @rkst in slack that this description is on target.

PR #829, opened 2019-06-26
Tweak span.circle cursor behaviour

matildepark:
Before this PR, in the `terminal` sections, if you hover your cursor over the circles (mimicking a GUI window) then the cursor is textual and encourages the user highlight the circles, breaking the illusion of a terminal window.

Now it shows the default cursor, better befitting a window — unless you've hovering over the terminal content, where it is a textual cursor, as expected.


nfagerlund:
Ha, nice catch. Thank you!

PR #830, opened 2019-06-26
Update TFE migration docs to use remote backend, finally

nfagerlund:
at long last 😭

### DIFF
--- a/content/source/docs/enterprise/migrate/workspaces.html.md
+++ b/content/source/docs/enterprise/migrate/workspaces.html.md
@@ -18,13 +18,13 @@ These workspaces, managed with the `terraform workspace` command, aren't the sam
 
 If you use multiple workspaces, you'll need to migrate each one to a separate TFE workspace.
 
-Migrating multiple workspaces is similar to [migrating a single workspace](./index.html), but it requires a different attribute in the `terraform { backend ...` block.
+Migrating multiple workspaces is similar to [migrating a single workspace](./index.html), but it requires a slightly different backend configuration.
 
 -> **API:** See the [State Versions API](../api/state-versions.html). Be sure to stop Terraform runs before migrating state to TFE, and only import state into TFE workspaces that have never performed a run.
 
-## Step 1: Ensure Terraform ≥ 0.11.8 is Installed
+## Step 1: Ensure Terraform ≥ 0.11.13 is Installed
 
-To follow these instructions, you need Terraform 0.11.8 or later on the workstation where you are performing the migration. The remote backend this process relies on is not present in older versions.
+To follow these instructions, you need Terraform 0.11.13 or later on the workstation where you are performing the migration. In older versions, [the `remote` backend][remote-backend] is either absent or only partially supported.
 
 ## Step 2: Gather Credentials, Data, and Code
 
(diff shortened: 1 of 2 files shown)


content/source/docs/enterprise/migrate/workspaces.html.md
```diff
@@ -59,94 +55,86 @@ In your shell, go to the directory with the Terraform configuration you're migra
...
-At this point, it's a good idea to use a checklist of workspaces to keep track of your progress.
+  If you answer "yes", Terraform will migrate all states. If you answer
+  "no", Terraform will abort.
 
-## Step 11: Run `terraform init` and Answer "No"
+  Enter a value:
+```
 
-Run `terraform init`.
+Answer "yes."
```
mildwonkey:
```suggestion
Answer "yes".
```

content/source/docs/enterprise/migrate/workspaces.html.md
```diff
@@ -59,94 +55,86 @@ In your shell, go to the directory with the Terraform configuration you're migra
...
 
-When asked if you want to migrate state, **answer "no."**
+Some backends, including the default `local` backend, allow a special `default` workspace that doesn't have a specific name. If you previously used a combination of named workspaces and the special `default` workspace, the prompt will next ask you to choose a new name for the `default` workspace, since TFE doesn't support unnamed workspaces:
```
mildwonkey: Do we have any conventions/suggestions/best practices for the alternate `default` namespace? (This is more about my own curiosity and not a request to make a change).

nfagerlund: TBH, I don't think we know enough about how people are using alternate workspaces in CLI terraform. Part of why the default case was blowing up on migration for so long is that no one here thought it made any sense to keep using `default` once you had any named workspaces. 

If I think in terms of the order things would happen in... probably you'd start a new configuration on default, tweaking and adjusting some new infrastructure you're not using yet. Then you'd get it the way you liked it and start using it as production. Then you'd want to spin up an alternate instance of it to test something dangerous, so you'd make a new workspace. In that scenario, `default` would be prod. But I don't know about other scenarios. I have to just assume that if you're using alternate workspaces, you already had to think it through enough that you probably know what your default workspace is used for.

content/source/docs/enterprise/migrate/workspaces.html.md
```diff
@@ -59,94 +55,86 @@ In your shell, go to the directory with the Terraform configuration you're migra
...
-
-- **"Do you want to copy only your current workspace?"** Answer "yes."
+- Use the `remote` backend.
+- In the `organization` attribute, specify the name of your TFE organization.
+- The `hostname` attribute is only necessary with private TFE instances. You can omit it if you're using the SaaS version of TFE.
+- In the `prefix` attribute in the `workspaces` block, specify a prefix to use when naming newly created TFE workspaces. You'll use short workspace names on the CLI, and long workspace names with the prefix added in TFE's UI and API. See [the remote backend documentation][remote-backend] for details.
 
-    Terraform init can only handle the default workspace, so you'll use a different command for the others.
-- **"Do you want to copy existing state to the new backend?"** Answer "yes."
+## Step 6: Run `terraform init`, Answer "Yes", and Name the Default Workspace
```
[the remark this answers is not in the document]
nfagerlund: Sure!

nfagerlund:
@mildwonkey Thanks for the review! 🙌🏼

PR #832, opened 2019-06-28
Global policy sets no longer contain workspace relationships

chrisarcand:
If a policy set is marked as global, all the organization's workspaces are implicitly related to the set, so we no longer return every one of them in the relationships response.


content/source/docs/enterprise/api/policy-sets.html.md
```diff
@@ -57,7 +57,7 @@ Key path                               | Type            | Default | Description
...
-`data.relationships.workspaces.data[]` | array\[object\] | `[]`    | A list of resource identifier objects that defines which workspaces the new set will be attached to. These objects must contain `id` and `type` properties, and the `type` property must be `workspaces` (e.g. `{ "id": "ws-2HRvNs49EWPjDqT1", "type": "workspaces" }`). Obtain workspace IDs from the [workspace settings](../workspaces/settings.html) or the [Show Workspace](./workspaces.html#show-workspace) endpoint.
+`data.relationships.workspaces.data[]` | array\[object\] | `[]`    | A list of resource identifier objects that defines which workspaces the new set will be attached to. These objects must contain `id` and `type` properties, and the `type` property must be `workspaces` (e.g. `{ "id": "ws-2HRvNs49EWPjDqT1", "type": "workspaces" }`). Obtain workspace IDs from the [workspace settings](../workspaces/settings.html) or the [Show Workspace](./workspaces.html#show-workspace) endpoint. Note: this key is omitted from policy sets marked as global; global policy sets are implicitly related to all of the organization's workspaces.
```
nfagerlund: What happens if I send it when creating a global set? I'm assuming it's ignored, yeah? If so, I suggest,

```suggestion
`data.relationships.workspaces.data[]` | array\[object\] | `[]`    | A list of resource identifier objects that defines which workspaces the new set will be attached to. These objects must contain `id` and `type` properties, and the `type` property must be `workspaces` (e.g. `{ "id": "ws-2HRvNs49EWPjDqT1", "type": "workspaces" }`). Obtain workspace IDs from the [workspace settings](../workspaces/settings.html) or the [Show Workspace](./workspaces.html#show-workspace) endpoint. Note: this key is ignored for policy sets marked as global; global policy sets are implicitly related to all of the organization's workspaces.
```

chrisarcand:
I actually made this edit in the wrong place from what I was intending; I wanted to make a note on the `Show a policy set` doc. I think 'omitted' is a good word here because we omit a key in the requested resource based on a state.

'ignored' does make sense where you suggested it but it isn't ignored - we helpfully validate that you cannot attach workspaces to a global policy set in all locations (attaching them as relationships in a policy-set creation, adding or removing them individually on the relationships endpoints, etc) and return an informative error message.

I've added this point in the policy set creation endpoint, though I'm unsure how you'd like to present that bit of info in the relationships endpoints (or if you even want to bother).

content/source/docs/enterprise/api/policy-sets.html.md
```diff
@@ -57,7 +57,7 @@ Key path                               | Type            | Default | Description
...
-`data.relationships.workspaces.data[]` | array\[object\] | `[]`    | A list of resource identifier objects that defines which workspaces the new set will be attached to. These objects must contain `id` and `type` properties, and the `type` property must be `workspaces` (e.g. `{ "id": "ws-2HRvNs49EWPjDqT1", "type": "workspaces" }`). Obtain workspace IDs from the [workspace settings](../workspaces/settings.html) or the [Show Workspace](./workspaces.html#show-workspace) endpoint.
+`data.relationships.workspaces.data[]` | array\[object\] | `[]`    | A list of resource identifier objects that defines which workspaces the new set will be attached to. These objects must contain `id` and `type` properties, and the `type` property must be `workspaces` (e.g. `{ "id": "ws-2HRvNs49EWPjDqT1", "type": "workspaces" }`). Obtain workspace IDs from the [workspace settings](../workspaces/settings.html) or the [Show Workspace](./workspaces.html#show-workspace) endpoint. Individual Workspaces cannot be attached to the policy set when `data.attributes.global` is `true`.
```
nfagerlund: Capitalization.

```suggestion
`data.relationships.workspaces.data[]` | array\[object\] | `[]`    | A list of resource identifier objects that defines which workspaces the new set will be attached to. These objects must contain `id` and `type` properties, and the `type` property must be `workspaces` (e.g. `{ "id": "ws-2HRvNs49EWPjDqT1", "type": "workspaces" }`). Obtain workspace IDs from the [workspace settings](../workspaces/settings.html) or the [Show Workspace](./workspaces.html#show-workspace) endpoint. Individual workspaces cannot be attached to the policy set when `data.attributes.global` is `true`.
```

content/source/docs/enterprise/api/policy-sets.html.md
```diff
@@ -268,6 +268,8 @@ curl --request GET \
 }
 ```
 
+-> **Note:** The `data.relationships.workspaces` refers to workspaces directly attached to the policy set. This key is omitted for policy sets marked as global; in this case, global policy sets are implicitly related to all of the organization's workspaces.
```
nfagerlund:
```suggestion
-> **Note:** The `data.relationships.workspaces` object refers to workspaces directly attached to the policy set. This key is omitted for policy sets marked as global, which are implicitly related to all of the organization's workspaces.
```

PR #835, opened 2019-07-02
[Terraform Enterprise] Using remote plan CLI when TF working directory is set

vwordsworth:
If merged, these changes add documentation for using the Terraform CLI for remote plans when the Terraform Enterprise workspace has a `Terraform Working Directory` configured.

This distinction is important to call out for TFE users who plan to build tooling around the remote plan functionality and have nested repository structures, as we had initially assumed that the `terraform init` and `terraform plan` were supposed to be run from the subdirectory in which the workspace's configuration files were located.

### DIFF
--- a/content/source/docs/enterprise/run/cli.html.md
+++ b/content/source/docs/enterprise/run/cli.html.md
@@ -81,6 +81,24 @@ rerun this command to reinitialize your working directory. If you forget, other
 commands will detect it and remind you to do so if necessary.
 ```
 
+### Terraform Init and Terraform Working Directory
+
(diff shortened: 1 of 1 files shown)


nfagerlund:
@vwordsworth This was a bug! Introduced in 0.12, fixed in 0.12.5. Running from the nested workdir was always intended to work, but we didn't think all the way through some of the implications, and then a change to something else broke it. 

Writing updated docs now; will close this PR when I can paste a link to new PR.

nfagerlund:
Updated docs are going into some add-on commits in https://github.com/hashicorp/terraform-website/pull/854 (currently has part of the story but not all).

PR #836, opened 2019-07-02
Update API docs and getting started guide

ryanuber:
This PR contains the API documentation changes as well as updates to the hugely outdated getting started guide for Sentinel VCS.

I'm also working on the guides section, which is more involved as it's very intertwined with the flow of other pages, so I decided to open early with just the API docs and getting started, to get feedback on this in the meantime.


content/source/docs/enterprise/api/policies.html.md
```diff
@@ -27,7 +27,9 @@ sidebar_current: "docs-enterprise2-api-policies"
...
 
-This page documents the API endpoints to create, read, update, and delete the Sentinel policies in an organization. To view and manage the results of a specific run's policy check, use the [Runs API](./run.html).
+This page documents the API endpoints to create, read, update, and delete the Sentinel policies in an organization. Use of these endpoints provides a method to manually manage individual policies within Terraform Enterprise. To view and manage the results of a specific run's policy check, use the [Runs API](./run.html).
+
+~> **Note**: Use of these endpoints is not required to use Sentinel features in Terraform Enterprise. Alternatively, you can manage versioned policy sets, which allows integrating with VCS or uploading policy set data and configuration as a whole. Use the [policy sets API](./policy-sets.html) to leverage versioned policy sets.
```
nfagerlund: I wonder if this should be more strongly worded — "These endpoints are no longer the recommended way to use Sentinel features in Terraform Enterprise. We suggest using..."

content/source/docs/enterprise/api/policy-sets.html.md
```diff
@@ -25,7 +25,7 @@ sidebar_current: "docs-enterprise2-api-policy-sets"
 
 [Sentinel Policy as Code](../sentinel/index.html) is an embedded policy as code framework integrated with Terraform Enterprise.
 
-Policy sets are groups of policies that are applied together to related workspaces. By using policy sets, you can group your policies by attributes such as environment or region. Individual policies that are members of policy sets will only be checked for workspaces that the policy set is attached to. In order for a policy to be active and checked during runs, it must be a member of at least one policy set that is attached to workspaces.
+Policy sets are groups of policies that are applied together to related workspaces. By using policy sets, you can group your policies by attributes such as environment or region. Individual policies within a policy set will only be checked for workspaces that the policy set is attached to. Policy sets can group indidual policies created via the [policies API](./policies.html), or act as versioned sets which are sourced from either a version control system (such as GitHub) or uploaded as a whole via the [policy set versions API](#create-a-policy-set-version).
```
nfagerlund:
```suggestion
Policy sets are groups of policies that are applied together to related workspaces. By using policy sets, you can group your policies by attributes such as environment or region. Individual policies within a policy set will only be checked for workspaces that the policy set is attached to. Policy sets can group indidual policies created via the [policies API](./policies.html), or act as versioned sets which are either sourced from a version control system (such as GitHub) or uploaded as a whole via the [policy set versions API](#create-a-policy-set-version).
```

content/source/docs/enterprise/api/policy-sets.html.md
```diff
@@ -50,14 +50,20 @@ This POST endpoint requires a JSON object with the following properties as a req
...
+`data.attributes.vcs-repo.branch`             | string          | `null`  | The branch of the VCS repo. If empty, the VCS provider's default branch value will be used.
+`data.attributes.vcs-repo.identifier`         | string          |         | The identifier of the VCS repository in the format `<namespace>/<repo>`. For example, on GitHub, this would be something like `hashicorp/my-policy-set`.
+`data.attributes.vcs-repo.oauth-token-id`     | string          |         | The OAuth Token ID to use to connect to the VCS host.
+`data.attributes.vcs-repo.ingress-submodules` | boolean         | `false` | Determines whether repository submodules will be instantiated during the clone operation.
+`data.attributes.policies-path`               | string          | `null`  | The sub-path within the attached VCS repository to clone. All files and directories outside of this sub-path will be ignored. This option may only be specified when a VCS repo is present. **Note:** When this option is configured, commits in the repository which are outside of the given policies-path will be ignored.
```
nfagerlund: Hmm. This should be clearer, but let's see whether I can actually manage that:

```suggestion
`data.attributes.policies-path`               | string          | `null`  | The subdirectory of the attached VCS repository that contains the policies for this policy set. Files and directories outside of this sub-path will be ignored, and changes to those unrelated files won't cause the policy set to be updated. This option may only be specified when a VCS repo is present.
```

content/source/docs/enterprise/api/policy-sets.html.md
```diff
@@ -50,14 +50,20 @@ This POST endpoint requires a JSON object with the following properties as a req
...
+`data.relationships.policies.data[]`          | array\[object\] | `[]`    | A list of resource identifier objects that defines which policies will be members of the new set. These objects must contain `id` and `type` properties, and the `type` property must be `policies` (e.g. `{ "id": "pol-u3S5p2Uwk21keu1s", "type": "policies" }`). **Note:** This option will be deprecated in the future in favor of VCS policy sets.
```
nfagerlund:
```suggestion
`data.relationships.policies.data[]`          | array\[object\] | `[]`    | **DEPRECATED.** A list of resource identifier objects that defines which policies will be members of the new set. These objects must contain `id` and `type` properties, and the `type` property must be `policies` (e.g. `{ "id": "pol-u3S5p2Uwk21keu1s", "type": "policies" }`). **Important:** This deprecated option will be removed in the future in favor of VCS policy sets.
```

content/source/docs/enterprise/api/policy-sets.html.md
```diff
@@ -148,16 +223,17 @@ Status  | Response                                      | Reason
...
 This endpoint supports pagination [with standard URL query parameters](./index.html#query-parameters); remember to percent-encode `[` as `%5B` and `]` as `%5D` if your tooling doesn't automatically encode URLs.
 
-Parameter      | Description
----------------|------------
-`page[number]` | **Optional.** If omitted, the endpoint will return the first page.
-`page[size]`   | **Optional.** If omitted, the endpoint will return 20 policy sets per page.
-`search[name]` | **Optional.** Allows searching the organization's policy sets by name.
+Parameter           | Description
+--------------------|------------
+`filter[versioned]` | **Optional.** Allows filtering policy sets based on whether they are versioned (VCS or API), or manually updated. Accepts a boolean true/false value. If omitted, all policy sets are returned.
```
nfagerlund:
```suggestion
`filter[versioned]` | **Optional.** `filter[true]` will only show versioned policy sets (VCS-managed or API-managed), and `filter[false]` will only show manually updated policy sets. If omitted, all policy sets are returned.
```

content/source/docs/enterprise/api/policy-sets.html.md
```diff
@@ -282,19 +404,24 @@ Status  | Response                                      | Reason
...
+`data.attributes.vcs-repo.identifier`         | string  | (previous value) | The identifier of the VCS repository in the format `<namespace>/<repo>`. For example, on GitHub, this would be something like `hashicorp/my-policy-set`.
+`data.attributes.vcs-repo.oauth-token-id`     | string  | (previous value) | The OAuth Token ID to use to connect to the VCS host.
+`data.attributes.vcs-repo.ingress-submodules` | boolean | (previous value) | Determines whether repository submodules will be instantiated during the clone operation.
+`data.attributes.policies-path`               | boolean | (previous value) | The sub-path within the attached VCS repository to clone. All files and directories outside of this sub-path will be ignored. This option may only be specified when a VCS repo is present. **Note:** When this option is configured, commits in the repository which are outside of the given policies-path will be ignored. When this option is changed, the repository will be automatically cloned again to ensure this path value is reflected in the policy set contents.
```
nfagerlund: (same as above)

content/source/docs/enterprise/api/policy-sets.html.md
```diff
@@ -574,9 +705,60 @@ curl \
   https://app.terraform.io/api/v2/policy-sets/polset-3yVQZvHzf5j3WRJ1
 ```
 
-## Available Related Resources
+## Create a Policy Set Version
+
+For versioned policy sets which have no VCS repository attached, versions of policy code may be uploaded directly to the API by creating a new Policy Set Version, and uploading a tarball (tar.gz) of data to it.
```
nfagerlund:
```suggestion
For versioned policy sets which have no VCS repository attached, versions of policy code may be uploaded directly to the API by creating a new policy set version and, in a subsequent request, uploading a tarball (tar.gz) of data to it.
```

content/source/docs/enterprise/getting-started/sentinel.html.md
```diff
@@ -0,0 +1,136 @@
+---
+layout: "enterprise2"
+page_title: "Configuring Sentinel Policies - Getting Started - Terraform Enterprise"
```
nfagerlund: Without going too deep on this right now: we should try and get this material into learn.hashicorp.com, @tr0njavolta's bandwidth permitting. The getting-started stuff in the TFE docs could benefit from a pare-back; I'd like to reduce it to more of an overview of the tasks you need to do to get up and running, linking out to both learn and the per-feature docs for all of the heavy lifting. 

But in the meantime, something like this is a promising temporary measure. I think we'll have a better time if we preserve the existing `policies.html.md` filename instead of renaming — we can change the title as needed, but moving things means permanent redirects.

content/source/docs/enterprise/api/policy-sets.html.md
```diff
@@ -86,6 +92,35 @@ Key path                               | Type            | Default | Description
...
+        "data": [
+          { "id": "ws-2HRvNs49EWPjDqT1", "type": "workspaces" }
+        ]
+      }
+    },
+    "type": "policy-sets"
+  }
+}
+```
+
```
chrisarcand: Consider inverting this? "Sample Payload", and then take the existing one and put it below it as "Sample Payload (Deprecated)" or similar.

chrisarcand: Same for the other examples.

content/source/docs/enterprise/api/policy-sets.html.md
```diff
@@ -148,16 +223,17 @@ Status  | Response                                      | Reason
...
 
-Parameter      | Description
----------------|------------
-`page[number]` | **Optional.** If omitted, the endpoint will return the first page.
-`page[size]`   | **Optional.** If omitted, the endpoint will return 20 policy sets per page.
-`search[name]` | **Optional.** Allows searching the organization's policy sets by name.
+Parameter           | Description
+--------------------|------------
+`filter[versioned]` | **Optional.** Allows filtering policy sets based on whether they are versioned (VCS or API), or manually updated. Accepts a boolean true/false value. If omitted, all policy sets are returned.
+`include`           | **Optional.** Allows including related resource data. Value must be one of `workspaces`, `policies`, or `current_version`.
```
chrisarcand: I think it's good you're doc'ing this `include` stuff as we're really bad at it in general, but where we do doc it we usually leave it as a separate section, [like in the Workspaces API](https://www.terraform.io/docs/enterprise/api/workspaces.html#available-related-resources). Up to @nfagerlund really, I just thought I'd point out where we're differing from the consistent norm here.

chrisarcand: (though if we're documenting the pagination stuff here I see no reason not to also...include....`include`, too) 🤷‍♂

content/source/docs/enterprise/api/policy-sets.html.md
ryanuber: The `versioned` key still needs to be specified so like `filter[versioned]=true` or `filter[versioned]=false`. I hadn't seen this long-form used anywhere else, but I'm not opposed to it.

content/source/docs/enterprise/api/policy-sets.html.md
ryanuber: I had also noticed that, and thought "this is probably where I'd come looking for this option", so thus I included it here.

ryanuber:
I think I've addressed the feedback, and I spent a little time updating the actual Sentinel guides section as well to focus on VCS and API-uploaded sets.

PR #839, opened 2019-07-08
API: Terraform Enterprise Policy description attribute

dylanegan:
CHANGELOG: Document the description attribute for Terraform Enterprise policies.

A small change to the API documentation to state that there is a `description` attribute for policies. This field is rendered in the Terraform Enterprise UI and returned in API calls for a policy.


nfagerlund:
👍🏼 thank you!

PR #840, opened 2019-07-08
Update VCS Connection Screenshots

pwelch:
Updating the screenshots for VCS connection settings to show
the OAuth Token ID that is now included.


nfagerlund:
Awesome, thank you for catching these.

PR #844, opened 2019-07-10
API updates for Team

dylanegan:
I intended to add the attribute `visible` to the Team API documentation, but noticed that the endpoint for updating a Team was missing itself. This PR includes that, as well as the aforementioned attribute, which is marked as **(beta)**.


nfagerlund:
Previewed, and it all looks good. Thanks!

dylanegan:
@nfagerlund whoops, I meant to make this draft, as the subsequent change hasn't made its way in yet.

PR #847, opened 2019-07-11
extend: Split Resources page into separate pages and remove hardcoded https://www.terraform.io from links

bflad:
The single page content was very long and we have planned additions to various sections of it. To split this into its own section, we create a new subfolder, convert the existing page into an index page, and setup a redirect for the previous URL.

The content on the section subpages is a lift and shift of the existing content while the index page follows our typical model of introducing each subpage. The sections of introductions on the index page were slightly reordered to reflect more common use cases.


content/source/docs/extend/resources/customizing-differences.html.md
```diff
@@ -0,0 +1,60 @@
...
+sidebar_current: "docs-extend-resources-customizing-differences"
+description: |-
+  Difference customization within Resources.
+---
+
+# Resources - Customizing Differences
+
+Terraform tracks the state of provisioned resources in its state file. The user passed configuration is compared against what is in the state file. When there is a detected discrepancy the user is presented with the difference of what is configured versus what is in state. Sometimes these scenarios require special handling, this is where the `CustomizeDiff` function is used. It is passed a `*schema.ResourceDiff`, a structure similar to `schema.ResourceData` but lacking most write functions like `Set`, while introducing new functions that work with the difference such as `SetNew`, `SetNewComputed`, and `ForceNew`.
+
+While any function can be provided for difference customization, it is recommended to try and compose the behavior using the [customdiff](https://godoc.org/github.com/hashicorp/terraform/helper/customdiff) helper package. This will allow for a more declarative configuration; however, it should not be overused, so for highly custom requirements, opt for a tailor-made function.
```
tombuildsstuff: > While any function can be provided for difference customization, it is recommended to try and compose the behavior using the [customdiff](https://godoc.org/github.com/hashicorp/terraform/helper/customdiff) helper package

Since the customiseDiff method is broken (as it doesn't work if a value's interpolated) - I actually think we shouldn't be recommending customizeDiff as a route (or at least flagging it with this note), since it leads to duplicating the validation effort when that value could be interpolated (as most fields could) - WDYT?

bflad: I'll add a `~> *NOTE:*` at the top for now since removing it leaves out (potentially) valuable documentation for the limited situations it does work. 🙁

PR #853, opened 2019-07-17
sentinel: tfplan, note that applied is omitted on destroyed resources

vancluever:
This will be coming in an upcoming version of Terraform Cloud, which
is a correction to an issue of a destroyed resource's proposed state
still existing in various ways when it should be missing completely
from tfplan.


vancluever:
Merging this ahead of release which should be done later today or early tomorrow morning.

PR #859, opened 2019-07-24
Updates w/ missing word & paren

tracypholmes:
- Example code was missing a `)`
- Paragraph was missing an "and".


nfagerlund:
Thanks!

PR #860, opened 2019-07-29
Add missing details of required proxy settings

hashitop:
This is related to the same issue from several tickets in the past where customers tried to configure TFE behind proxy infrastructure and failing with incorrect setting. 

Existing document is located here on public address

https://www.terraform.io/docs/enterprise/private/install-installer.html#reconfiguring-the-proxy

The PR proposes to add more specific details to address requirement at Docker layer as well as minor correction to add reference to `REPLICATED_OPERATOR_OPTS` that belong to `replicated-operator` service.


nfagerlund:
## Do not merge

This is paused because I'm moving all of the content around right now. I'll make sure it gets rebased onto the new situation, and I'll post here again when it's possible to merge.

glasner:
@nfagerlund this one is good to merge once you're ready for it

nfagerlund:
Rebased, re-filed as #872, and merged!

PR #863, opened 2019-08-01
Update tfe smtp supported protocol

hashitop:
The update is only to get rid of ambiguity around supported SSL communication in SMTP settings where TFE only supports STARTTLS, not SMTPS.

https://app.asana.com/0/157256879238414/1132819292389541/f


glasner:
### DO NOT MERGE
Until @nfagerlund says so :)

nfagerlund:
Rebased and merged as #873! Thank you!

PR #867, opened 2019-08-02
Add new tfrun Sentinel import docs

chrisarcand:
Adds documentation for the upcoming `tfrun` Sentinel import for Terraform Cloud / Enterprise.


content/source/docs/enterprise/sentinel/import/tfrun.html.md
```diff
@@ -0,0 +1,111 @@
...
+required, using [policy sets](/docs/enterprise/sentinel/manage-policies.html)
+to only enforce the policy on non-development workspaces is more appropriate.
+
+[run-glossary]: /docs/glossary.html#run
+[workspace-glossary]: /docs/glossary.html#workspace
+
+## Namespace: Workspace
+
+```
+workspace
```
vancluever: To be consistent with the rest of the imports, can we move the sub-key views into the main namespace view?

content/source/docs/enterprise/sentinel/import/tfrun.html.md
```diff
@@ -0,0 +1,111 @@
...
+
+* **Value Type:** A string-keyed map of values.
+
+The `vcs_repo` value within the [workspace namespace](#namespace-workspace)
+data associated with a VCS repository connected to the workspace.
+
+Details regarding each attribute can be found in the documentation for the Terraform Cloud [Workspaces API](/docs/enterprise/api/workspaces.html).
+
+```
+vcs_repo (map of keys)
```
vancluever: As per the last comment on namespace tree views.

If indeed we are not documenting this under the assumption that it's going to be prone to future change, we could probably just omit this and keep it as the `(map of keys)` that's documented in the workspace namespace view.

chrisarcand: Not that it's going to be prone to change; just that I didn't see a point in basically rewriting and having to have two sources of truth for everything that's already in the Workspaces API docs.

vancluever: Alright, I think this is fine. Just as long as the conveyance here is clear that we're not really making a habit of documenting individual namespaces like this and that this is mainly to illustrate the layout of an attribute that's mainly just supposed to be a map, versus static fields.

content/source/docs/cloud/sentinel/import/tfrun.html.md
```diff
@@ -0,0 +1,108 @@
...
+* **Value Type:** String.
+
+The `working_directory` value within the [workspace namespace](#namespace-workspace)
+contains the configured [Terraform working directory](../../workspaces/settings.html#terraform-working-directory) of the workspace.
+
+This value is an empty string (`""`) when no working directory is set.
+
+### Value: `vcs_repo`
+
+* **Value Type:** A string-keyed map of values.
```
nfagerlund: Can this be null if the workspace isn't connected to a repo?

chrisarcand: Yes! Will note.

chrisarcand:
Awesome, thanks @nfagerlund I was wondering about that as this was sitting waiting to be merged after the split! This will be merged in on Monday.

content/source/docs/cloud/sentinel/import/index.html.md
```diff
@@ -9,7 +9,7 @@ Sentinel Policies for Terraform are defined using the [Sentinel policy
 language](https://docs.hashicorp.com/sentinel/language/). A policy can include
 [imports](https://docs.hashicorp.com/sentinel/concepts/imports) which enable a
 policy to access reusable libraries, external data and functions. Terraform
-Enterprise provides three imports to define policy rules for the configuration,
+Enterprise provides four imports to define policy rules for the configuration,
```
rberlind: Given the addition of the tfrun import and the ordering of the four imports, would it be better to write the following?

"Terraform Enterprise provides four imports to define policy rules for the plan, configuration, and state associated with runs."

Also, should references to "Terraform Enterprise" be changed to "Terraform Cloud"?

rberlind:
I presume we will also update https://www.terraform.io/docs/cloud/sentinel/mock.html#using-mock-data to indicate that the generated tarball will contain 4 files (with the tfrun mock being the fourth).

content/source/docs/cloud/sentinel/import/index.html.md
nfagerlund: When I was originally searching for TFE references to update, I missed several that were broken up by hard wrap, like this one. (Great example of why hard wrap is awful for English prose.) Then I fixed it upstream... which is why there are now merge conflicts again in this branch. I'll go fix that.

chrisarcand:
All of the feedback above addressed!

nfagerlund:
Sweet. Merge at your leisure once the import ships!

PR #870, opened 2019-08-06
Big split followup: broken links and stuff

nfagerlund:


nfagerlund:
Travis is borked, so I'm just goin for it

PR #876, opened 2019-08-09
Update schema-behaviors.html.md

mbfrahry:
After some discussion with @nmuesch and @lafentres,  I've drafted this PR to get rid of some confusion around values that are invalid for attributes.

### DIFF
--- a/content/source/docs/extend/schemas/schema-behaviors.html.md
+++ b/content/source/docs/extend/schemas/schema-behaviors.html.md
@@ -21,6 +21,9 @@ and then recreate the resource if any of the attributes change in the
 configuration, as opposed to trying to update the existing resource.
 
 ## Primitive Behaviors 
+
+-> **Note:** The primitive behavior fields cannot be set to `false`. You can opt out of a behavior by omitting it.
+
 ### Optional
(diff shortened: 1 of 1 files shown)


nfagerlund:
Hmm, so am I understanding right that the only valid uses are "present and `true`" or "absent"? If so, do you think that might want a little more explanation, or is it a familiar situation in Go plugins?

mbfrahry:
Right, it's either true or you don't use it. We've got validation in place that complains if you use `Required: false` instead of `Optional: true`. I don't have the full context for why that is the way it is but it's tripping up provider developers. I'm open to hearing a better solution on representing this issue.

content/source/docs/extend/schemas/schema-behaviors.html.md
```diff
@@ -23,7 +23,7 @@ configuration, as opposed to trying to update the existing resource.
 ## Primitive Behaviors
```
nfagerlund:
```suggestion
## Primitive Behaviors 

-> **Note:** The primitive behavior fields cannot be set to `false`. You can opt out of a behavior by omitting it.

```

mbfrahry: Sounds good! Thanks!

PR #882, opened 2019-08-17
Fix show state version code block

thisisshi:
This pr fixes the code block in the show state version section, it was missing the backtick at the end of the line.


nfagerlund:
Oh, good catch, thank you.

PR #883, opened 2019-08-19
Removed references to Pipelines

rkst:
Removed old references to pipelines, this will be updated when we have more to add about best practices there.


nfagerlund:
Yeah, makes sense to me.

PR #884, opened 2019-08-20
added BBDC support statements/language

vaficionado:
Requested by @glasner


content/source/docs/cloud/vcs/bitbucket-server.html.md
```diff
@@ -1,20 +1,22 @@
...
+Note that Bitbucket Server/Data Center requires both OAuth authentication and an SSH key. The instructions below include SSH key configuration.
 
 -> **Version note:** Terraform Cloud supports Bitbucket Server versions 4.9.1 and newer. HashiCorp does not test older versions of Bitbucket Server with Terraform Cloud, and they might not work as expected. Also note that, although we do not deliberately remove support for versions that have reached end of life (per the [Atlassian Support End of Life Policy](https://confluence.atlassian.com/support/atlassian-support-end-of-life-policy-201851003.html)), our ability to resolve customer issues with end of life versions might be limited.
 
+-> **Bitbucket Data Center Support:** Please note that these instructions are equally applicable to both Bitbucket Server and Bitbucket Data Center. Bitbucket Data Center users will follow the same instructions as a user with Bitbucket Server **with version greater than v5.4.0**, omitting the installation of the Web Post Hooks add-on. You will still select **Bitbucket Server** as the VCS Provider type.
```
caseylang: Is this call out not sufficient? I don't love the `Bitbucket Server/Data Center` scattered throughout and wonder if this note on its own is enough to satisfy the needs of the addition. Presumably those that run BBDC as also aware that it is BitBucket Server and would find their way to this connection type without the modification to title and naming?

vaficionado: I considered that, and am not particularly precious either way - I just wanted to make sure it was explicitly called out in at least one place for SEO/other discoverability purposes. Will gladly revert the other instances and leave it just in the title with this callout if others agree.

caseylang: What does navigation look like with that long of a title? 
image

caseylang: ah, I guess the navigation link isn't being changed at the moment, so we'd look like this

image

vaficionado: I haven't seen it rendered but the sidebar has a good amount of padding around it, I don't expect it would be any wider than "Users, Teams, Organizations" below

vaficionado: OK @caseylang take another look, I've reduced the changes now.

caseylang: image

Keep in mind that your change currently does not modify the navigation link. To do that you would need to update the following as well: 
https://github.com/hashicorp/terraform-website/blob/b7b7e6c33c2f9092bb7620227207d47bf6785988/content/source/layouts/cloud.erb#L60-L62

caseylang:
Cool, I prefer the version with just the call out personally. As long as you find that sufficient from a customer's perspective and Nick agrees, I think we're good

content/source/docs/cloud/vcs/bitbucket-server.html.md
vaficionado: Good call, thanks - added to PR

content/source/docs/cloud/vcs/bitbucket-server.html.md
```diff
@@ -1,17 +1,19 @@
...
-page_title: "Bitbucket Server - VCS Providers - Terraform Cloud"
+page_title: "Bitbucket Server and Data Center - VCS Providers - Terraform Cloud"
 ---
 
 
-# Configuring Bitbucket Server Access
+# Configuring Bitbucket Server/Data Center Access
 
-These instructions are for using Bitbucket Server for Terraform Cloud's VCS features. Bitbucket Server is the on-premise version of Bitbucket; [Bitbucket Cloud has separate instructions,](./bitbucket-cloud.html) as do the [other supported VCS providers.](./index.html)
+These instructions are for using Bitbucket Server for Terraform Cloud's VCS features. Bitbucket Server is the on-premises version of Bitbucket; [Bitbucket Cloud has separate instructions,](./bitbucket-cloud.html) as do the [other supported VCS providers.](./index.html)
```
nfagerlund:
```suggestion
These instructions are for using **Bitbucket Server** for Terraform Cloud's VCS features. 

These instructions also apply to **Bitbucket Data Center,** which is a variant of Bitbucket Server that supports clustering. Terraform Cloud treats these two products identically, and Bitbucket Data Center users will select **Bitbucket Server** as their VCS Provider type. Unless stated otherwise, any reference to Bitbucket Server in this document also applies to Bitbucket Data Center.

[Bitbucket Cloud has separate instructions,](./bitbucket-cloud.html) as do the [other supported VCS providers.](./index.html)
```

content/source/docs/cloud/vcs/bitbucket-server.html.md
```diff
@@ -1,17 +1,19 @@
...
 
 Note that Bitbucket Server requires both OAuth authentication and an SSH key. The instructions below include SSH key configuration.
 
 -> **Version note:** Terraform Cloud supports Bitbucket Server versions 4.9.1 and newer. HashiCorp does not test older versions of Bitbucket Server with Terraform Cloud, and they might not work as expected. Also note that, although we do not deliberately remove support for versions that have reached end of life (per the [Atlassian Support End of Life Policy](https://confluence.atlassian.com/support/atlassian-support-end-of-life-policy-201851003.html)), our ability to resolve customer issues with end of life versions might be limited.
 
+-> **Bitbucket Data Center Support:** Please note that these instructions are equally applicable to both Bitbucket Server and Bitbucket Data Center. Bitbucket Data Center users will follow the same instructions as a user with Bitbucket Server **with version greater than v5.4.0**, omitting the installation of the Web Post Hooks add-on. You will still select **Bitbucket Server** as the VCS Provider type.
+
```
nfagerlund:
```suggestion
```

content/source/docs/cloud/vcs/bitbucket-server.html.md
```diff
@@ -1,17 +1,19 @@
...
 
 Note that Bitbucket Server requires both OAuth authentication and an SSH key. The instructions below include SSH key configuration.
 
 -> **Version note:** Terraform Cloud supports Bitbucket Server versions 4.9.1 and newer. HashiCorp does not test older versions of Bitbucket Server with Terraform Cloud, and they might not work as expected. Also note that, although we do not deliberately remove support for versions that have reached end of life (per the [Atlassian Support End of Life Policy](https://confluence.atlassian.com/support/atlassian-support-end-of-life-policy-201851003.html)), our ability to resolve customer issues with end of life versions might be limited.
 
+-> **Bitbucket Data Center Support:** Please note that these instructions are equally applicable to both Bitbucket Server and Bitbucket Data Center. Bitbucket Data Center users will follow the same instructions as a user with Bitbucket Server **with version greater than v5.4.0**, omitting the installation of the Web Post Hooks add-on. You will still select **Bitbucket Server** as the VCS Provider type.
```
nfagerlund:
```suggestion
```

content/source/docs/cloud/vcs/bitbucket-server.html.md
```diff
@@ -1,17 +1,19 @@
...
+These instructions are for using Bitbucket Server for Terraform Cloud's VCS features. Bitbucket Server is the on-premises version of Bitbucket; [Bitbucket Cloud has separate instructions,](./bitbucket-cloud.html) as do the [other supported VCS providers.](./index.html)
 
 Note that Bitbucket Server requires both OAuth authentication and an SSH key. The instructions below include SSH key configuration.
 
 -> **Version note:** Terraform Cloud supports Bitbucket Server versions 4.9.1 and newer. HashiCorp does not test older versions of Bitbucket Server with Terraform Cloud, and they might not work as expected. Also note that, although we do not deliberately remove support for versions that have reached end of life (per the [Atlassian Support End of Life Policy](https://confluence.atlassian.com/support/atlassian-support-end-of-life-policy-201851003.html)), our ability to resolve customer issues with end of life versions might be limited.
```
nfagerlund:
```suggestion
-> **Version note:** Terraform Cloud supports Bitbucket Server versions 4.9.1 and newer, and Bitbucket Data Center versions 5.4.0 and newer. HashiCorp does not test older versions of Bitbucket Server with Terraform Cloud, and they might not work as expected. Also note that, although we do not deliberately remove support for versions that have reached end of life (per the [Atlassian Support End of Life Policy](https://confluence.atlassian.com/support/atlassian-support-end-of-life-policy-201851003.html)), our ability to resolve customer issues with end of life versions might be limited.
```

content/source/docs/cloud/vcs/bitbucket-server.html.md
vaficionado: See my following comment; I think we should maintain this as I wrote it

content/source/docs/cloud/vcs/bitbucket-server.html.md
vaficionado: I think this is actually confusing as it doesn't draw a distinction between Bitbucket Server (legacy) and Bitbucket Server (native - new) - we have two distinct repo types now, because of technical differences between them - and we have explicit version statements for both, one of which allows us to use Data Center and one of which does not. I think the prior support statement I wrote was clearer as it laid.

content/source/docs/cloud/vcs/bitbucket-server.html.md
vaficionado: You say 'unless stated otherwise' but I don't see anywhere that we actually state otherwise?

nfagerlund: The only spot is in the version note, which has a stricter requirement if you want to use BBDC.

content/source/docs/cloud/vcs/bitbucket-server.html.md
nfagerlund: I'm not sure what you mean here; the rest of the docs don't draw a named distinction between the two, and the UI when creating the connection only asks whether your version number is over the mark. 

The reason I found your original wording confusing was this part: 

> Bitbucket Data Center users will follow the same instructions as a user with Bitbucket Server **with version greater than v5.4.0**, omitting the installation of the Web Post Hooks add-on.

That reads as "any BBDC user should behave as though they were on BBS version 5.4+," when in fact it's sort of the opposite: BBDC users must be on version 5.4+, and then should just follow the appropriate instructions.

vaficionado: OK, I can see how you would think that. This seems OK to me if you are satisfied with it. Given this much more streamlined wording, do you still want to change the title of the page and sidebar nav link or no?

nfagerlund: Yeah, go ahead and leave them as "server and data center." That way users won't have any doubt about which page they need to go to, and once they're there they should be able to understand the situation.

vaficionado:
OK @nfagerlund I applied your suggestions. Take another review pass when you have a minute!

nfagerlund:
And feel free to ignore Travis today; that's unrelated, and I'm following up with it elsewhere.

vaficionado:
Thanks Nic - will just stand by for the next merge, am assuming all is well unless Human You tells me otherwise :)

PR #892, opened 2019-08-22
Add details on editing an OAuth Client

davidcelis:
We had an internal request to specify details on why we only allow updating an OAuth Client's name or key/secret pair, so this is my attempt at that. @nfagerlund let me know if you feel like this should be relocated or reworded!


davidcelis:
@nfagerlund Oh, thanks for that link! I think that based on what was in the Asana task, that fits the bill for what I was trying to do. I don't think it's that people aren't looking at that, it seems like I was just given an outdated ticket. I'm just going to go ahead and close this PR as well as mark the ticket as closed 🎉

nfagerlund:
Ok! 🙌🏼 🎉

PR #896, opened 2019-08-28
Update index.html.md to change backend from 'atlas' to 'remote'

anilmurty:
https://app.asana.com/0/1110699467321258/1133215178188085/f


nfagerlund:
This has some added complexities, so I can't merge as-is until I check some stuff.

nfagerlund:
I updated some outdated info in there, and fixed the syntax of the config. These two backends use totally different config paths for their token, and within TFC's Terraform worker, the remote backend relies on the auto-written CLI config file. (It might as well, since private modules need that too.) 

This also spawned https://github.com/hashicorp/terraform/pull/22636 while I was testing the current behavior. 

Thanks for finding this weird little outdated corner!

PR #900, opened 2019-08-30
Add rel="canonical" links to every page, pointing to terraform.io

nfagerlund:
We discovered that a random copy of the website on the public internet (at an IP
address we didn't recognize) was being indexed by Google and rated more highly
than the terraform.io copy for a particular search.

We can make future accidental copies harmless by marking the real version as
canonical; that should tell search engines to ignore your random test build and
point people to us.

Of course, malicious copies can just remove the canonical link, but we can issue
takedowns or something for those because they're probably also doing something
legitimately bad.

### DIFF
--- a/content/source/layouts/layout.erb
+++ b/content/source/layouts/layout.erb
@@ -6,6 +6,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta name="description" content="<%= description_for(current_page) %>">
 
+    <link rel="canonical" href="<%= File.join(base_url, current_page.url) %>" />
     <link rel="apple-touch-icon" sizes="180x180" href="<%= image_path("favicons/apple-touch-icon.png") %>">
     <link rel="icon" type="image/png" href="<%= image_path("favicons/favicon-32x32.png") %>" sizes="32x32">
     <link rel="icon" type="image/png" href="<%= image_path("favicons/favicon-16x16.png") %>" sizes="16x16">


content/source/layouts/layout.erb
```diff
@@ -6,6 +6,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta name="description" content="<%= description_for(current_page) %>">
 
+    <link rel="canonical" href="<%= File.join(base_url, current_page.url) %>" />
```
paultyng: Is there anyway to override `base_url`? Does it change when people run the site locally or anything? Curious if it would be better to just hardcode this to `https://www.terraform.io/` here. Otherwise though, LGTM

nfagerlund: I made sure to check! That's [hardcoded in content/config.rb](https://github.com/hashicorp/terraform-website/blob/master/content/config.rb#L1), and doesn't change in preview builds. So you'd have to mess with it on purpose. I checked the source of a preview build, too: 

Screen Shot 2019-08-30 at 11 14 20 AM

glasner: This maybe beyond the scope of this PR, but we should be picking a single URL for `/index.html` pages. In the example above, both of these URLs work:

https://www.terraform.io/docs/cloud/
https://www.terraform.io/docs/cloud/index.html

We need to choose one and then redirect the other. If we're going to choose to leave of the `/index.html` the code above needs to trim it from `current_page.url`

nfagerlund: Right now it does consolidate on a single URL. If you go to the page with the bare /, it'll say index.html is the canonical one. That matches with what most of our links do (for some historical reason, we basically always link to index.html). 

Legit to disagree with which one should be the "official" path, but yeah, that's a conversation for the next web platform, not for this.

PR #902, opened 2019-09-04
Fix syntax on test case example

lgfa29:
`resource.TestCase` is a `struct` that is passed as a parameter to the `resource.Test` function, so the parentheses should be placed at the end of the function call.


nfagerlund:
Good catch, thank you.

PR #903, opened 2019-09-04
Fix indentation in code sample for resource.TestCase Providers

lgfa29:
Seems like there was a tab char instead of spaces that was being rendered as 8 spaces causing the `struct` fields to look misaligned.


nfagerlund:
Cool cool. I took a peek, and actually this file has a completely wild mix of hardtabs and spaces, but I guess this is the only example that looked wrong per se, so I'm not inclined to go clean it up. (I think all our go code might use a formatting commit hook to enforce hardtabs, and the docs text _definitely doesn't,_ so some copy/paste mixups seem inevitable.)

PR #909, opened 2019-09-08
Update index.html.md

jlorich:
Premise -> Premises.

"A premise is a proposition or presupposition in a logical argument. Premises means a property—such as a building and its grounds."

https://en.wikipedia.org/wiki/On-premises_software (Yes I'm citing wikipedia, take that teachers).

### DIFF
--- a/content/source/docs/enterprise/index.html.md
+++ b/content/source/docs/enterprise/index.html.md
@@ -9,7 +9,7 @@ This is the documentation for Terraform Enterprise.
 
 ## About Terraform Cloud and Terraform Enterprise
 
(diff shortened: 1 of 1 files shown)


nfagerlund:
@jlorich Okay, you're right. Thank you. 😅 Could I get you to sign the CLA? A spelling fix is probably small enough that we could work out an exception, but it's easier and cleaner if you're ok just ticking the box.

nfagerlund:
Looks like we decided to work around this by calling it "self-hosted" instead. 😅 Well, all's well that ends well.

nfagerlund:
https://github.com/hashicorp/terraform-website/pull/932/

PR #918, opened 2019-09-17
Fix the navigation sidebar title for migrating from Local Terraform

flagbug:
In https://github.com/hashicorp/terraform-website/pull/865, the migration document has changed to mention Terraform Cloud, but it seems like the navigation sidebar title hasn't been updated


nfagerlund:
Ah, good catch. (I wish we were just pulling these titles out of the page frontmatter. Maybe when we switch site generator platforms.)

PR #919, opened 2019-09-17
WIP: Update Sentinel tag enforcement for AWS to point at second generation

justinretzolk:
We've seen a few situations lately where people have referred to the first generation of the AWS tag enforcement Sentinel policy and have ran into issues due to not compensating for destroyed resources.

Initially I'd considered updating all of the links within the example pages to point at the second generation, however, there are some that show up on both generations and some that don't. I'm more than happy to add the others/change things around more, but wanted to focus on getting this one corrected as a starting point.

### DIFF
--- a/content/source/docs/cloud/sentinel/examples.html.md
+++ b/content/source/docs/cloud/sentinel/examples.html.md
@@ -12,7 +12,7 @@ This page lists some example Sentinel policies. These examples are not exhaustiv
 ### Amazon Web Services
 
 * [Enforce owner allow list on `aws_ami` data source](https://github.com/hashicorp/terraform-guides/blob/master/governance/first-generation/aws/enforce-ami-owners.sentinel)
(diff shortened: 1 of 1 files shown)


rberlind:
Pointing at second generation policies whenver they exist is definitely a good idea.  But the SE team has not yet implemented all of the first-generation policies as second-generation policies.  I think I did review this a few months ago and updated all that could be updated to second-gen at that time, but I might have missed the aws tags policy.

One good thing is that the second-gen policies now do test `r.destroy and not r.requires_new` instead of just testing `r.destroy` which erroneously skipped resources being destroyed and then re-created.

vancluever:
PS: I did notice that Roger did mention that he did not have a tags example in that repo, but I don't think that's the point here - I think that the point is that the old repo has some patterns that we are not endorsing anymore and as such that repo should be deprecated.

rberlind:
Actually, we do have tags policies in the second-generation directory, but there are a few other types of policies that have not yet been implemented in the second-generation style.

Some of the SEs wanted to keep the first-generation policies around for historical reasons and to use in demos or workshops where the intention is not to use production-ready policies but to keep them simpler and easier to understand.

PR #922, opened 2019-09-18
Revise install software page to steer people away from doing that

nfagerlund:
...and catch up with the lack of sudo in the current Terraform Cloud.


content/source/docs/cloud/run/install-software.html.md
```diff
@@ -5,28 +5,39 @@ page_title: "Installing Software in the Run Environment - Runs - Terraform Cloud
...
+    chmod 0755 jq
+    # Do some kind of JSON processing with ./jq
+    EOH
+      }
+    }
+    ```
+
+When downloading software with `local-exec`, try to associate the provisioner block with the resource(s) that the software will interact with. If you use a null resource with a `local-exec` provisioner, you must ensure it can be properly configured with [triggers](/docs/provisioners/null_resource.html#example-usage). Otherwise, a null resource with the `local-exec` provisioner will only install software on the initial run where the `null_resource` is created. The `null_resource` will not be automatically recreated in subsequent runs and the related software won't be installed, which may cause runs to encounter errors.
+
+-> **Note:** Terraform Enterprise instances can sometimes allow `sudo` commands during Terraform runs. However, even when `sudo` is allowed, using the worker OS's package tools during runs is still usually a bad idea. You will have a much better experience if you can move your provisioner actions into a custom provider or an immutable machine image.
```
alisdair: "sometimes" is a little wishy-washy. I think this is accurate:

```suggestion
-> **Note:** Terraform Enterprise instances can be configured to allow `sudo` commands during Terraform runs. However, even when `sudo` is allowed, using the worker OS's package tools during runs is still usually a bad idea. You will have a much better experience if you can move your provisioner actions into a custom provider or an immutable machine image.
```

PR #924, opened 2019-09-19
Updated Cloud to Enterprise

chuysmans:


nfagerlund:
@chuysmans Within documentation, we deliberately refer to the application as "Terraform Cloud". See the [front page](https://www.terraform.io/docs/enterprise/index.html) of the TFE docs for further explanation!

PR #929, opened 2019-09-20
Specify what permissions a custom provider plugin file must have

achuchulev:
Plugin file of a custom provider must have read and execute permissions for Terraform to use it so it is good to include this to the documentation.


rberlind:
The change makes sense to me.

nfagerlund:
Yeah, seems good. Thanks!

PR #930, opened 2019-09-23
Cloud API token docs

ryanuber:
Dusting this off a bit from some revisions I made a while back in response to customers constantly asking questions about various access levels of tokens. We didn't do a great job of spelling out precisely what a token could or could not do, so this adjustment attempts to clear that up a bit.

I renamed the section from "Service Accounts" to "API Tokens", because a user doesn't think about this as a service account really - it's just a token to them, the fact that it's a service account under the hood is just a detail they don't need to worry about.


ryanuber:
https://app.asana.com/0/114662248701346/1111383297904812

content/source/docs/cloud/users-teams-organizations/api-tokens.html.md
```diff
@@ -0,0 +1,87 @@
...
+sidebar_current: "docs-cloud-users-teams-organizations-api-tokens"
+---
+
+# API Tokens
+
+Terraform Cloud supports three distinct types of API tokens with varying levels of access: user, team, and organization. There are differences in access levels and generation workflows for each of these token types, which are outlined below.
+
+Like with [user tokens](./users.html#api-tokens), team and organization API tokens are displayed only once when they are created, and are obfuscated thereafter. If the token is lost, it must be regenerated. When a token is regenerated, the previous token immediately becomes invalid.
+
+~> **Important:** Unlike user tokens, team and organization tokens cannot authenticate with Terraform's `atlas` backend. This means they can't be used to run Terraform locally against state stored in Terraform Cloud, access `terraform_remote_state` data sources during a run, or [migrate local state into Terraform Cloud](../migrate/index.html). If you need to use the `atlas` backend for any reason, use a [user token][./users.html#api-tokens].
```
chrisarcand: I thought we finally got rid of this, that enough time has passed where the `atlas` backend is moot at this point.

content/source/docs/cloud/users-teams-organizations/api-tokens.html.md
```diff
@@ -0,0 +1,87 @@
...
+layout: "cloud"
+page_title: "API Tokens - Terraform Cloud"
+sidebar_current: "docs-cloud-users-teams-organizations-api-tokens"
+---
+
+# API Tokens
+
+Terraform Cloud supports three distinct types of API tokens with varying levels of access: user, team, and organization. There are differences in access levels and generation workflows for each of these token types, which are outlined below.
+
+Like with [user tokens](./users.html#api-tokens), team and organization API tokens are displayed only once when they are created, and are obfuscated thereafter. If the token is lost, it must be regenerated. When a token is regenerated, the previous token immediately becomes invalid.
```
chrisarcand: This feels out of place here; maybe after the sections below?  Or maybe it's just the 'Like with user tokens' part, and redoing like this would work:

> API tokens are displayed only once when they are created, and are obfuscated thereafter. If the token is lost, it must be regenerated. With a team or an organization token, the previous token immediately becomes invalid.

chrisarcand: (I know you probably just copied this sections that already existed)

content/source/docs/cloud/users-teams-organizations/api-tokens.html.md
```diff
@@ -0,0 +1,87 @@
...
+| Create or modify organizations | 🔶          |             |                     |
+| Manage organization tokens     | 🔶          |             |                     |
+| **Sentinel**                   |             |             |                     |
+| Manage Sentinel policies       | 🔶          | 🔶          | 🔷                  |
+| Manage policy sets             | 🔶          | 🔶          | 🔷                  |
+| **Integrations**               |             |             |                     |
+| Manage VCS connections         | 🔶          | 🔶          | 🔷                  |
+| Manage SSH keys                | 🔶          | 🔶          | 🔷                  |
+| **Modules**                    |             |             |                     |
+| Manage Terraform modules       | 🔶          | 🔶          |                     |
```
chrisarcand: Whooooa this is doooope. 💯 

`Override policy checks` is one that should be added here, we get asked that one a lot.

If `Manage organization tokens` is here, might want to add `Manage team tokens` (pretty sure I added managing team tokens with org tokens a while back 😑) (though IMO should just leave both out, maybe)

content/source/docs/cloud/users-teams-organizations/api-tokens.html.md
ryanuber: Oh you are right. I wrote the original commit so long ago it was actually relevant back in those days 😂 I'll just remove this.

content/source/docs/cloud/users-teams-organizations/api-tokens.html.md
ryanuber: Yep good call, these came byte-for-byte from a different file, but now is a good time to fix this up.

content/source/docs/cloud/users-teams-organizations/api-tokens.html.md
ryanuber: 👍 on `Override policy checks`, good call.

`Manage team tokens` is already in here, under `Teams`!

ryanuber:
I think this is ready for another glance, made the suggested changes 👍. I updated links in various places as well but its mostly find-and-replace.

content/source/docs/cloud/users-teams-organizations/api-tokens.html.md
```diff
@@ -0,0 +1,86 @@
...
+layout: "cloud"
+page_title: "API Tokens - Terraform Cloud"
+sidebar_current: "docs-cloud-users-teams-organizations-api-tokens"
+---
+
+# API Tokens
+
+Terraform Cloud supports three distinct types of API tokens with varying levels of access: user, team, and organization. There are differences in access levels and generation workflows for each of these token types, which are outlined below.
+
+API tokens are displayed only once when they are created, and are obfuscated thereafter. If the token is lost, it must be regenerated. When a token is regenerated, the previous token immediately becomes invalid.
```
chrisarcand: Heh the only thing I noticed the last time we reviewed this line is that this _isn't_ true for user tokens. You can have multiple tokens and the previous 'one' isn't invalidated when you generate a new one.

ryanuber: Ah yeah! It's mentioned in the individual team + org sections anyways so this is also redundant. Removing it.

PR #932, opened 2019-09-24
Change tfe description

rberlind:
It is incorrect to describe TFE as "on-premise".  While it is on-premise when installed on VMware, it is often installed in virtual private networks (VPCs) in AWS, Azure, and Google Compute Cloud.  So, it is better to use the term "self-hosted" which is also what we use on https://www.hashicorp.com/products/terraform/pricing


rkst:
If @nfagerlund is fine with this it looks good to me 🛳️

PR #942, opened 2019-09-28
(feature) Updates starter workflow to use new yaml format for GitHub Actions v2

booyaa:
To address this issue: https://github.com/hashicorp/terraform-github-actions/issues/67

Still need to update individual action references, but will do as separate PRs.

### DIFF
--- a/content/source/docs/github-actions/getting-started/index.html.md
+++ b/content/source/docs/github-actions/getting-started/index.html.md
@@ -31,54 +31,39 @@ Terraform's GitHub Actions on new and updated pull requests.
 
 1. Replace the default workflow with the following:
 
-    ```hcl
-    workflow "Terraform" {
-      resolves = "terraform-plan"
-      on = "pull_request"
-    }
-
-    action "filter-to-pr-open-synced" {
-      uses = "actions/bin/filter@master"
-      args = "action 'opened|synchronize'"
-    }
-
-    action "terraform-fmt" {
-      uses = "hashicorp/terraform-github-actions/fmt@v<latest version>"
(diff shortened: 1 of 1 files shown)


content/source/docs/github-actions/getting-started/index.html.md
```diff
@@ -31,54 +31,39 @@ Terraform's GitHub Actions on new and updated pull requests.
...
+    name: Terraform
+    jobs:
+      filter-to-pr-open-synced:
+        runs-on: ubuntu-latest
+        steps:
+        - uses: actions/checkout@master
+        - name: filter-to-pr-open-synced
+          uses: actions/bin/filter@master
+          with:
+            args: action 'opened|synchronize'
```
ethomson: This step is outdated in the YAML syntax.  We've added event filtering directly to the YAML.  You can achieve this by movig the types up into the `on` trigger above.  In this case:

```yaml
on:
  pull_request:
    types: [ opened, synchronize ]
```

However, the default triggers for `on: pull_request` are actually [opened, synchronized, reopened](https://help.github.com/en/articles/events-that-trigger-workflows) which seem like reasonable defaults here, too, so you might just remove this step and not update the `on: pull_request` line at all.

James-Quigley: Additionally with the latest release of 0.4.2, you no longer have to run these checks exclusively on PRs. You can just run it on a normal push

nfagerlund:
@booyaa @ethomson @James-Quigley I think this is superseded by #988, right? Shall I close?

PR #944, opened 2019-10-01
Update SAML AAD documentation to reflect Azure's updated UI

dylanegan:
This came through feedback from a customer. The existing documentation doesn't seem to match how the UI is currently in Azure. I verified these steps before they sent through the change suggestions themselves.

They've also suggested adding notes like the following to step 4.

> If you're using the custom app (not the Azure Marketplace app), you see two default roles: user and msiam_access. For the Marketplace app, msiam_access is the only default role. You don't need to make any changes in the default roles.

> You can only add new roles after msiam_access for the patch operation. Also, you can add as many roles as your organization needs. Azure AD will send the value of these roles as the claim value in the SAML response. To generate the GUID values for the ID of new roles use the web tools like this

These notes come from https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-enterprise-app-role-management.

My only other thought for this page is that we might want to consider talking about the `site-admins` role as the example, rather than an arbitrary `Dev` role that doesn't seem to connect to anything else in our SAML documentation.


content/source/docs/enterprise/saml/identity-provider-configuration-aad.html.md
```diff
@@ -45,11 +45,11 @@ Follow these steps to configure Azure Active Directory (AAD) as the identity pro
...
     ![Screenshot: AAD App Registration](./images/sso-aad-saml-app-registration.png)
 3. In the manifest editor, locate the "appRoles" block. This is where you will add additional roles that map users and groups to teams in TFE.
     ![Screenshot: Manifest Editor](./images/sso-aad-saml-manifest-approles.png)
-4. The "appRoles" block may contain roles automatically generated by AAD. New roles should be added after the system roles and must contain a unique GUID value for the ID value of the new role. You can use a tool such as [GUID Generator](https://www.guidgenerator.com) to create these GUIDs. Click "Save" to add the roles.
+4. The "appRoles" block may contain roles automatically generated by AAD. Leave the automatically generated roles GUID with their default values.  New roles should be added after the system roles and must contain a unique GUID value for the ID value of the new role. You can use a tool such as [GUID Generator](https://www.guidgenerator.com) to create the GUIDs for these new roles. Click "Save" to add the roles.
```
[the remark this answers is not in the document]
dylanegan: Thanks! This was purely a copy and paste from their recommendations. I wonder if they don't realise that they can contribute directly.

dylanegan:
@nfagerlund any thoughts on their other recommendation around notes? Also should we adjust the role to be something more useful? I am neither here nor there on those changes.

dylanegan:
@burstingninja are you referring to the first screen in section 2 or in another section? The first screenshot is updated (https://github.com/hashicorp/terraform-website/blob/84c9ef003ee83177fb68532a02223de35fc8eafd/content/source/docs/enterprise/saml/images/sso-aad-saml-app-registration.png) and will be reflected on the website once the PR has been merged and deployed.

dylanegan:
@burstingninja ah, odd. Thanks for double checking though. 🙇

burstingninja:
> @burstingninja ah, odd. Thanks for double checking though. 🙇

No problem @dylanegan  :)

nfagerlund:
@dylanegan Ah, I missed the notes first time around! Hmm:

- I don't think the number of autogenerated roles is that important to point out.
- It looks like we already say that you need to put your roles after the system roles. 
- The "you can add as many as you need" bit might be worth adding, and that also seems like a good time to mention the site-admins role (which yeah, seems like a good idea).

content/source/docs/enterprise/saml/identity-provider-configuration-aad.html.md
```diff
@@ -45,13 +45,16 @@ Follow these steps to configure Azure Active Directory (AAD) as the identity pro
...
 3. In the manifest editor, locate the "appRoles" block. This is where you will add additional roles that map users and groups to teams in TFE.
     ![Screenshot: Manifest Editor](./images/sso-aad-saml-manifest-approles.png)
-4. The "appRoles" block may contain roles automatically generated by AAD. New roles should be added after the system roles and must contain a unique GUID value for the ID value of the new role. You can use a tool such as [GUID Generator](https://www.guidgenerator.com) to create these GUIDs. Click "Save" to add the roles.
+4. The "appRoles" block may contain roles automatically generated by AAD. Leave the automatically generated role GUIDs with their default values.  New roles should be added after the system roles and must contain a unique GUID value for the ID value of the new role. You can use a tool such as [GUID Generator](https://www.guidgenerator.com) to create the GUIDs for these new roles. Click "Save" to add the roles.
+
+    -> **Note:** You can add as many roles as your organization needs, such as the `site-admins` role. Azure AD will send the value of these roles as the claim value in the SAML response.
```
dylanegan: @nfagerlund like this?

Screen Shot 2019-10-03 at 15 09 28

nfagerlund: @dylanegan Yeah!!

I'm going to go ahead and merge this, and we can follow up with more adjustments if we feel the urge later.

dylanegan:
@nfagerlund thanks for merging!

PR #945, opened 2019-10-01
Tidy up header layout, and enable registry links

nfagerlund:
This CSS was a mess (full disclosure: _my_ mess), so here's a better approach using only grid, align-self, and line-height.

The goal here:

- The baselines of the main page title and the sidebar title (if present) need to more or less line up. 
- The dynamic navigation controls (expand/filter and "jump to section") need to line up. 
- Two-line page titles happen fairly frequently, so they should look reasonable and not misalign things. (Three-line titles are rare, so they can go ahead and misalign things.)

#### Looks about the same, which was the hope. 

Screenshot_2019-11-07 Pre-Install Checklist - Before Installing - Terraform Enterprise - Terraform by HashiCorp

#### With gridlines turned on: 

Screenshot_2019-11-07 Pre-Install Checklist - Before Installing - Terraform Enterprise - Terraform by HashiCorp(1)

#### Baselines don't look too far off. Pardon my improvised ruler. 

Screen Shot 2019-11-07 at 4 17 05 PM

Screen Shot 2019-11-07 at 4 17 24 PM

### Also

The reason I'm doing this now is because the registry team needs to be able to link to registry pages from the front page of a given provider's documentation, without having to edit that provider's actual docs text. The second commit shows how we can do that (without turning it on for any pages yet), and the CSS follows the design I was given for those links. 

Screen Shot 2019-11-07 at 4 27 32 PM

Baselines line up for that link, too. The square-with-arrow icon is just pulled out of the Bootstrap stuff that's already available, but I can swap in a tidier-looking SVG if you've got one for me. 

If there's no link enabled for a page, that grid column just squishes down and stays out of the way. 

-----

Notable drawback: Since this uses javascript to arrange things up top, it can look a little goofy if you have javascript disabled.

image

I have some ideas for how to mitigate that, but our sites are gonna do a lot of other other goofy stuff with JS disabled, so a little extra whitespace might be the least of it.

### DIFF
--- a/content/source/layouts/inner.erb
+++ b/content/source/layouts/inner.erb
@@ -5,6 +5,7 @@
       <a href="#inner" id="screenreader-skip" class="sr-only sr-only-focusable">Skip to content</a>
       <a href="#inner" id="mobile-skip" class="visible-xs">(Skip to content ⤵︎)</a>
       <div id="controls-placeholder" style="display: none;"></div>
+      <div id="sidebar-header-grid"></div>
       <% if content_for?(:sidebar) %>
       <%
         sidebar_text = yield_content(:sidebar)
@@ -16,6 +17,10 @@
     </div>
 
     <div id="inner" class="col-sm-8 col-md-9 col-xs-12" role="main">
+      <div id="inner-header-grid">
+        <div id="external-docs-link">
+        </div>
+      </div>
       <%= yield %>
     </div>
   </div>
--- a/content/source/layouts/inner.erb
+++ b/content/source/layouts/inner.erb
@@ -16,9 +16,17 @@
       <% end %>
     </div>
 
+    <%
+      external_docs_links = {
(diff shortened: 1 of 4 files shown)


content/source/layouts/inner.erb
```diff
@@ -15,7 +16,19 @@
       <% end %>
     </div>
 
+    <%
+      external_docs_links = {
+        # '/docs/providers/aws/index.html' => 'https://registry.terraform.io/path/to/provider',
```
topfunky: Will this be populated in future commits? It's blank now so it seems that the "View on Terraform Registry" link will never appear.

topfunky: Update: Commit says "without turning it on for any pages yet" so it will be used later.

gregone:
@nfagerlund is this ready to merge? or can I help with pushing it over the line?

nfagerlund:
@gregone okay! Please let me know when your side is ready to add links to registry pages; if I have a full list, I can enable links very quickly.

PR #946, opened 2019-10-02
Proposing update bitbucket server integration document

hashitop:
A proposal to incorporate a note regarding port 7999 requirement for Bitbucket Server SSH communication


nfagerlund:
This isn't the right place for this information; instead, it should be in the TFE network access requirements. If you think it's necessary, we can put a reminder/link from this page.

nfagerlund:
Hmm, I might be wrong about that. Lemme look a little closer.

nfagerlund:
No, ok, it SHOULD be on this page, but it needs to be up at the top as part of the note about network connectivity.

PR #947, opened 2019-10-02
Propose adding a note for wildcard proxy setting on replicated console

hashitop:
Propose adding a note for wildcard proxy setting on replicated console to state clearly that it isn't supported


content/source/docs/enterprise/install/installer.html.md
```diff
@@ -44,7 +44,11 @@ On the Console Settings page, there is a section for HTTP Proxy:
 
 ![Terraform Enterprise HTTP Proxy Settings](./assets/ptfe-http-proxy.png)
 
-This change updates the proxy settings for the Terraform Enterprise application services. To update the proxy settings for the installer (for example, to handle configuration tests correctly), additional steps are necessary:
+This change updates the proxy settings for the Terraform Enterprise application services.
+
+~> **Note:** Wildcard syntax is not supported, for example, `*.yourdomain.com` will not be recognized by Terraform Enterprise application services.
```
nfagerlund: This addition doesn't seem to make sense here -- the field shown in the settings screenshot expects a hostname for a proxy server, not a range of hosts, so it doesn't seem necessary to clarify that it doesn't take a wildcard. It's like saying that the URL bar in your web browser doesn't accept wildcard syntax.

What was the situation that inspired this? Did someone enter a wildcard in this exact setting?

glasner: Yeah, I agree here. Maybe we can just make the help text on the application be more specific that it's the address of the proxy server?

nfagerlund:
Since this PR doesn't make sense as filed, I'm going to close it. Feel free to open a new PR if we can figure out what the customer had a problem with.

PR #948, opened 2019-10-02
Update provisioners in tfconfig

rberlind:
Added note to indicate that the provisioners value under data sources will always be an empty map or null since data sources cannot actually have provisioners.
Also removed "or data source" from the provisioners namespace.

I also wonder if we should remove provisioners from under data in the tree at the top of this page.  But since provisioners will show up under data sources in tfconfig as an empty map or null, one can argue that we should not change the tree.


rberlind:
It's not really a TOC, but the namespace tree.  It would show up under data sources, but only as an empty list or null.

PR #774, opened 2019-05-08
Update the API driven runs documentation

justinretzolk:
- Updated Step 5 to match the documentation on the [Upload Configuration Files](https://www.terraform.io/docs/enterprise/api/configuration-versions.html#upload-configuration-files) API endpoint
- Updated the corresponding code from Step 5 in the complete script
- Added comments in the complete script that align with each step, along with a link to the documentation
- Removed reference to tfe-cli

PR #950, opened 2019-10-08
Fixing typo in `tar` command.

sudomateo:
Adding `-C` back to `tar` command to ensure the proper directory gets archived. It was incorrectly removed in https://github.com/hashicorp/terraform-website/pull/774.


sudomateo:
It might be one of those differences, but at least the `-C` explicitly shows the reader that `tar` is changing into the directory before processing any files.

Thank you for the catch on the other `tar` command. I just amended my commit to include it.

nfagerlund:
Travis failure is a broken link in the oracle provider, and it's already fixed; feel free to ignore.

PR #956, opened 2019-10-11
Create dedicated SDK Community page

radeksimko:
Both pages link to each other, but I decided to make the link from SDK Community page more visible, because there is higher chance that people will keep landing there for some time, until the removal of the redirect is fully propagated, and/or they may just come through an outdated link.

### DIFF
--- a/content/redirects.txt
+++ b/content/redirects.txt
@@ -119,7 +119,6 @@
 
 # Extending Terraform section
 /guides/writing-custom-terraform-providers.html          /docs/extend/writing-custom-providers.html
-/community.html                                          /docs/extend/community/index.html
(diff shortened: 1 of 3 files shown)


content/redirects.txt
```diff
@@ -119,7 +119,6 @@
 
 # Extending Terraform section
 /guides/writing-custom-terraform-providers.html          /docs/extend/writing-custom-providers.html
-/community.html                                          /docs/extend/community/index.html
```
radeksimko: As mentioned above in this document we'll need to coordinate the removal of this redirect with engserv team.

appilon (approved):

PR #961, opened 2019-10-15
Update Homepage [Cloud] How TF Works Image

JMFURY:
**This PR updates an image on the Homepage to be consistent with the update here:** https://www.hashicorp.com/products/terraform#how-terraform-works

I have tested this to ensure this slightly larger image is still good to go cross-browser.

Other Notes:
- The removal of an explicit `width` on Line 28 of the `_pricing-tile.scss` component stylesheet fixes a tiny issue that remained from previous launch (happy to provide a screenshot of this as necessary). Figured this was small enough that it was okay to include here, but let me know otherwise.
- Similarly to the above, the `alt` text was wrong for the Image we're swapping out here so I've included this fix as well


JMFURY:
@nfagerlund - Thanks very much for the review. It doesn't look like I have write access to merge, so if you're okay with it I will let you handle it. 

I _think_ these changes are unrelated to the build failing on this, but do let me know if there's anything I can help with there as well..

nfagerlund:
@JMFURY Yeah, the travis failure is a transitory thing -- when we add a new provider, it needs modifications in both the terraform repo and here, and if the TF ones get merged first, you get a broken link for a while.

PR #964, opened 2019-10-16
Be more emphatic about SSH access to BBS, call out default port

nfagerlund:
This supersedes https://github.com/hashicorp/terraform-website/pull/946. (Sorry @hashitop! I needed to do an edit that wouldn't work with the suggestions feature, and couldn't push to your branch.)

image


nfagerlund:
FUN FACT: once you pointed out the whacky default port, I had an ominous feeling about how we handle non-default SSH ports, so I went and asked some questions. Long story short, Casey thinks we now know the answer to one of our grossest and longest-standing BBS bugs (randomly failing initial ingress for workspaces that clears up after we get the first webhook).

nfagerlund:
yeah, same.

PR #965, opened 2019-10-17
Update installer.html.md

susiesolis:
Changed order of 'Run the Installer - Airgapped - Prepare the instance' to show Docker first.


justinretzolk:
Makes sense to me. We've had some questions from time to time about whether Docker needs to be pre-installed, so moving this to the top seems like a wise move.

PR #971, opened 2019-10-25
Add to project readme: screenshot guidelines, and more info about teamcity deploys

nfagerlund:


nfagerlund:
Just in time to be accurate for another day or two! 😂 Alvin's moving us to circleCI, so I'll probably be back to this section shortly.

PR #974, opened 2019-10-30
Add docs for importing providers.

paddycarver:
Add a guide to cover the ways you can depend on providers (gRPC, schema
export) and explicitly call out that importing providers as a Go module
is not supported.

### DIFF
--- a/content/source/docs/extend/best-practices/depending-on-providers.html.md
+++ b/content/source/docs/extend/best-practices/depending-on-providers.html.md
@@ -41,6 +41,10 @@ If not, please reach out and [open an issue](https://github.com/hashicorp/terraf
 outlining your use case, and we'll work with you to find an appropriate way
 to interface with Terraform to meet your use case.
 
+~> **Note**: a notable exception to this is providers importing other providers
+in their test files. A better solution for that use case is in the works, but
+at the moment importing the providers is the best available solution.
+
 ## Exporting the Schema
 
 Some projects just care about the schema and resources a provider presents. As
--- a/content/source/docs/extend/best-practices/depending-on-providers.html.md
+++ b/content/source/docs/extend/best-practices/depending-on-providers.html.md
@@ -1,0 +1,55 @@
+---
+layout: "extend"
+page_title: "Extending Terraform: Depending on Providers"
+sidebar_current: "docs-extend-best-practices-depending-on-providers"
+description: |-
+  How to safely depend on providers and understand their interfaces.
+---
+
+## Depending on Providers
+
+Terraform's providers are a substantial amount of code, and occasionally it
+makes sense to depend on their functionality. The most straightforward and
+obvious way to depend on a provider is to depend on the Terraform CLI, but
+occasionally it makes sense to rely on a provider in a different context.
+
+This guide lays out the supported ways to interface with and depend on
+Terraform's providers. Unless the provider explicitly states otherwise, no
+other compatibility guarantees are provided.
+
+### Importing Providers as Go Modules
+
+Terraform's providers are written as Go packages, and they mostly use Go
+modules as their dependency management solution. This makes it tempting to
+import the provider as a dependency of your Go code, and to call its exposed
+interface. This is explicitly an **unsupported** way to interact with providers
+and provider maintainers make no guarantees around backwards compatibility or
+the continued functioning of code that does this.
+
+Providers are unable to be imported as Go modules reliably because their
+versioning scheme is intended to convey information about the Terraform config
+interface the provider presents. It's unable to capture both the configuration
+interface _and_ the Go API interface in a useful way, as what is compatible in
+one may not be compatible in the other. Rather than give the impression that
+the package should be imported by using the `/vX` suffix now mandated for
+versions after 2.0.0, providers have chosen to make their incompatibility with
+being imported into Go code explicit.
+
+If you find yourself needing to do this, perhaps one of the methods below will
+work for you, and is explicitly supported and covered under versioning policies.
+If not, please reach out and [open an issue](https://github.com/hashicorp/terraform/issues/new)
+and we'll try to come to an appropriate solution.
+
+### Exporting the Schema
+
+Some projects just care about the schema and resouces a provider presents. As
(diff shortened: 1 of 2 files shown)


content/source/docs/extend/best-practices/depending-on-providers.html.md
```diff
@@ -0,0 +1,55 @@
...
+interface _and_ the Go API interface in a useful way, as what is compatible in
+one may not be compatible in the other. Rather than give the impression that
+the package should be imported by using the `/vX` suffix now mandated for
+versions after 2.0.0, providers have chosen to make their incompatibility with
+being imported into Go code explicit.
+
+If you find yourself needing to do this, perhaps one of the methods below will
+work for you, and is explicitly supported and covered under versioning policies.
+If not, please reach out and [open an issue](https://github.com/hashicorp/terraform/issues/new)
+and we'll try to come to an appropriate solution.
```
radeksimko: I would be more explicit and try to nudge them to come up with a specific use case here, if possible. I mean - correct me if I'm wrong but it's unlikely that we make a provider importable just because someone asks for it, but it's more likely that we'd come up with a solution to a problem that doesn't involve importing providers.

paddycarver: Pushed a quick change! Let me know if that addresses your concern.

radeksimko: That new wording is perfect! :shipit:

content/source/docs/extend/best-practices/depending-on-providers.html.md
```diff
@@ -0,0 +1,56 @@
+---
+layout: "extend"
+page_title: "Extending Terraform: Depending on Providers"
+sidebar_current: "docs-extend-best-practices-depending-on-providers"
+description: |-
+  How to safely depend on providers and understand their interfaces.
+---
+
+## Depending on Providers
```
nfagerlund:
```suggestion
# Depending on Providers
```

Also, the other headers in the file should all turn into H2s.

content/source/docs/extend/best-practices/depending-on-providers.html.md
```diff
@@ -0,0 +1,56 @@
...
+Terraform's providers are a substantial amount of code, and occasionally it
+makes sense to depend on their functionality. The most straightforward and
+obvious way to depend on a provider is to depend on the Terraform CLI, but
+occasionally it makes sense to rely on a provider in a different context.
+
+This guide lays out the supported ways to interface with and depend on
+Terraform's providers. Unless the provider explicitly states otherwise, no
+other compatibility guarantees are provided.
+
+### Importing Providers as Go Modules
```
nfagerlund:
```suggestion
## Do Not Import Providers as Go Modules
```

As written, this seems to imply that this is one of the options, but it's really an explanation for why you shouldn't.

content/source/docs/extend/best-practices/depending-on-providers.html.md
```diff
@@ -0,0 +1,56 @@
...
+### Exporting the Schema
+
+Some projects just care about the schema and resources a provider presents. As
+of Terraform 0.12, the `terraform providers schema -json` command can be used
+to export a JSON representation of the schemas for the providers used in a
+workspace.
+
+### Using the RPC Protocol
+
+For projects that actually want to drive the provider, using the [gRPC protocol](https://github.com/hashicorp/terraform/tree/master/docs/plugin-protocol)
```
nfagerlund:
```suggestion
For projects that actually want to drive the provider, the supported option is to use the [gRPC protocol](https://github.com/hashicorp/terraform/tree/master/docs/plugin-protocol)
```

content/source/docs/extend/best-practices/depending-on-providers.html.md
```diff
@@ -0,0 +1,56 @@
...
+and using the RPC calls the protocol supplies is a supported option. This
```
nfagerlund:
```suggestion
and the RPC calls the protocol supplies. This
```

content/source/docs/extend/best-practices/depending-on-providers.html.md
```diff
@@ -0,0 +1,56 @@
...
+versions after 2.0.0, providers have chosen to make their incompatibility with
+being imported into Go code explicit.
+
+If you find yourself needing to do this, perhaps one of the methods below will
+work for you, and is explicitly supported and covered under versioning policies.
+If not, please reach out and [open an issue](https://github.com/hashicorp/terraform/issues/new)
+outlining your use case, and we'll work with you to find an appropriate way
+to interface with Terraform to meet your use case.
+
+### Exporting the Schema
```
nfagerlund:
```suggestion
## Exporting the Schema
```

content/source/docs/extend/best-practices/depending-on-providers.html.md
```diff
@@ -0,0 +1,56 @@
...
+
+Some projects just care about the schema and resources a provider presents. As
+of Terraform 0.12, the `terraform providers schema -json` command can be used
+to export a JSON representation of the schemas for the providers used in a
+workspace.
+
+### Using the RPC Protocol
```
nfagerlund:
```suggestion
## Using the RPC Protocol
```

content/source/docs/extend/best-practices/depending-on-providers.html.md
```diff
@@ -0,0 +1,56 @@
...
+to interface with Terraform to meet your use case.
+
+## Exporting the Schema
+
+Some projects just care about the schema and resources a provider presents. As
+of Terraform 0.12, the `terraform providers schema -json` command can be used
+to export a JSON representation of the schemas for the providers used in a
+workspace.
+
+## Using the RPC Protocol
```
appilon: This part is great! We may or may not want to mention the setup required to interact with the provider over gRPC. A developer would need to start the provider binary with the right magic environment stuff set. Perhaps a link to our `go-plugin` lib would be helpful. I think those details can come down the line this is all excellently worded for today!

paddycarver:
@kmoe I added a note about providers importing other providers for test purposes being an exception. Does that address your concern?

PR #978, opened 2019-10-30
Add note about admin permissions to avoid confusion

acespacecase:
To avoid confusion in case the Azure DevOps Services Profile link doesn't work, I propose we add an Important box like we do for other VCS pages.

### DIFF
--- a/content/source/docs/cloud/vcs/azure-devops-services.html.md
+++ b/content/source/docs/cloud/vcs/azure-devops-services.html.md
@@ -22,7 +22,7 @@ The rest of the page explains these steps in more detail.
 
 1. Open your [Azure DevOps Services Profile](https://aex.dev.azure.com) in a browser tab; log in to your Azure DevOps Services account if necessary.
 
-    ~> **Important:** The Azure DevOps Services account you use for connecting Terraform Cloud must have admin access since creating webhooks requires admin permissions.
+    ~> **Important:** The Azure DevOps Services account you use for connecting Terraform Cloud must have admin access since creating webhooks requires admin permissions. If you're unable to load the link above, you can create a new application for the next step at one of the following links: `https://aex.dev.azure.com/app/register?mkt=en-US` or `https://app.vsaex.visualstudio.com/app/register?mkt=en-US`.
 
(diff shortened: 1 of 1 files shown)


content/source/docs/cloud/vcs/azure-devops-services.html.md
```diff
@@ -22,6 +22,8 @@ The rest of the page explains these steps in more detail.
 
 1. Open your [Azure DevOps Services Profile](https://aex.dev.azure.com) in a browser tab; log in to your Azure DevOps Services account if necessary.
 
+    ~> **Important:** The Azure DevOps Services account you use for connecting Terraform Cloud must have admin access since creating webhooks requires admin permissions.
```
nfagerlund: 1. I'm a little confused about which part of the process this applies to, and at which point I would notice a problem if something isn't right. 

    My guess is that the part where you create an application would be fine, but once all the text boxes were filled in and I clicked the "Connect organization `<NAME>`"  button, then I'd have a problem. Right? 

2. Does "admin access" fully explain what permissions my user needs on ADO? Admin access on what, specifically? (Sorry, this is something that'd be obvious to me if I'd used ADO before, but I don't know what kind of organizational unit it uses for a group of repos. Is it called an "organization," or something else?)

acespacecase: Depending on a user's permissions, navigating to `https://aex.dev.azure.com` may not load correctly, or it may load but you may not see the `Create new application` link. I've been working to try to get more specific information about what permissions a user needs to see the link correctly, but I'm struggling to get the details needed. I'm going to reword this a little to provide an alternative link, though I don't think I can add more specific details about permissions at this time.

acespacecase: I've updated the `Important` block. Unfortunately I wasn't able to get additional information regarding permissions to view the `aex.dev.azure.com` link, and I also uncovered more fun: there are two links to register an application, that also depend on the type of account/permissions you have, so @vaficionado and I have suggested including both in attempt to help a wider breadth of users.

PR #980, opened 2019-10-30
website - font and brand updates

rjspiker:
**NOTE:** The HashiCorp Marketing Team would like to launch this update in a coordinated fashion with font/brand updates to other HashiCorp product sites.  We'd like to get this approved, but hold off on merging at the moment.  Thanks.

---

This PR accomplishes the following:
- Add newly approved fonts (Gilmer, Metro Sans, Deja Vu) and remove deprecated fonts where no longer needed
- Global typography updates (font-size, line-height, etc), adding .g-type classes shared across sites
- A few spacing updates
- Color updates to match color scheme with other HashiCorp sites
- Button style updates

cc: @kylemac @mwickett


nfagerlund:
@rjspiker Could you please include some screenshots in this PR to help simplify reviews?

nfagerlund:
I think this is in decent shape, and I do like the new fonts. Two things stuck out to me:

- Text in the filter input (in the docs nav sidebar) is too small to be legible at .625rem. I tried .8rem and it looks closer to right: 

    image

    .750rem is almost acceptable but still noticeably tiny.
- Our syntax highlighting styles some constructs as bold text, and bold Deja Vu is quite nasty on non-retina displays (at least for macOS, I don't have a windows or linux browser in easy reach rn). http://localhost:4567/docs/configuration/syntax-json.html (in a local preview) has some good examples. Check out the awesome blurring on those double-quotes:

    image

    Has this come up yet in any of the other places you're implementing Deja Vu? How'd you solve it there? We rely pretty heavily on code examples, and this'd be a pretty noticeable regression, so I'd like to avoid it if possible. That might mean creating a modified Rouge theme; right now it looks like we're using this one https://github.com/rouge-ruby/rouge/blob/master/lib/rouge/themes/github.rb

rjspiker:
Thanks for the feedback @nfagerlund.  I'll adjust the font size for the filter input.  I'll also bring your concerns with Deja Vu to the rest of the team for discussion and get back to you.

Appreciate the feedback!

zchsh:
@nfagerlund Great feedback on those crusty quotes. I'm not sure how essential it is to the syntax highlighting generally, but removing the `bold` weight might be an option:

## Before

dejavusansmono-bold

## After

dejavusansmono

zchsh:
Actually on the font note, @nfagerlund curious to get your thoughts on a "system monospace" stack, like GitHub uses ( something like `font-family: "SF Mono", Consolas, Liberation Mono, Menlo, monospace;` ). Maybe this would help with weird rendering issues for bolder weights?

All i have for visual here is SF Mono, retina screen, probably not very helpful for this issue, but here it is just in case:

sfmono


_Note: I haven't got a full picture on whether we'd be open to fully adopting this idea on the marketing design side... I think we would, but I think we'd likely also want some kind of clear monospace font recommendation for more static mediums like print, slide decks, etc._

nfagerlund:
Hmm, yeah, just tried it myself and SF Mono is noticeably more legible on non-retina. 

I also spent a few minutes playing around with the alternate built-in Rouge themes, and don't really love most of them (especially when applied to our site), but I don't love the current GitHub-emulating one we use either. (It's a lil clashy with our purple.) But supposedly any Pygments stylesheet will work with the markup, so if anyone has run across one that would fit us nicely, we could try that. 

That's all a bit of a rabbit hole. I like SF Mono better on non-retina, but given that switching to Deja Vu wouldn't really be a regression from the current blurriness, it shouldn't block merge on this if we run out of time.

rjspiker:
I've updated the PR to address the concerns with the filter font size.  After discussion with the Marketing team, it sounds like we're going to revisit the monospace font discussion, but in the meantime we'd like to move forward using Deja Vu.

PR #982, opened 2019-10-31
Add link for the first Azure DevOps application

mikaelkrief:
In the step 1 I add the link for the first Azure DevOps application


nfagerlund:
Thanks for submitting this! This was also addressed by #978 — we ended up merging that one because it eventually included _another_ link you might need to try, depending on how your account is set up. 😅

cwoolum:
This was the link I needed to use to actually get the form to load. The other link didn't work for me.

PR #985, opened 2019-11-01
Sensitive variables can also end up in state files

rberlind:
Document that sensitive variables can end up in both logs and state files.
Also, point out that environment variables cannot end up in state files so are preferred when using them is possible.
My parenthetical statement that an environment variable can show up in a log file if TF_LOG is set to TRACE is based on personal experience.  It is possible that this could also happen with TF_LOG set to DEBUG, but I have not tested for that.


rberlind:
Thanks @nfagerlund . Agreed.

PR #990, opened 2019-11-05
Add note about instance profiles and build workers for TFE

bnferguson:
In Terraform Enterprise customers can use alternative Terraform worker images to do their Terraform runs. These short lived docker containers will pick up the instance profile of the machine they're running on unless env vars are passed in overriding them. 

Currently this isn't documented anywhere, and can be a bit confusing if you're not expecting it. This PR attempts to fix that fact.

### DIFF
--- a/content/source/docs/enterprise/install/installer.html.md
+++ b/content/source/docs/enterprise/install/installer.html.md
@@ -132,6 +132,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     sudo unzip daemontools git-core ssh wget curl psmisc iproute2 openssh-client redis-tools netcat-openbsd
 ```
 
+If an Instance Profile is used in AWS the worker image will use those permissions for Terraform operations unless other credentials are passed via environment variables.
+
 ### Image initialization
 To run initialization commands in your image during runtime, create a script at `/usr/local/bin/init_custom_worker.sh` and make it executable. This script, and all commands it invokes, will be executed before TFE runs `terraform init`.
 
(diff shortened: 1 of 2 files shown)


content/source/docs/enterprise/install/installer.html.md
```diff
@@ -132,6 +132,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     sudo unzip daemontools git-core ssh wget curl psmisc iproute2 openssh-client redis-tools netcat-openbsd
 ```
 
+If an Instance Profile is used in AWS the worker image will use those permissions for Terraform operations unless other credentials are passed via environment variables.
```
nfagerlund:
```suggestion
-> **Note:** If Terraform Enterprise is running in AWS with an instance profile, the worker containers will use that profile's permissions for Terraform operations unless other credentials are passed via environment variables.
```

bnferguson:
> Question: does this only apply to custom worker images, or does it also affect a default TFE install?

This is a great question. And glad you asked it because after asking and confirming, it does affect the default TFE build worker too. So this note is not just on the custom build worker (that's just what led to the ticket). Now I'm not super sure where best to put this.

nfagerlund:
Yeesh, that IS kind of a tough one. 😅 

My current best option is over here, I think:

https://www.terraform.io/docs/enterprise/before-installing/index.html#linux-instance

Hmm, instead of describing what to do, hold on, I'll whip up a commit...

nfagerlund:
@bnferguson Ok, how's that? Also, I added @erindatkinson and @evanphx to double-check me here (sorry, and thanks for your help in the channel earlier!).

nfagerlund:
Ok, in that case I'm gonna go for it! 

I'm not 100% convinced we managed to get this into a place where people will see it when they need it, but we can iterate later; in the meantime at least people can see it when they're thinking about how to configure their linux instances.

bnferguson:
@nfagerlund 🎉 This is excellent. Thanks for driving this one home!

PR #994, opened 2019-11-07
website - add -moz-osx-font-smoothing for smooth firefox osx fonts

rjspiker:
Adding `-moz-osx-font-smoothing: grayscale;` to `<body>`  in `global-styles` to smooth out fonts in Firefox on OSX.

Note that we are already using `-webkit-font-smoothing: antialiased;`.

#### Before

Screen Shot 2019-11-07 at 12 12 10 PM

---

### After

Screen Shot 2019-11-07 at 12 12 04 PM

### DIFF
--- a/content/source/assets/stylesheets/_global.scss
+++ b/content/source/assets/stylesheets/_global.scss
@@ -7,6 +7,7 @@ html {
 }
 
 body {
+  -moz-osx-font-smoothing: grayscale;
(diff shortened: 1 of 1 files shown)


rjspiker:
Thanks @nfagerlund.  And just to clarify, we're added this based on some feedback provided by some of our engineers who were struggling to read the fonts because they weren't rendering quite right in Firefox in OSX. 😄

PR #995, opened 2019-11-07
Fix un-minifiable javascript on terraform.io

nfagerlund:
Our semi-abandoned Middleman-based site builder uses Uglifier 2.x to minify
JavaScript, and it barfs on any es2015 syntax. Uglifier runs on JS files after
Sprockets processes them. If a Sprockets directive pulls an es2015 file into
application.js, it can't be compressed and will be served raw.

I could solve this by rewriting the lambdas and string templates in home-hero
into es5, but instead let's do a two-line fix that doesn't impact the
maintainability of that JS.

This still emits an error during builds, but the result is fine.

### DIFF
--- a/content/source/assets/javascripts/application.js
+++ b/content/source/assets/javascripts/application.js
@@ -9,7 +9,7 @@
 
 
 //= require terraform-overview/vendor-scripts/object-fit-images.min.js
-//= require terraform-overview/home-hero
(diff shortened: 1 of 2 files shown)


nfagerlund:
Just for posterity's sake: yeah, I absolutely consider this tech debt. 😅 But it seems better to float a small loan for this one than to try and integrate Babel into an abandoned toolchain, given that we're transitioning to a different toolchain soon.

PR #1001, opened 2019-11-11
Clarify token_period on AppRole

evanphx:
This clarifies that any token returned by the external vault's approle setup should be a periodic token with a renew period of 24 hours. This makes the tokens invulnerable to maximum TTLs.


evanphx:
`24h` matches what we do with internal vault and given folks have had issues with tokens expiring, felt best to error on the side of caution with the recommendation here.

nfagerlund:
In that case I'm good!

PR #1002, opened 2019-11-12
Fix additional checks for moved Azure auth guides

nfagerlund:
These already have redirects in redirects.txt, but the additional check
for common external links was causing spurious Travis failures.


sudomateo:
Thank you for this @nfagerlund!

PR #986, opened 2019-11-01
adding note about requiring 3-character usernames

mikegreen:
Related to
https://app.asana.com/0/1143232772343007/1146683609014113
This adds language to configure Okta with a scheme that will ensure >=3 character usernames

PR #1003, opened 2019-11-12
Revert "adding note about requiring 3-character usernames (#986)"

nfagerlund:
This reverts commit da7d25d7b56820531ddfe87052192d76550809fc, because that
username restriction has been removed, per @dylanegan in #986. 

Filing this as a draft, because I still have questions! 

- What version of TFE is it fixed in? Is it out yet? 
- Instead of just removing, should we change it to a note about the required TFE version for using Okta with non-email names?


dylanegan:
@nfagerlund I believe it'll make its way into the December release. /cc @alisdair

nfagerlund:
@dylanegan or @alisdair or @evanphx (or anyone really), did this TFE fix end up shipping yet? And if so, what's the minimum version?

dylanegan:
@nfagerlund I believe this was in v201912-1.

nfagerlund:
@dylanegan Thanks! Okay, how's this?

PR #1011, opened 2019-11-14
Update cloud how it works image

mwickett:
As requested by @apdavanzo - I dropped the largest res image here as it's massive and overkill from a size standpoint for how we're displaying it.


mwickett:
Yeah, you've got it. Sorry I made that harder than it needed to be, and I should have included details on what changed in my description. Next time!

PR #1015, opened 2019-11-15
Fix sidebar jank: wandering chevrons and cut-off titles

nfagerlund:
The commit messages have explanations for the fixes. Here's some screenshots: 

### Before:

This shows both the weird wandering chevron, and some cut-off titles. 

Screen Shot 2019-11-14 at 4 34 02 PM

This adds 

### After:

Safari: 

Screen Shot 2019-11-14 at 4 30 42 PM

Firefox (with slightly wider window breakpoint):

Screen Shot 2019-11-14 at 4 30 50 PM

### Misc:

Here's a before and after that adds a background color, to make it more obvious what's going on with those chevrons: 

Screenshot_2019-11-14 Google google_composer_image_versions - Terraform by HashiCorp(1)

Screenshot_2019-11-14 Google google_composer_image_versions - Terraform by HashiCorp(2)

Here's a shot with grid layout indicators turned on, so you can see how it gets laid out both with and without a subnav. (Although, one thing I really like about the `grid-template-areas` property is that you can "see" the grid right there in the CSS, so maybe this shot wasn't quite necessary.)

image

### DIFF
--- a/content/source/assets/stylesheets/_docs.scss
+++ b/content/source/assets/stylesheets/_docs.scss
@@ -1,6 +1,11 @@
 #docs-sidebar {
   margin-bottom: 30px;
-  overflow: hidden;
+
+  // fallback styles for gridless browsers
+  overflow-x: hidden;
+  @supports (display: grid) {
+    overflow-x: visible;
+  }
 
   // Re-using the .nav class name on lists means we inherit some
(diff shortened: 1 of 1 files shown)


nfagerlund:
Well, I got an experimental OK, so I'm going to go ahead with this. Can revisit if it messes anything up!

PR #1017, opened 2019-11-15
fix(doc): example indentation

JordanForeman:
The indentation for the examples when using `hashicorp/terraform-github-actions` was incorrect. Followed exactly, users would see an error like so:

```bash
- Your workflow file was invalid: The pipeline is not valid. .github/workflows/workflow.yml (Line: xx, Col: x): Unexpected value 'args'
```

This PR fixes this inaccuracy

### DIFF
--- a/content/source/docs/github-actions/common-tasks/variables.html.md
+++ b/content/source/docs/github-actions/common-tasks/variables.html.md
@@ -29,7 +29,7 @@ jobs:
           tf_actions_subcommand: 'init'
           tf_actions_working_dir: '.'
           tf_actions_comment: true
-        args: '-var="env=dev"'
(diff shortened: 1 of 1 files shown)


nfagerlund:
Oh ho, sure enough: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepswithargs Thank you for catching this! 

This isn't the only page affected. Instead of making more work for you, I'm gonna close this PR, add a few more commits on top of yours, and open a new final PR.

sudomateo:
Thank you Nick! I must have mistyped this when I added support for `args` in the code.

PR #1032, opened 2019-11-25
website: Invert check for Segment ID

mwickett:
With the switch from TeamCity to CircleCI, it looks like we're missing the `DEPLOY_ENV` environmental variable - I can't verify as I'm unable to access the CircleCI config for the project. We've been pushing production analytics to our staging destination in Segment since ~Nov 6 😢 

This inverts the check so that we default to production. The `make` file sets `DEPLOY_ENV` to `development`, so we should be covered when running this locally.

Ultimately it's better to leak a bit of dev activity analytics into production, then have no production analytics at all.

### DIFF
--- a/content/middleman_helpers.rb
+++ b/content/middleman_helpers.rb
@@ -2,10 +2,10 @@ module Helpers
   # Returns a segment tracking ID such that local development is not
   # tracked to production systems.
   def segmentId()
-    if (ENV['DEPLOY_ENV'] == 'production')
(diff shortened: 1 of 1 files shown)


mwickett:
@nfagerlund No worries - these things happen! Adjusting this default will help prevent it from happening. Might as well set that env var in the Circle job, yeah. I took a look for it, but couldn't track it down. Thank you!

nfagerlund:
Added!

PR #1037, opened 2019-11-26
Include section of about capacity management

evanphx:
I'm unsure if this the right place for this content, but it seems relevant to the overall architecture thus inclusion on that page.

### DIFF
--- a/content/source/docs/enterprise/before-installing/cluster-architecture.html.md
+++ b/content/source/docs/enterprise/before-installing/cluster-architecture.html.md
@@ -33,7 +33,7 @@ Run concurrency is usually limited by memory. By default, each Terraform run is 
 
 To calculate how big a cluster should be, first decide how many concurrent Terraform runs you require. For instance, if you need to be able to execute 100 runs simultaneously, then the cluster needs at least 50GB of RAM available for runs. Because the services that run the Terraform Enterprise application run also share that memory, we suggest adding 16GB to the number. So in this case, the cluster would require 66GB of RAM. The most common way to achieve at least 66GB of cluster RAM is a pool of 5 secondary instances and 3 primary instances, all with 16GB of RAM.
 
-The amount of memory used per Terraform run is configurable in the Management UI and changing that amount will alter the maximum concurrent capacity of the cluster.
+The amount of memory used per Terraform run is configurable in the installer dashboard (`https://<TFE HOSTNAME>:8800`), and changing that amount will alter the maximum concurrent capacity of the cluster.
 
 ## Architecture Diagrams
 
--- a/content/source/docs/enterprise/before-installing/cluster-architecture.html.md
+++ b/content/source/docs/enterprise/before-installing/cluster-architecture.html.md
@@ -27,7 +27,7 @@ A Terraform Enterprise cluster relies on a load balancer to direct traffic to ap
 
 ## Capacity Management
 
-Important aspect to the cluster is how resources are used to execute Terraform runs because this has a direct impact upon how much work a cluster can to concurrently.
+There is a direct relationship between a cluster's total resources and the amount of work it can do concurrently.
 
 By default, each Terraform run is allocated 512MB of RAM which has the majority of the impact upon the ability to schedule new Terraform runs.
 
@@ -58,4 +58,3 @@ Clustered deployment is designed for use with external data services. These incl
 These services are outside the scope of the module that deploys Terraform Enterprise; the operator is expected to provision and configure the services prior to installation, and provide access information and credentials as inputs to the module.
 
 The clustered deployment modules can also deploy a temporary demo of Terraform Enterprise that does not require external data services.
-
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/before-installing/cluster-architecture.html.md
```diff
@@ -25,6 +25,16 @@ Clustered deployment relies on a Terraform module to provision infrastructure. C
...
 
+## Capacity Management
+
+Important aspect to the cluster is how resources are used to execute Terraform runs because this has a direct impact upon how much work a cluster can to concurrently.
+
+By default, each Terraform run is allocated 512MB of RAM which has the majority of the impact upon the ability to schedule new Terraform runs.
+
+To calculate how big a cluster should be, a customer needs to first decide how many concurrent Terraform runs they require. For instance, if they need to be able to execute 100 runs simultaniously, then the cluster needs to have at least 50GB of RAM available for runs. Because the services that run the Terraform Enterprise application run also share that memory, we suggest adding 16GB to the number. So in this case, the cluster would require 66GB of RAM. **The most common way to achieve at least 66GB of cluster RAM is a pool of 5 secondary instances and 3 primary instances, all with 16GB of RAM**.
+
+The amount of memory used per Terraform run is configurable in the Management UI and changing that amount will alter the maximum concurrent capacity of the cluster.
```
nfagerlund: By "the Management UI," do you mean `:8080/dashboard`? Or is this something new under the site admin pages?

Also, I assume this is an experimental process whose outcome depends on the size/complexity of your configs. What's the failure signal that tells you you went too low and need to bump the RAM back up a bit?

content/source/docs/enterprise/before-installing/cluster-architecture.html.md
```diff
@@ -25,6 +25,16 @@ Clustered deployment relies on a Terraform module to provision infrastructure. C
 
 A Terraform Enterprise cluster relies on a load balancer to direct traffic to application instances.
 
+## Capacity Management
+
+Important aspect to the cluster is how resources are used to execute Terraform runs because this has a direct impact upon how much work a cluster can to concurrently.
```
nfagerlund:
```suggestion
There is a direct relationship between a cluster's total resources and the amount of work it can do concurrently.
```

content/source/docs/enterprise/before-installing/cluster-architecture.html.md
```diff
@@ -25,6 +25,16 @@ Clustered deployment relies on a Terraform module to provision infrastructure. C
...
+
+By default, each Terraform run is allocated 512MB of RAM which has the majority of the impact upon the ability to schedule new Terraform runs.
```
nfagerlund:
```suggestion
Run concurrency is usually limited by memory. By default, each Terraform run is allocated 512MB of RAM, and a cluster can schedule new runs as long as one of its instances has enough memory available. 
```

content/source/docs/enterprise/before-installing/cluster-architecture.html.md
```diff
@@ -25,6 +25,16 @@ Clustered deployment relies on a Terraform module to provision infrastructure. C
...
+
+To calculate how big a cluster should be, a customer needs to first decide how many concurrent Terraform runs they require. For instance, if they need to be able to execute 100 runs simultaniously, then the cluster needs to have at least 50GB of RAM available for runs. Because the services that run the Terraform Enterprise application run also share that memory, we suggest adding 16GB to the number. So in this case, the cluster would require 66GB of RAM. **The most common way to achieve at least 66GB of cluster RAM is a pool of 5 secondary instances and 3 primary instances, all with 16GB of RAM**.
```
nfagerlund:
```suggestion
To calculate how big a cluster should be, first decide how many concurrent Terraform runs you require. For instance, if you need to be able to execute 100 runs simultaneously, then the cluster needs at least 50GB of RAM available for runs. Because the services that run the Terraform Enterprise application run also share that memory, we suggest adding 16GB to the number. So in this case, the cluster would require 66GB of RAM. The most common way to achieve at least 66GB of cluster RAM is a pool of 5 secondary instances and 3 primary instances, all with 16GB of RAM.
```

content/source/docs/enterprise/before-installing/cluster-architecture.html.md
evanphx: Yes, I mean the :8800/dashboard. 

The failure signal is that few jobs are running concurrently and ergo take longer to run.

nfagerlund: Which reminds me: we need some documentation for that page. 

Well, in the meantime, let's make this look like our other references to the dashboard:

```suggestion
The amount of memory used per Terraform run is configurable in the installer dashboard (`https://<TFE HOSTNAME>:8800`), and changing that amount will alter the maximum concurrent capacity of the cluster.
```

PR #1046, opened 2019-12-04
Update writing-custom-providers.html.md

joel-g:
changes "driver_options" to "DriverConfig" in order to match JSON example


nfagerlund:
I believe that's not quite right: the example is mapping `DriverConfig` and the surrounding JSON values into a different schema, to present a more Terraform-like resource interface. 

If you look a little further down, you can see that we're actually pulling `DriverConfig` apart, mapping `Options` property to `driver_options` and its `Name` property to the `driver_name`:

```go
      if v.VolumeOptions.DriverConfig != nil {
        volumeOptions["driver_name"] = v.VolumeOptions.DriverConfig.Name
        volumeOptionsItem["driver_options"] = mapStringStringToMapStringInterface(v.VolumeOptions.DriverConfig.Options)
      }
```

Feel free to re-open and explain further if I've misunderstood something!

PR #1049, opened 2019-12-05
[DOC] Workspace naming

bazi:
As it turns out, workspace names can contain . (dot) as well.

I think documentation should be complete regarding workspace naming. Currently, dot symbol is not mentioned in docs as allowed, whereas Terraform allows workspace names to contain dot symbol. This is important for teams that use workspace names in other parts of their projects.

Tested with v0.11.13 and v0.12.17.


nfagerlund:
Hi! 

Terraform Cloud does _not_ allow periods in workspace names. Terraform CLI might allow them, but those workspace names could not be used with Terraform Cloud. 

I agree it would be better for Terraform Cloud users if the requirements were identical! However, Terraform Cloud is not the only state backend that allows multiple workspaces, and changing Terraform's naming requirements might break existing workflows with other backends. So, possibly desirable, but it'd have to involve a major version boundary and a deprecation period.

PR #1062, opened 2019-12-12
don’t use .html ext in links to learn

kylemac:
currently this throws off our analytics reporting as GA collects these two as separate paths. we will put in logic to merge them in GA but regardless, excluding .html is still a best practice for these links.


nfagerlund:
Thanks for pointing that out and catching these! I'll peek around and see if there are others we need to grab in other repos. (There was only one in the terraform core repo, PR incoming in a sec.)

PR #1063, opened 2019-12-13
Update tfrun import documentation

hcrhall:
This adds a documentation update for the tfrun import. It includes descriptions of the new value types that have been exposed by namespace and also includes a Sentinel example for each value type.

### DIFF
--- a/content/source/docs/cloud/sentinel/import/tfrun.html.md
+++ b/content/source/docs/cloud/sentinel/import/tfrun.html.md
@@ -9,7 +9,7 @@ description: |-
 
 The `tfrun` import provides access to data associated with a [Terraform run][run-glossary].
 
-This import currently consist of a `root` namespace, as well as namespaces for the `organization`, `workspace` and `cost-estimate`. Each namespace provides static data regarding the Terraform Cloud application that can then be consumed by Sentinel during a policy evaluation.
+This import currently consists of run attributes, as well as namespaces for the `organization`, `workspace` and `cost-estimate`. Each namespace provides static data regarding the Terraform Cloud application that can then be consumed by Sentinel during a policy evaluation.
 
 
 ```
@@ -55,13 +55,15 @@ The **root namespace** contains data associated with the current run.
 
 * **Value Type:** String.
 
-The `created_at` value within the [root namespace](#root-namespace) specifies the time that the run was created.
+The `created_at` value within the [root namespace](#root-namespace) specifies the time that the run was created. The timestamp returned follows the format outlined in [RFC3339](https://tools.ietf.org/html/rfc3339).
+
+Users can use the `time` import to [load](https://docs.hashicorp.com/sentinel/imports/time/#timeloadtimeish) a run timestamp and create a new timespace from the specicied value. See the `time` import [documentation](https://docs.hashicorp.com/sentinel/imports/time/#import-time) for available actions that can be performed on timespaces.
 
 ### Value: `message`
 
 * **Value Type:** String.
 
-The `message` value within the [root namespace](#root-namespace) specifies the message that is associated with the run.
+Specifies the message that is associated with the run.
 
 The default value is *"Queued manually via the Terraform Enterprise API"*.
 
@@ -69,19 +71,19 @@ The default value is *"Queued manually via the Terraform Enterprise API"*.
 
 * **Value Type:** Boolean.
 
-The `speculative` value within the [root namespace](#root-namespace) specifies whether the plan associated with the run is a [speculative plan](../../run/index.html#speculative-plans) only.
+Specifies whether the plan associated with the run is a [speculative plan](../../run/index.html#speculative-plans) only.
 
 ### Value: `is_destroy`
 
 * **Value Type:** Boolean.
 
-The `is_destroy` value within the [root namespace](#root-namespace) specifies whether the run will perform a destroy operation.
+Specifies if the plan is a destroy plan, which will destroy all provisioned resources.
 
 ### Value: `variables`
 
 * **Value Type:** A string-keyed map of values.
 
-The `variables` value within the [root namespace](#root-namespace) provides the names of the variables that are configured within the run and the [sensitivity](../../workspaces/variables.html#sensitive-values) state of the value.
+Provides the names of the variables that are configured within the run and the [sensitivity](../../workspaces/variables.html#sensitive-values) state of the value.
 
 ```
 variables (map of keys)
@@ -97,7 +99,7 @@ The **organization namespace** contains data associated with the current run's T
 
 * **Value Type:** String.
 
-The `name` value within the [organization namespace](#organization-namespace) specifies the name assigned to the Terraform Cloud organization.
+Specifies the name assigned to the Terraform Cloud organization.
 
 ## Namespace: workspace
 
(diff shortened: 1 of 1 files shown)


content/source/docs/cloud/sentinel/import/tfrun.html.md
```diff
@@ -42,6 +47,58 @@ to only enforce the policy on non-development workspaces is more appropriate.
 [run-glossary]: /docs/glossary.html#run
 [workspace-glossary]: /docs/glossary.html#workspace
 
+## Namespace: root
```
hcrhall: Using `root` in the docs but I am wondering if using `tfrun` is better as this is called out in the diagram above. Thoughts?

glenngillen: I'm going to ping @nfagerlund incase he is aware of some prior art in our docs as I'd prioritise consistency above most else.

I _think_ (though I'd be hesitant to stake any money on this) that the semantically correct thing here isn't a root namespace. I've got a series of attributes, on the root element. So maybe we can ditch the logical grouping of them in the docs (given that grouping doesn't really exist in the object) and just list the attributes we expose? Then the two namespaces can follow after that.

content/source/docs/cloud/sentinel/import/tfrun.html.md
```diff
@@ -42,6 +47,58 @@ to only enforce the policy on non-development workspaces is more appropriate.
...
+
+The **root namespace** contains data associated with the current run.
+
+### Value: `created_at`
+
+* **Value Type:** String.
+
+The `created_at` value within the [root namespace](#root-namespace) specifies the time that the run was created.
```
glenngillen: If this is a string I assume it's following some particular datetime format, could we be explicit about what that is and maybe provide an example to avoid any confusion?

content/source/docs/cloud/sentinel/import/tfrun.html.md
```diff
@@ -42,6 +47,58 @@ to only enforce the policy on non-development workspaces is more appropriate.
...
+
+* **Value Type:** Boolean.
+
+The `speculative` value within the [root namespace](#root-namespace) specifies whether the plan associated with the run is a [speculative plan](../../run/index.html#speculative-plans) only.
+
+### Value: `is_destroy`
+
+* **Value Type:** Boolean.
+
+The `is_destroy` value within the [root namespace](#root-namespace) specifies whether the run will perform a destroy operation.
```
glenngillen: Is this true if _any_ resource is being destroyed? Or only if everything is being destroyed as part of some cleanup/workspace destruction? I originally assumed it was the latter but the copy here makes me think it's if anything in the plan is being destroyed.

content/source/docs/cloud/sentinel/import/tfrun.html.md
```diff
@@ -42,6 +47,58 @@ to only enforce the policy on non-development workspaces is more appropriate.
...
+
+### Value: `variables`
```
glenngillen: I'm a little torn on this one. The nested nature of it makes it seem less like an attribute and more like a namespace. Are we clear on when we think about things as a namespace vs a nested set of attributes?

content/source/docs/cloud/sentinel/import/tfrun.html.md
```diff
@@ -42,6 +47,58 @@ to only enforce the policy on non-development workspaces is more appropriate.
...
+
+## Namespace: organization
+
+The **organization namespace** contains data associated with the current run's Terraform Cloud [organization](../../users-teams-organizations/organizations.html).
+
+### Value: `name`
+
+* **Value Type:** String.
+
+The `name` value within the [organization namespace](#organization-namespace) specifies the name assigned to the Terraform Cloud organization.
```
glenngillen: Wondering now if all of the "value within the namespace" statements are kinda redundant given the hierarchy within the document. Another I'll defer to @nfagerlund on though.

content/source/docs/cloud/sentinel/import/tfrun.html.md
```diff
@@ -9,13 +9,18 @@ description: |-
 
 The `tfrun` import provides access to data associated with a [Terraform run][run-glossary].
 
-This import currently includes two namespaces — `workspace`, which provides Sentinel policies with
-static data regarding the Terraform Cloud [workspace][workspace-glossary] associated with the run,
-and `cost_estimate`, which provides Sentinel policies with static data regarding the Terraform Cloud
-cost estimate associated with the run.
+This import currently consist of a `root` namespace, as well as namespaces for the `organization`, `workspace` and `cost-estimate`. Each namespace provides static data regarding the Terraform Cloud application that can then be consumed by Sentinel during a policy evaluation.
```
dylanegan: Tying into https://github.com/hashicorp/terraform-website/pull/1063/files#r357488915.

I think something along the lines of `consists of run attributes, as well as...`.

content/source/docs/cloud/sentinel/import/tfrun.html.md
hcrhall: @glenngillen it is an rfc3339 compliant format. I'll call this out in the docs and I will provide an example of how we can use the time import to load the value and act on it

content/source/docs/cloud/sentinel/import/tfrun.html.md
hcrhall: @glenngillen good catch,. I use the same description that we use in the API docs which indicates that it is the latter example you have called out

content/source/docs/cloud/sentinel/import/tfrun.html.md
hcrhall: @dylanegan definitely a better approach considering the feedback thus far.

content/source/docs/cloud/sentinel/import/tfrun.html.md
hcrhall: @glenngillen I originally had variables listed as a separate namespace but @chrisarcand commented in TF-199 and requested that I change it to an attribute instead.

content/source/docs/cloud/sentinel/import/tfrun.html.md
chrisarcand: "Root" is 'correct' https://www.terraform.io/docs/cloud/sentinel/import/tfplan.html#namespace-root

I think the import docs in general could use some reformatting/naming (and that 'root' is blah) but would agree that consistency should be prioritized highly.

content/source/docs/cloud/sentinel/import/tfrun.html.md
dylanegan: https://github.com/hashicorp/terraform-website/pull/1063#discussion_r358310804

It looks like `root` is correct.

content/source/docs/cloud/sentinel/import/tfrun.html.md
hcrhall: Thanks for confirming @chrisarcand

content/source/docs/cloud/sentinel/import/tfrun.html.md
hcrhall: @glenngillen pushed some changes, please review and resolve conversation if happy with changes. Thanks.

content/source/docs/cloud/sentinel/import/tfrun.html.md
hcrhall: @glenngillen pushed some changes, please review and resolve conversation if happy with changes. Thanks.

content/source/docs/cloud/sentinel/import/tfrun.html.md
hcrhall: @glenngillen pushed some changes, please review and resolve conversation if happy with changes. Thanks.

PR #1077, opened 2020-01-10
content: Add redirects for incorrectly named Terraform AWS Provider files

bflad:
The correctly named files will be published first via `stable-website` branch, before merging these redirects.


bflad:
Upstream https://github.com/terraform-providers/terraform-provider-aws/pull/11552 merged, pushed to `stable-website`, ran `make sync` and added `ext/provider/aws` so these will launch together.

bflad:
The broken link check failure is the same as on `master` and not introduced by this pull request -- merging!

PR #1083, opened 2020-01-14
updated css so headers with two lines dont look weird

im2nguyen:
- update css so headers with two lines don't look weird
- assuming that the line-height is 44px, since it's defined in the css

current:
image

proposed: 
image

single line:
image

### DIFF
--- a/content/source/assets/stylesheets/_docs.scss
+++ b/content/source/assets/stylesheets/_docs.scss
@@ -37,7 +37,7 @@
 
     margin-top: 20px; // fallback for non-grid browsers
     @supports (display: grid) {
-      margin-top: 0;
+      // margin-top: 0; // commented out to account for two row headers
     }
 
     h3, h4 {
--- a/content/source/assets/stylesheets/_docs.scss
+++ b/content/source/assets/stylesheets/_docs.scss
@@ -35,10 +35,7 @@
     grid-template-rows: minmax(90px, auto) 32px; // match #inner-header-grid
     margin-bottom: 16px;
 
-    margin-top: 20px; // fallback for non-grid browsers
-    @supports (display: grid) {
-      // margin-top: 0; // commented out to account for two row headers
-    }
+    margin-top: 20px;
 
     h3, h4 {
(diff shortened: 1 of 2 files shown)


im2nguyen:
@nfagerlund, I commented out margin-top: 0 if grid is supported for both the sidebar and the main content, which produces the desired result (everything is just shifted down 20px). please let me know if this works. if not, the other route is to change the h1 font size to 2.25 rem, which also produces the desired result

image

image

content/Gemfile.lock
```diff
@@ -15,7 +15,7 @@ GEM
     capybara (2.4.4)
       mime-types (>= 1.16)
       nokogiri (>= 1.3.3)
-      rack (>= 1.0.0)
+      rack (>= 1.6.12)
```
nfagerlund: Please revert all the gemfile changes -- those are fairly disruptive and require building and shipping a new Docker container, so they're not something that can be done as an aside in an unrelated PR.

content/source/assets/stylesheets/_docs.scss
```diff
@@ -37,7 +37,7 @@
 
     margin-top: 20px; // fallback for non-grid browsers
     @supports (display: grid) {
-      margin-top: 0;
+      // margin-top: 0; // commented out to account for two row headers
```
nfagerlund: The right way to do this is to remove these whole `@supports` query blocks, and also remove the "only for non-grid" comments above. But I approve the general approach of just leaving some extra space. 👍🏼

im2nguyen: yep, i would have removed all the unnecessary code, but wasn't sure if we were going to need it in the future, seeing how this is just a temp fix 

```
 // fallback for non-grid browsers
    @supports (display: grid) {
      // margin-top: 0; // commented out to account for two row headers
    }
```

content/Gemfile.lock
im2nguyen: sounds good, new commit will revert this change. will be happy to pair with you when we do choose to update the gem!

PR #1084, opened 2020-01-15
Add note about ssh key format

mengesb:
- SSH key formats can differ, and while most implementations of `ssh_keygen` state that they default to RFC4716/SSH2 formats, most Linux systems output a compatible PEM format. MacOS does output an RFC4716/SSH2 format, identified by the header `-----BEGIN OPENSSH PRIVATE KEY-----` versus the PEM format which is `-----BEGIN RSA PRIVATE KEY-----`

Signed-off-by: Brian Menges <mengesb@gmail.com>

## Description

Adding note about ssh key generation formats and specifying supported key format


nfagerlund:
Thank you for catching this! However, I'm going to close this PR, because another PR was filed around the same time that also catches a few extra instances of this info: https://github.com/hashicorp/terraform-website/pull/1082 -- our folks are doing some additional technical verification on that, and then we'll get it merged.

PR #1089, opened 2020-01-17
Add registry links for HashiCorp providers

nfagerlund:
This commit adds links to the registry pages for all of the HashiCorp providers
that are now available in the Terraform Registry. Omitting the version number
ensures that these links redirect to the current versions.

This uses the capability added in https://github.com/hashicorp/terraform-website/pull/945, so we can add these links without having to edit the actual docs for the affected providers. 

Links are only added to the front page of each provider (`/docs/providers/:name/index.html`).

List of affected providers: 

- alicloud
- archive
- aws
- azuread
- azurerm
- azurestack
- ciscoasa
- consul
- dns
- external
- github
- google
- helm
- http
- kubernetes
- local
- nomad
- null
- oci
- opc
- oraclepaas
- random
- template
- terraform
- tfe
- tls
- vault
- vsphere

image

### DIFF
--- a/content/source/layouts/inner.erb
+++ b/content/source/layouts/inner.erb
@@ -18,7 +18,34 @@
 
     <%
       external_docs_links = {
-        # '/docs/providers/aws/index.html' => 'https://registry.terraform.io/path/to/provider',
+        '/docs/providers/alicloud/index.html'       => 'https://registry.terraform.io/providers/hashicorp/alicloud/',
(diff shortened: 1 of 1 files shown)


nfagerlund:
@phinze I'm headed to lunch now, but if this matches what you expect, you can feel free to merge it!

phinze:
LGTM!

PR #1095, opened 2020-01-24
cloud: add "technology preview" notice on the v2 import pages

vancluever:
## Description

While implemented, supported, and live on production, the Sentinel v2
imports may be undergoing some changes in the next few weeks that would
entail changing the API. We are hence releasing these for a period under
a tech preview before declaring them stable. Hence, there needs to be a
notice indicating as such.

There's also been a small update to the blue info boxes regarding
wording on what the Sentinel imports are designed for.


vancluever:
Thanks @nfagerlund!

PR #1096, opened 2020-01-28
Update oauth-clients.html.md

susiesolis:
I think adding some spaces to lines 165-167 or highlighting this text will make it easier for people to see that this API token is required when creating an oauth client. While this information was present, I had a customer ticket run 2 months because she missed this portion of info on the site.


nfagerlund:
Totally worth a try.

PR #1100, opened 2020-01-29
Redirect digitalocean index.html to do

justincampbell:
To enable linking based only on the provider name


nfagerlund:
Are you still interested in having this? And if so, are there other providers we ought to hit?

justincampbell:
@nfagerlund Yes please! I think it would improve the experience when we link from somewhere and only know the provider name. I ran through all of the submodules and didn't see any others that diverged from the repo name.

nfagerlund:
Yeah, ugh, I was just fixing the "edit this page" link helper, and there's special behavior for digitalocean built into _that,_ too! WHY.

PR #1103, opened 2020-01-29
update tfe module docs to reference terraform 0.12

ryanking:
## Description

The module has been upgraded to 0.12-only syntax, so it will not longer work on 0.11.


ryanking:
@nfagerlund - I don't work at Hashicorp, or on those modules, so I have no idea. I was just trying to use it and noticed the staleness in the documentation

nfagerlund:
👍🏼 cool, I'll look into it further and get everything up to date.

PR #1108, opened 2020-02-03
cloud: update Sentinel tfconfig/v2 fields

vancluever:
## Description

Based on post-implementation feedback we are changing some of the field
names in tfconfig. This update reflects those changes:

* "expressions" is now "config"
* "count_expression" is now "count"
* "for_each_expression" is now "for_each"
* "expression" is now "value" (in outputs).

These changes will be reflected in TFC as soon as the pull requests are
approved, we'll merge this after they are live.

### DIFF
--- a/content/source/docs/cloud/sentinel/import/tfconfig-v2.html.md
+++ b/content/source/docs/cloud/sentinel/import/tfconfig-v2.html.md
@@ -146,7 +146,7 @@ main = rule {
 		rc.mode is "managed" and
 			rc.type is "aws_instance"
 	} as _, rc {
-		tfconfig.resources[tfconfig.approximate_address(rc.address)].expressions.ami.constant_value is "ami-abcdefgh012345"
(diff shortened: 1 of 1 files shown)


content/source/docs/cloud/sentinel/import/tfconfig-v2.html.md
```diff
@@ -177,7 +177,7 @@ resource value, one could do:
 import "tfconfig/v2" as tfconfig
 
 main = rule {
-	tfconfig.outputs["instance_id"].expressions.references is ["aws_instance.foo"]
+	tfconfig.outputs["instance_id"].value.references is ["aws_instance.foo"]
```
anamahalo: Just basing off PR desc -- looks like `expressions` is supposed to be `config` here

anamahalo: Oh nvm, this is outputs, right?  So this looks correct I think?

vancluever: Yeah this is good as per the latest PR to change this to `value` in the API.

PR #1118, opened 2020-02-06
Guts: Fix the "Edit this page" link

### LINKED ISSUE #45: "Edit this page" link is (probably?) broken
It's been a while since I last clicked it, so I can't be entirely sure that the behavior has actually changed, but it looks like the "Edit this page link" is broken.

I assume the intended functionality is to open a PR to edit the page being displayed.  Instead, it brings me to the repo's homepage.

nfagerlund:
Fixes #45 (again 😩)

This feature broke when we switched to CircleCI for site builds, but it took us
enough time to notice the problem that we didn't immediately associate it with
the build changes.

Anyway, this method depended on the exact layout of the container's filesystem
root when invoking Docker in a particular way, and Circle doesn't set up the
volumes like that. So let's make it more resilient. It still depends on the
relative layout of these directories, but that's what the symlinks themselves
rely on, so it seems like a good boundary for assumptions.

Adding this additional path math on every page might cause a build time
performance hit, so we should check the average build times a little while after
this ships to make sure it's not too obnoxious. (If so, we can do the path math
more efficiently; it'll just be less legible and more error-prone.)

### DIFF
--- a/content/middleman_helpers.rb
+++ b/content/middleman_helpers.rb
@@ -1,3 +1,5 @@
+require 'pathname'
+
 module Helpers
   # Returns a segment tracking ID such that local development is not
   # tracked to production systems.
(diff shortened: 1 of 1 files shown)


alvin-huang:
This looks great @nfagerlund. I tested the 3 cases the function runs against and they seemed to work properly. Fixing it to be based on a relative path rather than an absolute path will also help us remove a dependency on the working directory of the docker container. In the future if we ever rebuild or use another docker container, the directory structure may be different and there may not be a `/website` so this will be more futureproof.

nfagerlund:
Thanks y'all! Gonna ship it and revisit if we see any more problems.

PR #1119, opened 2020-02-06
cloud: add tree overview to tfconfig/v2, add missing output name doc

vancluever:
## Description

This adds a tree view to the tfconfig/v2 import overview, similar to the
ones the v2 imports have. The other two imports will follow soon.

Also, it looks like the "name" field was missing from the output
documentation. This has been added.


vancluever:
Don't think the errors here are related to this PR.

PR #1126, opened 2020-02-08
cloud: add v1 paths to mock examples

vancluever:
## Description

It's important to note that the v1 paths exist as we want to start to
get people in the habit of possibly future-proofing some of their legacy
policies if necessary. The v1 path allows them to do that and prepare
for a time when they might be using a workspace with v2 as the default.


dylanegan:
@vancluever I saw this pop up in my notifications. Is this good to go?

vancluever:
Ah good catch, looks like I forgot to merge this one!

PR #1128, opened 2020-02-10
Update hashicorp-provider-design-principles.html.md

paultyng:
## Description


nfagerlund:
I'm not sure why I lost track of this one, but I don't see anything controversial or confusing in here, so I'm gonna ship it. Please pardon the wild delay.

PR #1131, opened 2020-02-11
Added formatting for Azure DevOps

justinretzolk:
## Description

In doing some testing, I found that the standard `org/repo` format does not work for Azure DevOps. When I tried `org/project/repo` it also didn't work, so I created a workspace connected to an ADO repository and hit the "Show a workspace" API endpoint. From there, I was able to find that the identifier for Azure DevOps is formatted as `org/project/_git/repo`.


nfagerlund:
Thanks, I hate it!!! (By which I mean, many thanks for the testing and for the patch! 👍🏼 But **wow** that's a nasty identifier format.)

PR #1146, opened 2020-02-20
fix properties names in change password section of API docs

slavrd:
Fixes some mistakes in `docs/cloud/api/account.html` page, `change-your-password` section.


nfagerlund:
Thank you! 

Well, this looks accurate, so I'm gonna go ahead and merge it. But it ALSO looks like maybe it's unintended/undesirable, so I'm asking around about that.

PR #1149, opened 2020-02-24
Update network-requirements.html.md

iojohnso:
misspelled 'application'

## Description

misspelled 'application'


nfagerlund:
Ah, thank you for catching that!

PR #1151, opened 2020-02-25
Add quorum documentation

cgriggs01:
Adding required submodule and symlinks for the new Terraform provider for Quorum


cgriggs01:
The failing travis check is dues to broken azurerm website links

PR #1159, opened 2020-03-04
[Website] Update middleman-hashicorp to 0.3.44

EnMod:
## Description

This PR updates the website's `middleman-hashicorp` dependency to `0.3.44`.   This version intends to mitigate potential security issues by upgrading the package's jQuery to the latest 2.x stable.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 
 references:
   images:
-    middleman: &MIDDLEMAN_IMAGE hashicorp/middleman-hashicorp:0.3.43
+    middleman: &MIDDLEMAN_IMAGE hashicorp/middleman-hashicorp:0.3.44
 
   cache:
     rubygem: &RUBYGEM_CACHE_KEY static-site-gems-v1-{{ checksum "content/Gemfile.lock" }}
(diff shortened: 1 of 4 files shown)


nfagerlund:
OK, I checked out the Gemfile.lock situation -- for some reason, setting the `json` gem to 2.3.0 causes the container to build and install an additional gem every time, which is annoying for local previews. I can't remember what the problem was with that, but I definitely wrestled with it before and decided the best approach was to keep the lockfile pointed at the older gem that's already installed in the container. So I'd be grateful if you could either: 

- Revert that json change in Gemfile.lock
- Figure out how to get the container to build with json 2.3.0 without breaking something else weird

Thanks again for running this update!

EnMod:
Thanks for the notes, @nfagerlund ! 

I ran `git grep 0.3.43` a few times (didn't have `ack`/`ag`/`rg` available to me), and the only value that came up was the CircleCI config.  Not sure where else to change the travis config; it seems the `Travis.yml` doesn't reference `middleman-hashicorp`.  Is there somewhere else it should be changed?

As for the Gemfile.lock situation, great points! I'm going to revert the version bump in Gemfile.lock in the next commit.

PR #1167, opened 2020-03-10
Updated new users URL on users.html.md

KryptionX:
## Description

With TFE v202001-01 and the underlying changes made on https://github.com/hashicorp/atlas/pull/8544 the documentation is now pointing to a no longer valid account creation URL.


nfagerlund:
Ah yikes, thank you for catching this! And extra kudos for leaving a trail back to the relevant change.

PR #1169, opened 2020-03-11
Add clarafication comment to what is backed up

ml4:
This clarifies what is exactly backed up by the new backup API.

### DIFF
--- a/content/source/docs/enterprise/admin/backup-restore.html.md
+++ b/content/source/docs/enterprise/admin/backup-restore.html.md
@@ -13,7 +13,7 @@ Backing up and restoring into a new installation of Terraform Enterprise is the 
 
 ## About Backups
 
-Terraform Enterprise's backup utility backs up all of the data stored in a Terraform Enterprise installation, including both the blob storage and the PostgreSQL database. It does not back up the the installation configuration.
+Terraform Enterprise's backup utility backs up all of the data stored in a Terraform Enterprise installation, including both the blob storage (Terraform workspace state) and the PostgreSQL database. It does not back up the the installation configuration.
 
 ### API Authentication Token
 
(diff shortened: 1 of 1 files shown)


nfagerlund:
Hmm, that's incomplete, and I think it's maybe incomplete enough to be misleading. [Compare here](https://www.terraform.io/docs/enterprise/system-overview/data-security.html) — the object storage also has all the ingressed configuration versions and private modules, all of the binary plan files, and the run logs.

ml4:
Agreed - thanks, I'd not seen the Data Security page.  I do however think it would be beneficial to link to the Data Security page to increase clarity around this.  Both a customer and, so it appears, me missed this.

content/source/docs/enterprise/admin/backup-restore.html.md
```diff
@@ -13,7 +13,7 @@ Backing up and restoring into a new installation of Terraform Enterprise is the
 
 ## About Backups
 
-Terraform Enterprise's backup utility backs up all of the data stored in a Terraform Enterprise installation, including both the blob storage and the PostgreSQL database. It does not back up the the installation configuration.
+Terraform Enterprise's backup utility backs up all of the data stored in a Terraform Enterprise installation, including both the blob storage (Terraform workspace state) and the PostgreSQL database. It does not back up the the installation configuration.
```
nfagerlund: Okay, how about this: 

```suggestion
Terraform Enterprise's backup utility backs up all of the data stored in a Terraform Enterprise installation, including both the blob storage and the PostgreSQL database. It does not back up the the installation configuration.

See [Data Security](../system-overview/data-security.html) for details about the contents of Terraform Enterprise's blob storage and database.
```

ml4:
👍

PR #1173, opened 2020-03-18
Add a cross link to the new state migration guide

judithpatudith:
## Description

Add a cross link to the newly rewritten state migration guide from the state migration docs. We had talked about automating the addition of UTM tags, but I'd like to see them in action for a while before we go all in on them.


nfagerlund:
Thanks for this!

PR #1176, opened 2020-03-19
Delete and redirect an obsolete Sentinel page

nfagerlund:
This page was no longer reachable from the sidebar nav. It was obsoleted by our
modern VCS support for policy sets.


nfagerlund:
A docs page either dies a hero, or lives long enough to become a villain. 😂

PR #1180, opened 2020-03-25
Reverse type mentioned in flattener example

alexkappa:
## Description

The introduction to the first flattener function (`flattenTaskSpec`) explains the types used in the inverse order. I believe it was meant to be the other way.

### DIFF
--- a/content/source/docs/extend/writing-custom-providers.html.md
+++ b/content/source/docs/extend/writing-custom-providers.html.md
@@ -823,7 +823,7 @@ func resourceServerRead(d *schema.ResourceData, m interface{}) error {
 }
 ```
 
-The so-called `flatteners` are in a separate file `structures_server.go`. The outermost data structure is a `map[string]interface{}` and each item a `[]interface{}`:
+The so-called `flatteners` are in a separate file `structures_server.go`. The outermost data structure is a `[]interface{}` and each item a `map[string]interface{}`:
 
 ```go
 func flattenTaskSpec(in *server.TaskSpec) []interface{} {
--- a/content/source/docs/extend/writing-custom-providers.html.md
+++ b/content/source/docs/extend/writing-custom-providers.html.md
@@ -823,7 +823,15 @@ func resourceServerRead(d *schema.ResourceData, m interface{}) error {
 }
 ```
 
-The so-called `flatteners` are in a separate file `structures_server.go`. The outermost data structure is a `[]interface{}` and each item a `map[string]interface{}`:
+The "flattener" helper functions are located in a separate file (`structures_server.go`), and each flattener function handles one data structure and its children. (For complex structures, a flattener should handle children by calling another flattener.)
+
+The return type of a flattener should match the portion of the schema it's responsible for; any internal structures created by the flatteners also must match the corresponding schema types. (See [the schema types page](/docs/extend/schemas/schema-types.html) for details on the underlying Go type for each schema type.)
+
+In this example: 
+
+- `task_spec` is defined as `schema.TypeList`, so `flattenTaskSpec` must return a slice (`[]interface{}`).
+- Same for `container_spec` and `flattenContainerSpec`.
+- `mounts` is defined as `schema.TypeSet`, so `flattenServiceMounts` must return a slice. Since the elements of the set are `&schema.Resource` objects (`map[string]interface{}`), the return type of `flattenServiceMounts is declared as `[]map[string]interface{}`.
 
 ```go
 func flattenTaskSpec(in *server.TaskSpec) []interface{} {


nfagerlund:
This looks right to me, but could I get a double-check from someone who's elbow-deep in provider dev stuff? I've tagged a couple people who seem likely.

content/source/docs/extend/writing-custom-providers.html.md
```diff
@@ -823,7 +823,7 @@ func resourceServerRead(d *schema.ResourceData, m interface{}) error {
 }
 ```
 
-The so-called `flatteners` are in a separate file `structures_server.go`. The outermost data structure is a `map[string]interface{}` and each item a `[]interface{}`:
+The so-called `flatteners` are in a separate file `structures_server.go`. The outermost data structure is a `[]interface{}` and each item a `map[string]interface{}`:
```
im2nguyen:
```suggestion
The `flatteners` are in a separate file `structures_server.go`. Each flattener function corresponds with its respective data structure (`TaskTemplate - []interface{} - schema.TypeList` -> `ContainerSpec []interface{} - schema.TypeList` -> `Mounts - []map[string]interface{} - schema.TypeSet{}`).
```

This is a more accurate (but uglier) description of this. It's basically takes in the root element,`TaskTemplate`, and maps its child elements to the appropriate schema.

- TaskTemplate is `schema.TypeList` (albeit with one item), therefore it's []interface{}.
- ContainerSpec is also `schema.TypeList` with one item, therefore it's []interface{}.
- Mounts is a `schema.TypeSet` (explained in the docs, each item is unique). Unlike its parent element, it contains a list of object map[string]interface{}. Therefore, it is []map[string]interface{}.

There's three levels of nesting. 

Outermost (TaskTemplate) is `[]interface{}`, next layer (ContainerSpec) is also `[]interface{}`, next layer (Mounts) is also `[]interface{}` which contains the last layer (individual mount object) `map[string]interface{}`.

```
"TaskTemplate": {
  "ContainerSpec": {
    "Mounts": [
      {
        "Type": "volume",
        "Source": "tftest-volume",
        "Target": "/mount/test",
        "VolumeOptions": {
          "NoCopy": true,
          "DriverConfig": {}
        }
      }
    ]
  }
}
```

alexkappa: If I understand correctly, all of these are an outer `[]interface{}` wrapping one or more `map[string]interface{}`. Whether a set or a list, flattening always relies on maps being wrapped in a `[]interface{}`. Just when there are more than one, the slice can have more items - which is the case with mounts.

I can see though `flattenContainerSpec` is unnecessarily complex as it has only one element. Therefore it could be more similar to `flattenTaskSpec`.

im2nguyen: I think you're correct. It's just that the following sounds like there's one `[]interface{}` and each child element is `map[string]interface{}`, which isn't the case.

> The so-called `flatteners` are in a separate file `structures_server.go`. The outermost data structure is a `[]interface{}` and each item a `map[string]interface{}`

A more accurate description would be:

> The outermost data structure is a `[]interface{}` and each item a `[]map[string]interface{}`.

But I think that's more confusing because the first two flattener functions return an `[]interface{}`

nfagerlund: I've got to say, I'm glad you submitted this; as someone who's not familiar with provider development, I found it extremely confusing. 

After staring at this for a while, I think part of the problem is that the example has a bunch of random implementation differences for each flattener that aren't relevant to the explanation at the top. What do you think about something like this?

```suggestion
The "flattener" helper functions are located in a separate file (`structures_server.go`), and each flattener function handles one data structure and its children. (For complex structures, a flattener should handle children by calling another flattener.)

The return type of a flattener should match the portion of the schema it's responsible for. (See [the schema types page](/docs/extend/schemas/schema-types.html) for details on the underlying Go type for each schema type.) In this example: 

- Both `task_spec` and `container_spec` are defined as `schema.TypeList`, so `flattenTaskSpec` and `flattenContainerSpec` must return a slice (`[]interface{}`).
- `mounts` is defined as a `schema.TypeSet`, so `flattenServiceMounts` must also return a slice; in this case, the return type is declared as `[]map[string]interface{}`, but `[]interface{}` would also be accurate.
- Any internal structures created by the flatteners also must match the corresponding schema types; elements that were defined as `&schema.Resource` are represented in Go as `map[string]interface{}`, and the innermost `schema.TypeMap` values (`labels` and `driver_options` inside the individual mounts) must be specifically cast to `map[string]interface{}` rather than `map[string]string`.
```

alexkappa: Hi Nick, good suggestion! In general there is more body to describe flatteners in more detail.

```
`mounts` is defined as a `schema.TypeSet`, so `flattenServiceMounts` must also return a slice; in this case, the return type is declared as `[]map[string]interface{}`, but `[]interface{}` would also be accurate.
```

The last sentence though might leave some room for confusion. What if if phrase is as folllows:

```
the return type is declared as `[]interface{}` which holds elements of type `map[string]interface{}`.
```

What do you think?

nfagerlund: Hmm, I find _that_ confusing, since the return type doesn't get _declared_ as `[]interface{}` for the flattener function. 😅 

But how about this: 

```suggestion
The "flattener" helper functions are located in a separate file (`structures_server.go`), and each flattener function handles one data structure and its children. (For complex structures, a flattener should handle children by calling another flattener.)

The return type of a flattener should match the portion of the schema it's responsible for; any internal structures created by the flatteners also must match the corresponding schema types. (See [the schema types page](/docs/extend/schemas/schema-types.html) for details on the underlying Go type for each schema type.)

In this example: 

- `task_spec` is defined as `schema.TypeList`, so `flattenTaskSpec` must return a slice (`[]interface{}`).
- Same for `container_spec` and `flattenContainerSpec`.
- `mounts` is defined as `schema.TypeSet`, so `flattenServiceMounts` must return a slice. Since the elements of the set are `&schema.Resource` objects (`map[string]interface{}`), the return type of `flattenServiceMounts is declared as `[]map[string]interface{}`.
```

alexkappa: Perfect, this is quite clear 👍 Thanks!

nfagerlund:
Okay, rad!! I'm going to go ahead and merge this... and if we get other suggestions later, that can happen on another pull request. 👍🏼

PR #1182, opened 2020-03-26
Move tests from travis to circle

mdeggies:
## Description

This is a small PR to move the current tests out of travis and into circle. The tests will continue to run on any branch push, PR, or merge to master. I also updated the circle config to v2.1. Example from this PR:

Screen Shot 2020-03-26 at 8 12 14 AM

When you click on details, you'll be linked to the success logs: https://circleci.com/gh/hashicorp/terraform-website/647?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Edit: I'm using the machine executor here because the website-test make target mounts volumes in the docker run cmd, and I couldn't find a nice way to do that from within a docker image in Circle: https://circleci.com/docs/2.0/building-docker-images/#mounting-folders. At the bottom of the docs here, it suggests using the machine executor to avoid the mess.


.circleci/config.yml
```diff
@@ -9,6 +9,26 @@ references:
...
+      - checkout
+
+      - run:
+          name: gem and bundle install 
+          command: |
+            gem install bundler -v '1.17.3' --no-document
+            bundle _1.17.3_ install --jobs=3 --retry=3
+
+      - run:
+          name: make sync
```
alvin-huang: The name will be the command if you don't set this. So doing `run: make sync` will give you the same result.

.circleci/config.yml
```diff
@@ -9,6 +9,26 @@ references:
...
+          command: make sync
+
+      - run:
+          name: make website-test
```
alvin-huang: Same here.

.circleci/config.yml
```diff
@@ -9,6 +9,26 @@ references:
     rubygem: &RUBYGEM_CACHE_KEY static-site-gems-v1-{{ checksum "content/Gemfile.lock" }}
 
 jobs:
+  website-test:
+    machine:
```
alvin-huang: This probably doesn't need a machine image and could use the docker container but it looks like in Travis it used 2.3.1 and that circleci/ruby version doesn't exist. Since the build for the deploy uses the `*MIDDLEMAN_IMAGE` I would just reuse that here.

mdeggies: I'm actually using the machine executor here because the `website-test` target mounts volumes in the docker run cmd, and I couldn't find a nice way to do that from a base docker image: https://circleci.com/docs/2.0/building-docker-images/#mounting-folders. At the bottom of the docs here, it suggests using the machine executor to avoid the mess :D

.circleci/config.yml
mdeggies: thanks alvin!

.circleci/config.yml
```diff
@@ -1,14 +1,30 @@
...
+      image: *UBUNTU_IMAGE
+    steps:
+      - checkout
+
+      - run:
+          name: gem and bundle install 
+          command: |
+            gem install bundler -v '1.17.3' --no-document
```
nfagerlund: Just as fyi/context, the need to specify this bundler version was originally an artifact of Travis's build workers being configured in some bizarre way. I don't know whether Circle's ubuntu image would have the same problem, but it should be harmless to keep this, and the experiments we'd need to do before removing it are probably not a great use of our time, given that our middleman platform is on borrowed time anyway.

mdeggies:
Thanks for reviewing, @nfagerlund! I modified the README- let me know if it's safe to merge or if it needs more edits. Alvin is working on migrating https://github.com/hashicorp/terraform/blob/master/.travis.yml to Circle as well, so I removed the Travis section altogether.

nfagerlund:
@mdeggies Hello! Is your team still interested in moving these tests to Circle? It looks like this is still mergeable, but I'm not sure whether it still matches what you all want done.

mdeggies:
Hey @nfagerlund, thanks for the nudge on this one :) 
I'm working on fixing the tests, but I'm not sure what to do when the stable-website branch doesn't exist for certain providers (e.g. archive). May ping you on slack if I'm not able to figure it out soon, but I'll plan to wrap this up shortly.

PR #1183, opened 2020-03-26
TFE: In nav, move clustered install below individual install

nfagerlund:
By request of TFE team.


nfagerlund:
That diff is being displayed in the least useful possible way. 😂 Ignore all the interleaved changes, it's really just moving the "clustered" chunk to below the other two chunks.

glasner:
@nfagerlund thank you 🙌 lgtm

PR #1184, opened 2020-03-27
TFE: Don't imply that external Vault config is better

nfagerlund:
(Since it's mostly not.) By request of TFE team.

@amy-hashi, how's this?


amy-hashi:
Looks perfect!

PR #1188, opened 2020-03-31
cloud: add section on configuring Sentinel modules

vancluever:
## Description

This adds a small section for configuring modules to the policy
management documentation.

Nothing too fancy or meaty about it - just how to add it to the
sentinel.hcl file and a callout to the fact that you can't include
modules outside of the policy set working directory hierarchy right now.

### DIFF
--- a/content/source/docs/cloud/sentinel/manage-policies.html.md
+++ b/content/source/docs/cloud/sentinel/manage-policies.html.md
@@ -47,13 +47,31 @@ Every policy set requires a configuration file named `sentinel.hcl`. This config
 
 The `sentinel.hcl` configuration file may contain any number of entries which look like this:
 
-```python
+```hcl
 policy "sunny-day" {
     enforcement_level = "hard-mandatory"
 }
 ```
 
 In the above, a policy named `sunny-day` is defined with a `hard-mandatory` [enforcement level](#enforcement-levels).
+
+#### Modules
+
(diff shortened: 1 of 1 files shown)


content/source/docs/cloud/sentinel/manage-policies.html.md
```diff
@@ -47,14 +47,32 @@ Every policy set requires a configuration file named `sentinel.hcl`. This config
...
 }
 ```
 
 In the above, a policy named `sunny-day` is defined with a `hard-mandatory` [enforcement level](#enforcement-levels).
 
+#### Modules
+
+Terraform Cloud has support for Sentinel's modules feature. This allows you to write re-usable policy code that can be imported from within several policies as once, reducing the amount of boilerplate from within a policy itself.
+
+-> **NOTE:** We recommend you read the [Sentinel runtime's modules documentation](https://docs.hashicorp.com/sentinel/extending/modules/) in order to understand how to fully use modules within Sentinel. Note that the configuration examples in the runtime documentation pertain to the Sentinel CLI and not Terraform Cloud. HCL-based configuration will be coming to the Sentinel CLI in future releases, at which point in time both Sentinel and TFC configurations will be much similar.
```
cam-stitt, line 62:
```suggestion
-> **NOTE:** We recommend you read the [Sentinel runtime's modules documentation](https://docs.hashicorp.com/sentinel/extending/modules/) in order to understand how to fully use modules within Sentinel. Note that the configuration examples in the runtime documentation pertain to the Sentinel CLI and not Terraform Cloud. HCL-based configuration will be coming to the Sentinel CLI in future releases, at which point in time both Sentinel and TFC configurations will be much more similar.
```

PR #1189, opened 2020-04-01
Updated Getting Started section for Policies

hcrhall:
## Description

This PR will update the [Configuring Sentinel Policies](https://www.terraform.io/docs/cloud/getting-started/policies.html) section within the Terraform Cloud Getting Started documentation. The changes include:

- Title change from _Configuring Sentinel Policies_ to _Enforcing Policies_
- Simplify the messaging for Terraform by moving away from referring to policies as "Sentinel policies" and instead use "policies".  
- The inclusion of a policy sample that can be immediately enforced on a workspace.
- Updated images to reflect changes that have been made in the UI since the guidance was originally authored.

Incremental changes will be made over the coming weeks to align the existing [Sentinel](https://www.terraform.io/docs/cloud/sentinel/index.html) documentation, with a focus on policy management in Terraform and less so around Sentinel itself. 

We will handoff to Sentinel only when discussing concepts that are specific to the Sentinel language and framework.


content/source/docs/cloud/getting-started/policies.html.md
```diff
@@ -54,33 +54,37 @@ vi sentinel.hcl
...
+vi allowed-terraform-version.sentinel
 ```
 
-The following policy code can be used to perform the most basic check:
+The following policy code can be used to determine the Terraform version that is in use:
 
-```python
-main = rule { true }
+```sentinel
+import "tfplan"
```
dylanegan, line 80: Thinking about this a bit. If we specified `tfplan/v2` that would automatically enforce a Terraform version check, correct?

content/source/docs/cloud/getting-started/policies.html.md
```diff
@@ -1,24 +1,24 @@
 ---
 layout: "cloud"
-page_title: "Configuring Sentinel Policies - Getting Started - Terraform Cloud"
+page_title: "Configuring Policies - Getting Started - Terraform Cloud"
```
dylanegan, line 3: :chef_kissing_fingers:

content/source/docs/cloud/getting-started/policies.html.md
```diff
@@ -1,24 +1,24 @@
...
 A policy consists of:
 
-- The Sentinel policy code
+- The policy controls defined as code
 - An enforcement level that changes how a policy affects the run lifecycle
 
 ### Enforcement Levels
 
-Enforcement levels in Sentinel are used for defining behavior when policies fail to evaluate successfully. Sentinel provides three enforcement modes:
+Enforcement levels in Terraform are used for defining behavior when policies fail to evaluate successfully. Terraform provides three enforcement modes:
```
dylanegan, line 21: This is a place where we don't want to switch. Enforcement is purely a Sentinel concept. We still introduce Sentinel in this guide so I think it is fine to introduce Sentinel concepts throughout.

content/source/docs/cloud/getting-started/policies.html.md
```diff
@@ -54,33 +54,37 @@ vi sentinel.hcl
...
-The following policy code can be used to perform the most basic check:
+The following policy code can be used to determine the Terraform version that is in use:
 
-```python
-main = rule { true }
+```sentinel
+import "tfplan"
+
+main = rule {
+  tfplan.terraform_version matches "^0\\.12\\.\\d+$"
```
dylanegan, line 83: :chef_kissing_fingers:

content/source/docs/cloud/getting-started/policies.html.md
```diff
@@ -114,19 +118,19 @@ The navigation bar on the left hand side lists various settings pages. Click the
 
 ![navigate](./images/policy-sets-navigate-2.png)
 
-The Policy Sets page lists all existing policy sets. The list will be empty if this is your first policy set. Click on the "Create a new policy set" button.
+The Policy Sets page lists all existing policy sets. The list will be empty if this is your first policy set. Click on the "Connect a new policy set" button.
```
dylanegan, line 121: Good catch!

content/source/docs/cloud/getting-started/policies.html.md
hcrhall: I've rolled this back

content/source/docs/cloud/getting-started/policies.html.md
hcrhall: @dylanegan the v2 imports do not present themselves if the version of Terraform is less than 0.12.x. so there is no version check.

dylanegan: Clarified over Zoom that my thoughts were about importing `tfplan/v2` would itself be a version check if the policy outright fails because the import wouldn't exist.

content/source/docs/cloud/getting-started/policies.html.md
```diff
@@ -1,19 +1,19 @@
 ---
 layout: "cloud"
-page_title: "Configuring Sentinel Policies - Getting Started - Terraform Cloud"
+page_title: "Enforcing Policies - Getting Started - Terraform Cloud"
 ---
 
-# Configuring Sentinel Policies
+# Enforcing Policies
```
dylanegan, line 6: I found https://github.com/hashicorp/terraform-website/blob/master/content/source/docs/cloud/getting-started/runs.html.md#next-steps mentions `start applying policy checks`. Not sure if you wanted to enforce `enforce` there too.

nfagerlund:
I'd like to discuss this briefly with the Terraform education team before merging -- I'll try and make some progress on that tomorrow.

dylanegan:
@nfagerlund sounds good. Keen to hear from y'all.

content/source/docs/cloud/getting-started/policies.html.md
hcrhall: Fixed

PR #1208, opened 2020-04-20
Test code fails to run if the error parameter is not assigned

rajgandhi0011:
The test code fails to run if the error parameter is not assigned. Please see the error below:
~ terraform apply
An execution plan has been generated and is shown below.Resource actions are indicated with the following symbols:  + create
Terraform will perform the following actions:
  # example_server.my-server will be created  + resource "example_server" "my-server" {      + address = "1.2.3.4"      + id      = (known after apply)    }
Plan: 1 to add, 0 to change, 0 to destroy.
Do you want to perform these actions?  Terraform will perform the actions described above.  Only 'yes' will be accepted to approve.
  Enter a value: yes
example_server.my-server: Creating...
Error: Provider produced inconsistent result after apply
When applying changes to example_server.my-server, provider"registry.terraform.io/-/example" produced an unexpected new value for waspresent, but now absent.
This is a bug in the provider, which should be reported in the provider's ownissue tracker.

## Description

### DIFF
--- a/content/source/docs/extend/writing-custom-providers.html.md
+++ b/content/source/docs/extend/writing-custom-providers.html.md
@@ -199,19 +199,19 @@ readability. Fill in those stubs now, paying close attention to method
 signatures.
 
 ```golang
-func resourceServerCreate(d *schema.ResourceData, m interface{}) error {
+func resourceServerCreate(d *schema.ResourceData, m interface{}) (err error) {
         return resourceServerRead(d, m)
 }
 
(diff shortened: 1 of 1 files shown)


nfagerlund:
This guide was replaced by a more extensive tutorial on learn.hashicorp.com (https://learn.hashicorp.com/tutorials/terraform/provider-setup), which doesn't have the reported issue. So I'm going to close the pull request. Thank you for your effort, and sorry we ended up replacing the guide before we could use it!

PR #1209, opened 2020-04-21
CSS: Restore size of code spans within headers

nfagerlund:
Partial revert of c353899183.

@kylemac, this was your commit that I'm proposing we back out, so I wanted to pull you in for context and approval! Do you remember what visual problem the original commit was addressing? Did it predate the most recent font stack refresh? 

Right now, code spans within headers look tiny, like they got lost and don't belong in the header:

Screenshot_2020-04-20 Enforcing Policies - Getting Started - Terraform Cloud - Terraform by HashiCorp(1)

Going back to 90% makes them look less distracting:

Screenshot_2020-04-20 Enforcing Policies - Getting Started - Terraform Cloud - Terraform by HashiCorp

### DIFF
--- a/content/source/assets/stylesheets/_inner.scss
+++ b/content/source/assets/stylesheets/_inner.scss
@@ -197,7 +197,7 @@
   h4 > code,
   h5 > code h6 > code {
     color: inherit;
-    font-size: 50%;
+    font-size: 90%;
     margin: 0 5px;
   }
 


kylemac:
@nfagerlund the issue came when the inline code blocks dropped to a second line.  this PR has the context: https://github.com/hashicorp/terraform-website/pull/1035 . I think you'll just want to compare this PR against the version you're proposing here.

that said, the sentinel example you have is def an improvement.

nfagerlund:
@kylemac Ah, thank you! I knew there must have been a good reason for that change, I just couldn't remember what it was. Let's see what that problematic page looks like these days:

**Currently, in production:** 

image

**After this PR:** 

image

So, this change looks like it's probably fine. I know I did some other repairs on that header area a while back, and those must have helped with whatever was causing those wild misalignments before.

nfagerlund:
image

Seems pretty decent, actually! So I'm gonna go ahead and merge this.

PR #1210, opened 2020-04-21
Adding note: org tokens cannot manage workspace's ssh keys

yaroslav-007:
Adding note that org tokens cannot manage workspace ssh keys

## Description

Adding note on **Runs API** docs page stating that Managing SSH keys cannot be done via ORG token.

This can be checked from https://www.terraform.io/docs/cloud/users-teams-organizations/api-tokens.html#access-levels

### DIFF
--- a/content/source/docs/cloud/api/workspaces.html.md
+++ b/content/source/docs/cloud/api/workspaces.html.md
@@ -849,6 +849,8 @@ $ curl \
 
 ## Assign an SSH key to a workspace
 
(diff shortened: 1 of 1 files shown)


nfagerlund:
Hello! Thank you for the pull request! But I'm going to close it for now, because it's not accurate -- organization tokens can't interact directly with SSH key objects, but they _can_ edit every property of a workspace object, including the `relationships` object that can include an SSH key ID. (I've checked this personally, and also confirmed with the development team.) 

It IS totally impractical to use organization tokens for this -- since organization tokens can't list SSH keys, a system that only has an org token probably can't do anything useful with the SSH key relationship. (Well, it could null them out.) But it's not prohibited!

PR #1214, opened 2020-04-22
Update glossary.html.md

unacceptable:
Shameless plagiarism from @apparentlymart

## Description

_Adding some documentation about stale plans so I can point my team at the docs._


unacceptable:
I just realized that there was a broken link.  I will fix this at some point in the near future.

PR #1215, opened 2020-04-22
Removing Gitlab Subgroups unsupported note

vaficionado:
Gitlab subgroups are now supported, so removing the note indicating they aren't

## Description

We've had several customers confirm the work we did to unify the VCS experience has corrected this problem, so removing the statement in the docs that it exists


nfagerlund:
That's fantastic! Shippin' it.

PR #1217, opened 2020-04-24
Update index.html.md

siassaj:
## Description

Re https://github.com/hashicorp/terraform/issues/23407

I could not find any documentation making it clear that import will run never on Terraform Cloud's own servers even if Remote execution is configured.

I didn't know if there was a bug with my Terraform version, provider version or configuration & it took a fair bit of digging to find the relevant issue in github.

### DIFF
--- a/content/source/docs/cloud/run/index.html.md
+++ b/content/source/docs/cloud/run/index.html.md
@@ -81,4 +81,4 @@ For full details about the stages of a run, see [Run States and Stages][].
 
 Terraform Cloud does not support remote execution for `terraform import`. For this command the workspace acts only as a remote backend for Terraform state, with all execution occuring on your own workstations or continuous integration workers.
 
-Terraform variables defined in the workspace are available, however environment variables defined in the workspace are not available to `terraform import`, and must be defined within your local execution scope.
+Since `terraform import` runs locally, environment variables defined in the workspace are not available. Any environment variables required by the provider you're importing from must be defined within your local execution scope.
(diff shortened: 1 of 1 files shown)


content/source/docs/cloud/run/index.html.md
```diff
@@ -76,3 +76,9 @@ In the list of workspaces on Terraform Cloud's main page, each workspace shows t
 For full details about the stages of a run, see [Run States and Stages][].
 
 [Run States and Stages]: ./states.html
+
+## Import
+
+Terraform Cloud does not support remote execution for `terraform import`. For this command the workspace acts only as a remote backend for Terraform state, with all execution occuring on your own workstations or continuous integration workers.
+
+Terraform variables defined in the workspace are available, however environment variables defined in the workspace are not available to `terraform import`, and must be defined within your local execution scope.
```
nfagerlund, line 84:
```suggestion
Since `terraform import` runs locally, environment variables defined in the workspace are not available. Any environment variables required by the provider you're importing from must be defined within your local execution scope.
```

Since the `import` command can never really do anything with Terraform variables anyway, this paragraph becomes clearer if we don't mention them.

siassaj:
@nfagerlund build failed but I'm not certain that it's to do with the changes that we made, what do we do now?

nfagerlund:
@siassaj No worries on the build failure -- that's unrelated, and I'm dealing with it elsewhere, so we can ignore it here. Thank you again for the patch!

PR #1222, opened 2020-04-28
adding permissions considerations for vcs/terraform

vaficionado:
per TOB-TFE-036 in the recent TFE security assessment, it was noted that we are not explicit that permissions between VCS and TFC/E are not synchronized. This PR adds information to indicate that customers should be aware of the security implications of using any given VCS user as a conduit between these two systems.

### DIFF
--- a/content/source/docs/cloud/vcs/index.html.md
+++ b/content/source/docs/cloud/vcs/index.html.md
@@ -41,7 +41,7 @@ To use configurations from VCS, Terraform Cloud needs to do several things:
 - Register webhooks with your VCS provider, to get notified of new commits to a chosen branch.
 - Download the contents of a repository at a specific commit in order to run Terraform with that code.
 
-~> **Important:** Terraform Cloud's authorization controls are not inherited from your VCS provider. This means that the user you use when integrating Terraform Cloud with a VCS may have a different level of access to repositories than it does within Terraform Cloud itself. Keep this in mind when selecting a user, as it may affect your security posture in one or both systems.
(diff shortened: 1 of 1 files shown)


eastebry:
Hey this looks good, thanks for making the change!

I might recommend adding a a sentence that makes it very explicit what the implications of this are. Something along the lines of: 
> "Authentication and authorization controls are not applied to commits to connected VCS repositories. Organizations _(not sure this is the right word_) should take precautions to ensure that only trusted contributors can write to connected repositories.

Not entirely sure how to phrase that, but that's the point I think we're trying to drive home.

vaficionado:
@eastebry I considered phrasing it like that, but since the disconnected permissions model is kind of bidirectional (maybe you have a high privilege user in GitHub and a low privilege one in Terraform, or maybe it's the other way around) I went more generic intentionally. We don't want to be prescriptive about how they decide to configure their VCS provider, but do want to make them aware that there are two different AAA models at play.

content/source/docs/cloud/vcs/index.html.md
```diff
@@ -41,6 +41,8 @@ To use configurations from VCS, Terraform Cloud needs to do several things:
 - Register webhooks with your VCS provider, to get notified of new commits to a chosen branch.
 - Download the contents of a repository at a specific commit in order to run Terraform with that code.
 
+~> **Important:** Terraform Cloud's authorization controls are not inherited from your VCS provider. This means that the user you use when integrating Terraform Cloud with a VCS may have a different level of access to repositories than it does within Terraform Cloud itself. Keep this in mind when selecting a user, as it may affect your security posture in one or both systems.
```
nfagerlund, line 44:
```suggestion
~> **Important:** Terraform Cloud usually performs VCS actions using a designated VCS user account, but it has no other knowledge about your VCS's authorization controls and does not associate Terraform Cloud user accounts with VCS user accounts. This means Terraform Cloud's VCS user might have a different level of access to repositories than any given Terraform Cloud user. Keep this in mind when selecting a VCS user, as it may affect your security posture in one or both systems.
```

PR #1227, opened 2020-04-30
Updated the Managing Sentinel Policies documentation to incl. Sentinel Modules

hcrhall:
Updating the Constructing a Policy Set guidance 

## Description

Updating the [Constructing a Policy Set](https://www.terraform.io/docs/cloud/sentinel/manage-policies.html#constructing-a-policy-set) to provide a working example of how Terraform Cloud users can take advantage of modules within Sentinel to abstract away a lot of the boiler-plate and re-use Sentinel code as an `import`.

### DIFF
--- a/content/source/docs/cloud/sentinel/manage-policies.html.md
+++ b/content/source/docs/cloud/sentinel/manage-policies.html.md
@@ -104,7 +104,7 @@ Sentinel policies themselves are defined in individual files (one per policy) in
 
 Using the `enforce-terraform-maintenance-windows.sentinel` policy as an example, we can use the `time` and `tfrun` imports along with our custom `timezone` module to enforce checks that:
 
-1. Load the time when the Terraform Run occurred
+1. Load the time when the Terraform run occurred
 1. Convert the loaded time to PST
 1. Verify that the provisioning operation is only going to occur on an agreed upon day
 
--- a/content/source/docs/cloud/sentinel/manage-policies.html.md
+++ b/content/source/docs/cloud/sentinel/manage-policies.html.md
@@ -69,6 +69,8 @@ module "timezone" {
 }
 ```
 
+-> **NOTE:** At this point in time, you cannot load modules from outside of the policy set working directory hierarchy. This means in the above example, a `source` of `../modules/timezone.sentinel` will not work.
+
 Create a `modules` directory within the policy set working directory, and create a `timezone.sentinel` that looks as follows:
 
 ```sentinel
@@ -95,8 +97,6 @@ offset = func(abbr) {
 ```
 
 The above configuration would tell a policy check to load the code at `./modules/timezone.sentinel` relative to the policy set working directory and make it available to be imported with the statement `import "timezone"` at the top of your Sentinel policy code. This module will be available to all of the policies within the policy set.
-
--> **NOTE:** At this point in time, you cannot load modules from outside of policy set working directory hierarchy. This means in the above example, a `source` of `../modules/timezone.sentinel` will not work.
 
 ### Sentinel policy code files
 
--- a/content/source/docs/cloud/sentinel/manage-policies.html.md
+++ b/content/source/docs/cloud/sentinel/manage-policies.html.md
@@ -48,12 +48,12 @@ Every policy set requires a configuration file named `sentinel.hcl`. This config
 The `sentinel.hcl` configuration file may contain any number of entries which look like this:
 
 ```hcl
-policy "enforce-terraform-maintenance-windows" {
+policy "terraform-maintenance-windows" {
     enforcement_level = "hard-mandatory"
 }
 ```
 
-In the above, a policy named `enforce-terraform-maintenance-windows` is defined with a `hard-mandatory` [enforcement level](#enforcement-levels).
+In the above, a policy named `terraform-maintenance-windows` is defined with a `hard-mandatory` [enforcement level](#enforcement-levels).
 
 #### Modules
 
@@ -100,9 +100,9 @@ The above configuration would tell a policy check to load the code at `./modules
 
 ### Sentinel policy code files
 
-Sentinel policies themselves are defined in individual files (one per policy) in the same directory as the `sentinel.hcl` file. These files must match the name of the policy from the configuration file and carry the `.sentinel` suffix. Using the configuration example above, a policy file named `enforce-terraform-maintenance-windows.sentinel` should also exist alongside the `sentinel.hcl` file to complete the policy set. 
+Sentinel policies themselves are defined in individual files (one per policy) in the same directory as the `sentinel.hcl` file. These files must match the name of the policy from the configuration file and carry the `.sentinel` suffix. Using the configuration example above, a policy file named `terraform-maintenance-windows.sentinel` should also exist alongside the `sentinel.hcl` file to complete the policy set. 
 
-Using the `enforce-terraform-maintenance-windows.sentinel` policy as an example, we can use the `time` and `tfrun` imports along with our custom `timezone` module to enforce checks that:
+Using the `terraform-maintenance-windows.sentinel` policy as an example, we can use the `time` and `tfrun` imports along with our custom `timezone` module to enforce checks that:
 
 1. Load the time when the Terraform run occurred
 1. Convert the loaded time to PST
--- a/content/source/docs/cloud/sentinel/manage-policies.html.md
+++ b/content/source/docs/cloud/sentinel/manage-policies.html.md
@@ -81,8 +81,8 @@ request = http.get(uri)
 response = json.unmarshal(request.body)
 
 offset = func(abbr) {
-	timezone = filter response as _, tz {
-		tz.abbr is strings.to_upper(abbr)
+	timezone = filter response as _, r {
+		r.abbr is strings.to_upper(abbr)
 	}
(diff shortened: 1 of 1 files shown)


hcrhall:
Keen to get some feedback on this draft PR. In particular, the amendments that I have made to the _Constructing a Policy Set_ section. 

@dylanegan I will continue updating the documentation and removing reference to `Sentinel Policies` where appropriate so don't worry about that for now 😉

content/source/docs/cloud/sentinel/manage-policies.html.md
```diff
@@ -64,18 +64,66 @@ Terraform Cloud has support for Sentinel's modules feature. This allows you to w
...
+supported_maintenance_day = rule {
+	tfrun_created_at.add(time.hour * timezone.offset(timezone_abbreviation)).weekday_name in maintenance_days
+}
+
+main = rule {
+	supported_maintenance_day
+}
+```
+
+-> **NOTE:** The above examples uses parameters to to facilitate module reuse within Terraform. For more information on parameters, see the [Sentinel parameter documentation](https://docs.hashicorp.com/sentinel/language/parameters/).
```
dylanegan, line 126:
```suggestion
-> **NOTE:** The above examples uses parameters to facilitate module reuse within Terraform Cloud. For more information on parameters, see the [Sentinel parameter documentation](https://docs.hashicorp.com/sentinel/language/parameters/).
```

It seems as though parameter documentation is limited to the API for Terraform Cloud. It would be good to note for the future that a Terraform Cloud specific guide to parameters would be useful.

content/source/docs/cloud/sentinel/manage-policies.html.md
```diff
@@ -64,18 +64,66 @@ Terraform Cloud has support for Sentinel's modules feature. This allows you to w
...
+module "timezone" {
+    source = "./modules/timezone.sentinel"
 }
 ```
 
-The following entry would tell a policy check to load the code at `./modules/foo.sentinel` relative to the policy set working directory and make it available to be imported with a statement such as `import "foo"` at the top of your Sentinel policy code. This module will be available to all of the policies within the policy set.
+Create a `modules` directory within the policy set working directory, and create a `timezone.sentinel` that looks as follows:
+
+```sentinel
+import "http"
```
dylanegan, line 75: Combo with HTTP. 👨‍🍳 😗

content/source/docs/cloud/sentinel/manage-policies.html.md
```diff
@@ -48,12 +48,12 @@ Every policy set requires a configuration file named `sentinel.hcl`. This config
 The `sentinel.hcl` configuration file may contain any number of entries which look like this:
 
 ```hcl
-policy "sunny-day" {
+policy "enforce-terraform-maintenance-windows" {
```
dylanegan, line 51: Would it make sense to describe the policy beforehand?

hcrhall: @dylanegan great suggestion. Take a look at the latest commit and let me know if anything doesn't make sense.

dylanegan: Looking great!

content/source/docs/cloud/sentinel/manage-policies.html.md
```diff
@@ -64,18 +64,72 @@ Terraform Cloud has support for Sentinel's modules feature. This allows you to w
...
+		print("Details:", tz.text)
+		print("Abbreviation:", tz.abbr)
+		print("Offset:", tz.offset)
+		return tz.offset
+	}
+}
+```
 
--> **NOTE:** At this point in time, you cannot load modules from outside of policy set working directory hierarchy. This means in the above example, a `source` of `../modules/foo.sentinel` will not work.
+The above configuration would tell a policy check to load the code at `./modules/timezone.sentinel` relative to the policy set working directory and make it available to be imported with the statement `import "timezone"` at the top of your Sentinel policy code. This module will be available to all of the policies within the policy set.
```
nfagerlund, line 97: This explanation seems to be attached to the HCL snippet above that makes the module available, and putting it after the Sentinel code example makes it a little harder to understand; on the other hand, moving it between them makes the section end too abruptly. 

I suggest splitting it; explain what the HCL block accomplishes next to the HCL block, then reiterate the `import` syntax down here.

content/source/docs/cloud/sentinel/manage-policies.html.md
```diff
@@ -64,18 +64,72 @@ Terraform Cloud has support for Sentinel's modules feature. This allows you to w
...
 
 ### Sentinel policy code files
 
-Sentinel policies themselves are defined in individual files (one per policy) in the same directory as the `sentinel.hcl` file. These files must match the name of the policy from the configuration file and carry the `.sentinel` suffix. Using the configuration example above, a policy file named `sunny-day.sentinel` should also exist alongside the `sentinel.hcl` file to complete the policy set.
+Sentinel policies themselves are defined in individual files (one per policy) in the same directory as the `sentinel.hcl` file. These files must match the name of the policy from the configuration file and carry the `.sentinel` suffix. Using the configuration example above, a policy file named `enforce-terraform-maintenance-windows.sentinel` should also exist alongside the `sentinel.hcl` file to complete the policy set. 
+
+Using the `enforce-terraform-maintenance-windows.sentinel` policy as an example, we can use the `time` and `tfrun` imports along with our custom `timezone` module to enforce checks that:
+
+1. Load the time when the Terraform Run occurred
```
nfagerlund, line 107:
```suggestion
1. Load the time when the Terraform run occurred
```

Normal nouns like "run" shouldn't be capitalized unless they function as brand names.

content/source/docs/cloud/sentinel/manage-policies.html.md
```diff
@@ -64,18 +64,72 @@ Terraform Cloud has support for Sentinel's modules feature. This allows you to w
...
+supported_maintenance_day = rule {
+	tfrun_created_at.add(time.hour * timezone.offset(timezone_abbreviation)).weekday_name in maintenance_days
+}
+
+main = rule {
+	supported_maintenance_day
+}
+```
+
+-> **NOTE:** The above examples uses parameters to to facilitate module reuse within Terraform. For more information on parameters, see the [Sentinel parameter documentation](https://docs.hashicorp.com/sentinel/language/parameters/). We could expand the enforcement logic to also restrict provisioning to occur out of hours using the [time.hour](https://docs.hashicorp.com/sentinel/imports/time/#time-hour) function.
```
nfagerlund, line 132:
```suggestion
For a more robust or flexible policy, we could expand the enforcement logic to also restrict provisioning to occur out of hours using the [time.hour](https://docs.hashicorp.com/sentinel/imports/time/#time-hour) function.

The above examples use parameters to to facilitate module reuse within Terraform. For more information on parameters, see the [Sentinel parameter documentation](https://docs.hashicorp.com/sentinel/language/parameters/). 
```

I think this can just be paragraphs instead of a blue box. 

I switched the order here to make a smoother transition out of the example. (Ponder fiddly time math -> consider other bits of fiddly time math that might be possible -> dump fiddly time math out of brain's cache -> move on to thinking about structural issues and generalities again.)

content/source/docs/cloud/sentinel/manage-policies.html.md
ryanuber: You don't like sunny days? 😢 Just kidding, I get you're trying to make this more realistic. Minor, but the name is pretty long, can you do something shorter like `maintenance-windows` or `terraform-maintenance-windows`? We know you're enforcing it because it's a sentinel policy 🙃

content/source/docs/cloud/sentinel/manage-policies.html.md
```diff
@@ -64,18 +64,74 @@ Terraform Cloud has support for Sentinel's modules feature. This allows you to w
...
+import "tfrun"
+import "timezone"
+
+param maintenance_days default ["Friday", "Saturday", "Sunday"]
+param timezone_abbreviation default "PST"
+
+tfrun_created_at = time.load(tfrun.created_at)
+
+supported_maintenance_day = rule {
+	tfrun_created_at.add(time.hour * timezone.offset(timezone_abbreviation)).weekday_name in maintenance_days
```
ryanuber, line 124: My only beef with this example is that it suffers the same temporal issues we have known about for a long time. If you plan this run on Saturday, the policy will pass, but nothing stops you from applying it on Wednesday. Personally I'd like to see us back away from this style of example until that's addressed. The description of the policy is actually inaccurate because of this as well.

ryanuber: Note that this is particularly true because this is specifically the Terraform Cloud documentation, where policies are currently only evaluated at plan time and not on an apply.

hcrhall: @ryanuber based on your feedback I've updated the policy to use a rule expression and the `when` predicate so the rule will evaluate on a workspace that has `auto_apply` enabled.

```
supported_maintenance_day = rule when tfrun.workspace.auto_apply is true {
	tfrun_created_at.add(time.hour * timezone.offset(timezone_abbreviation)).weekday_name in maintenance_days
}
```

I do see value in this approach and I _think_ it addresses your concern.

PR #1230, opened 2020-04-30
Safer schema creation

sudomateo:
## Description

It's good practice to only make changes when necessary.


sudomateo:
Failing TravisCI checks are from another provider's docs.

PR #1233, opened 2020-05-04
Fix link to Default anchor

abitrolly:
## Description


nfagerlund:
Good catch, thank you.

PR #1235, opened 2020-05-04
fix 404ing links to hashicorp.com/security

chair6:
An external party reported a broken link to https://hashicorp.com/security.html.  This updates two links to point to the correct location.


nfagerlund:
Ah, thanks for catching that, and my regards to the external party. 🥳

chair6:
🎉 🥳 🎈 🎂 🎉

PR #1239, opened 2020-05-05
Clarify that Docker v19.x is unsupported at this time.

amy-hashi:
## Description

The site currently states we support Docker version 17.06 or greater, but that is untrue and has lead to customers attempting to use v19.x with no success. This update clarifies the supported versions.


straubt1:
Should we also update the non RHEL/CentOS page?

https://github.com/hashicorp/terraform-website/blob/0b5c1826aa1fb44c86bbf3f168050a9a5aacb2c0/content/source/docs/enterprise/before-installing/index.html.md#software-requirements-standalone-deployment

PR #1241, opened 2020-05-06
fix minor typo in retry documentation

ericksoen:
## Description

Change package name from `scheme.TimeoutCreate` to `schema.TimeoutCreate` to package [Terraform package name](https://github.com/hashicorp/terraform/blob/master/helper/schema/resource_timeout.go)


nfagerlund:
Excellent catch, thank you. Could I get you to sign the CLA, and then I'll merge this? I might be able to swing an exception for a one-letter change, but I'm not actually sure what that would entail, so it'll be faster to sign it. (Besides, you might contribute more patches in the future! 😄)

ericksoen:
Done.

nfagerlund:
thank you!

PR #1245, opened 2020-05-07
Nat gateways do incur cost

nmarchini:
Unless I had read the documentation incorrectly NAT gateways do incur a cost on AWS but your docs are suggesting that they don't.
https://aws.amazon.com/vpc/pricing/ 
NAT Gateway Hourly Charge: NAT Gateway is charged on an hourly basis

## Description

Because the NAT gateway creation inside a VPC does incur a cost

### DIFF
--- a/content/source/docs/cloud/cost-estimation/aws.html.md
+++ b/content/source/docs/cloud/cost-estimation/aws.html.md
@@ -25,6 +25,7 @@ Below is a list of resources that cost estimation supports so far. Resources tha
 | aws_elb | incurs cost |
 | aws_instance | incurs cost |
 | aws_lb | incurs cost |
+| aws_nat_gateway | incurs cost |
 | aws_rds_cluster | incurs cost |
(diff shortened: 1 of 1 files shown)


nfagerlund:
Ah, thanks for pointing this out! 

I'm about to close this PR without merging it, BUT: I've added a ticket to the engineering team's board to clarify these docs (including renaming the flag that's currently called "incurs cost", see below), and make the auto-generator distinguish clearly between resources that are actually free and resources that might have a cost but we can't estimate it for some reason. 

The reason I'm not merging this is because these auto-generated docs use "incurs cost" to mean 

- the resource has a cost, AND
- Terraform Cloud understands that cost enough to estimate it. 

Since NAT gateways won't show up in cost estimation at the moment, they shouldn't be marked as "incurs cost." I know that's annoying, so we'll try and fix it; it's just gotta be fixed in a way that won't break when the docs are next auto-generated.

PR #1246, opened 2020-05-08
Backup and restore updates

sudomateo:
Updating the backup and restore documentation with the following
changes.

- Highlight the significance of the `password` provided to the backup
  and restore operations.
- Detail that the Terraform Enterprise version must remain the same
  between a backup and a restore.
- Various readability improvements.

### DIFF
--- a/content/source/docs/enterprise/admin/backup-restore.html.md
+++ b/content/source/docs/enterprise/admin/backup-restore.html.md
@@ -5,35 +5,56 @@ page_title: "Backups and Restores - Infrastructure Administration - Terraform En
 
 # Terraform Enterprise Backups and Restores
 
-Terraform Enterprise is able to backup and restore all of its data. Its backup and restore features work the same way in both clustered and standalone installations.
+Terraform Enterprise provides an API to backup and restore all of its application data. This backup and restore API works the same for both Standalone and Clustered installations.
 
-Backups and restores are initiated by API calls, which can be performed from a remote system. This API is separate from Terraform Enterprise's application-level APIs, and uses a different authorization token.
+The backup and restore API is separate from the Terraform Enterprise application-level APIs. As such, a separate authorization token is required to use the backup and restore API. See [Authentication](#Authentication) for more details.
 
-Backing up and restoring into a new installation of Terraform Enterprise is the only supported way to migrate between deployment types (Standalone vs. Clustered) and operational modes (*External Services* vs. *Mounted Disk*, etc.).
+Using the backup and restore API is the only supported way to:
 
-## About Backups
+- Migrate between deployment types (Standalone, Clustered).
+- Migrate between operational modes (Mounted Disk, External Services).
 
-Terraform Enterprise's backup utility backs up all of the data stored in a Terraform Enterprise installation, including both the blob storage and the PostgreSQL database. It does not back up the the installation configuration.
+## About Backups and Restores
 
-See [Data Security](../system-overview/data-security.html) for details about the contents of Terraform Enterprise's blob storage and database.
+The backup and restore API backs up all of the data stored in a Terraform Enterprise installation, including both the blob storage and the PostgreSQL database. It does not back up the installation configuration. This backup can then be restored to a new installation of Terraform Enterprise.
 
-### API Authentication Token
+Please note the following when using the backup and restore API:
 
-Calls to the backup API must be authenticated with a special bearer token, which can be found on the settings dashboard (`https://<TFE HOSTNAME>:8800/settings`) near the bottom of the page:
+- The version of Terraform Enterprise cannot be changed between a backup and restore. That is, a backup taken from one version of Terraform Enterprise cannot be restored to an installation running a different version of Terraform Enterprise.
+- The Terraform Enterprise installation that will be restored to must be a new, running installation with no existing application data.
+- Once a restore is completed, the Terraform Enterprise application will need to be restarted before it can use the restored data.
+
+See also:
+
+- [Data Security](../system-overview/data-security.html) for details about the contents of Terraform Enterprise's blob storage and PostgreSQL database.
+
+### Authentication 
+
+The backup and restore API uses a separate authorization token which can be found on the settings dashboard (`https://<TFE HOSTNAME>:8800/settings`) near the bottom of the page:
 
 ![Screenshot: the TFE install dashboard, with the API token visible](./images/token.png)
 
-~> **Important:** Since this token can access all of the data in a Terraform Enterprise installation, protect it very carefully.
+-> **Note:** This authorization token is specific to the Terraform Enterprise installation. As such, the authorization token used to create a backup will be different than the authorization token used to perform a restore.
 
-The backup API is not part of Terraform Enterprise's normal admin or application APIs, and cannot be accessed with normal user/team/organization tokens.
+The backup and restore API is separate from the Terraform Enterprise application-level APIs and cannot be accessed with Terraform Enterprise user, team, or organization API tokens.
 
-Use the standard `Authorization: Bearer <TOKEN>` header to authenticate backup and restore calls.
+To use this authorization token with the backup and restore API, pass the `Authorization: Bearer <TOKEN>` header in your API requests.
+
+~> **Important:** Since this authorization token can access all of the data in a Terraform Enterprise installation, protect it very carefully.
 
 ### Security and Encryption
 
-Much of Terraform Enterprise's data is encrypted with Vault before being stored at rest. Backups do not preserve the internal Vault keys; instead, the data is decrypted and then re-encrypted using a single password provided as part of the backup request. When restoring, this password must be provided; the data is then re-encrypted using the new installation's Vault keys.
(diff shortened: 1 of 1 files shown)


rkemery-hashicorp:
Looks good, but I did think of two additional customer discussions that have come up.

1) Backup API is only usable in environments that are mounted disk - not external services. Is this correct? It would seem to be the case since the backup api does the postgres and blob, which in external services are outside components. Came up in a discussion with a customer recently. In this case, we would recommend customers follow the guide to backup their postgres and blob according to their disaster recovery plan and best practices for their cloud provider.
2) On the same ticket with clarifying the payload, the bearer token has a few intricacies. First, the bearer token will generally be unique between instances unless a replicated settings json file was re-used in setup with the same bearer token and in the case where two instances have the same token that is acceptable given that the instance being restored has been restarted after any configuration change that made the bearer token the same as the backup instance.

sudomateo:
@rkemery-hashicorp Thank you for the feedback! Here are my in-line replies.

> Backup API is only usable in environments that are mounted disk - not external services. Is this correct?

I have not heard of this before. I understood that the backup and restore API works for all operational modes (Demo, Mounted Disk, External Services). Maybe @amy-hashi can confirm. That being said, External Services installations do not technically _need_ to use the backup and restore API. They can instead use their own native backup and restore functions. The backup and restore API provides a unified experience for users regardless of what blob storage or PostgreSQL server they are using.

> First, the bearer token will generally be unique between instances unless a replicated settings json file was re-used in setup with the same bearer token

This is true. I'll push a change to account for this. I don't think it's worth mentioning in much detail however since the point is to make sure a user is using the token associated with the instance they are attempting to communicate with.

> given that the instance being restored has been restarted after any configuration change that made the bearer token the same as the backup instance.

My testing showed otherwise. I backed up my Terraform Enterprise instance using token `1234`. I spun up another Terraform Enterprise instance that had token `5678` and used that `5678` token to restore. Once the restore was complete and the application restarted, the token retained the value `5678`.

amy-hashi:
@sudomateo you've got the right of it.

rkemery-hashicorp:
Thank you both! I guess some of the confusion on the backup API is this:

"Terraform Enterprise's backup utility backs up all of the data stored in a Terraform Enterprise installation, including both the blob storage and the PostgreSQL database. It does not back up the the installation configuration."

"including both the blob storage and postgresql" > in external services mode how would that work? The reason I ask is we recently had some customer confusion (and my own) on what this did. In mounted disk, I can see that being accessible by the backup API but in external services does it make a call out to an external postgres instance and blob to back those contents up? I guess I am mixing those two up and my customer did as well. If my postgres is somewhere else all the data in https://www.terraform.io/docs/enterprise/system-overview/data-security.html is now on that instance not on the same box that TFE sits on, yes? Is TFE still going to use its own internal postgres/blob for the stuff outlined in the data security page (in addition to my postgres I have externally?) or when we flip to external do those data pieces also go to that instance? Maybe we need some more clarity here.

> My testing showed otherwise. I backed up my Terraform Enterprise instance using token 1234. I spun up another Terraform Enterprise instance that had token 5678 and used that 5678 token to restore. Once the restore was complete and the application restarted, the token retained the value 5678.

Yes, that should be the case. I was trying to outline a case where the same replicated configuration was used on both the backup and restore instance resulting in the same token being presented. You can change the token to match instance A on instance B but the app needs restarted for that to take effect.

sudomateo:
> but in external services does it make a call out to an external postgres instance and blob to back those contents up?

When using Mounted Disk mode, a container running PostgreSQL is spun up. This container mounts a directory on the host as a volume and uses that as the storage for the PostgreSQL server process it is running. The same is true for Demo mode, albeit a Docker volume is used instead of mounting a directory from the host. In external services, the PostgreSQL server no longer runs on the same host as Terraform Enterprise. In all of these modes, the Terraform Enterprise application is configured with credentials to access the PostgreSQL database. With this configuration the backup and restore API can query the database for the necessary data and place it into the backup blob. I couldn't get into the technical details about it since that's beyond my knowledge at this point in time. Same concept for the blob storage.

> now on that instance not on the same box that TFE sits on, yes? 

No. On a backup the data is pulled from blob storage and PostgreSQL and sent to the caller of the backup and restore API. Where it temporarily lives on the Terraform Enteprise host during this transit I am unaware of.

> Is TFE still going to use its own internal postgres/blob for the stuff outlined in the data security page (in addition to my postgres I have externally?

Terraform Enterprise will use the blob storage and PostgreSQL storage it was configured with at installation time. That is why the backup and restore allows you to mode between operational modes.  If you are using External Services mode, the data will be written to your blob storage and PostgreSQL database on restore. That's why having a fresh installation to restore to is necessary.

> Yes, that should be the case. I was trying to outline a case where the same replicated configuration was used on both the backup and restore instance resulting in the same token being presented. You can change the token to match instance A on instance B but the app needs restarted for that to take effect.

I believe my latest change covers this from both angles. The wording advises the user to ensure they are using the token from the correct host (i.e. use the token from the backup host for the backup and use the token from the host to be restored to for the restore).

content/source/docs/enterprise/admin/backup-restore.html.md
```diff
@@ -5,35 +5,56 @@ page_title: "Backups and Restores - Infrastructure Administration - Terraform En
...
+- [Data Security](../system-overview/data-security.html) for details about the contents of Terraform Enterprise's blob storage and PostgreSQL database.
+
+### Authentication 
+
+The backup and restore API uses a separate authorization token which can be found on the settings dashboard (`https://<TFE HOSTNAME>:8800/settings`) near the bottom of the page:
 
 ![Screenshot: the TFE install dashboard, with the API token visible](./images/token.png)
 
-~> **Important:** Since this token can access all of the data in a Terraform Enterprise installation, protect it very carefully.
+-> **Note:** This authorization token is specific to the Terraform Enterprise installation. Unless otherwise specified during installation of the new Terraform Enteprise instance, the authorization token used to create a backup will be different than the authorization token used to perform a restore.
```
sudomateo, line 37: Open to suggestions on better wording for this section. Specifically this note.

sudomateo: Maybe this?

```suggestion
-> **Note:** This authorization token is specific to the Terraform Enterprise installation. If restoring to a separate Terraform Enterprise installation, the authorization token will be different for restore than it was for backup. Please ensure you are using the correct authorization token for the Terraform Enterprise installation you wish to backup from or restore to.
```

rkemery-hashicorp: > **Note:** This authorization token is specific to the Terraform Enterprise installation. For example, the authorization token will be different when restoring to a separate Terraform Enterprise installation. Please ensure you are using the correct authorization token of the instance you are communicating with when performing a backup or restore operation.

rkemery-hashicorp:
> When using Mounted Disk mode, a container running PostgreSQL is spun up. This container mounts a directory on the host as a volume and uses that as the storage for the PostgreSQL server process it is running. The same is true for Demo mode, albeit a Docker volume is used instead of mounting a directory from the host. In external services, the PostgreSQL server no longer runs on the same host as Terraform Enterprise. In all of these modes, the Terraform Enterprise application is configured with credentials to access the PostgreSQL database. With this configuration the backup and restore API can query the database for the necessary data and place it into the backup blob. I couldn't get into the technical details about it since that's beyond my knowledge at this point in time. Same concept for the blob storage.

That clears it up! Although in mounted disk mode it may be easier to think of the data on all one instance, when using external services mode, the backup API uses the Terraform Enterprise application settings to communicate with the external postgres database and blob storage to perform a backup or restore operation as well.

content/source/docs/enterprise/admin/backup-restore.html.md
```diff
@@ -71,14 +92,20 @@ Status  | Response           | Reason
```
justinretzolk, line 92: The PR as it stands looks good to me. One thing that I think may be of value to add is that in order for the backup to be successful, users *must* receive a `200 OK` back at the end of the call. I've seen cases where the backup timed out, no `200` was received, and the subsequent restore was missing data (as the backup had not completed successfully). 

Perhaps something like this?

```suggestion
~> **Important:** In order for a backup and restore to be successful, the backup API **must** return `200`. If `200` is not returned and the call silently closes, the backup blob may be incomplete, and may result in data loss. 
```

sudomateo: Done!

content/source/docs/enterprise/admin/backup-restore.html.md
sudomateo: Submitted changes!

content/source/docs/enterprise/admin/backup-restore.html.md
```diff
@@ -136,9 +173,7 @@ The JSON file used in the `config` form field should resemble the following:
 curl \
   --header "Authorization: Bearer $TOKEN" \
   --request POST \
-  --form config=@request.json \
+  --form config=@payload.json \
   --form snapshot=@backup.blob \
   https://<TFE HOSTNAME>/_backup/api/v1/restore
 ```
-
-~> **Note:** The Bearer token used to restore the backup is specific to the instance, so if restoring to a separate instance, the token will be different for restore as it was for backup.
```
nfagerlund, line 144: Mmm, I think we should keep a reminder about this inside the restore section... but it more properly belongs right at the START of the restore section. I'd put it right after the paragraph about "before starting, create an installation."

Basically we're just fallible humans out here, and I personally would be 100% guaranteed to forget about the difference in the authentication tokens somewhere between starting the backup and finally bringing up a fresh install.

sudomateo:
Holding off on the merge until @nfagerlund reviews. Not to rush you Nick, I just don't want someone to merge beforehand.

PR #1263, opened 2020-05-20
cloud/api and cloud/sentinel: add information about targeted runs

radditude:
## Description

Updates the API endpoint for creating runs to describe the new `target_addrs` attribute.

Also updates several places in the Sentinel documentation to warn about the possible side effects of using resource targeting, and how to write a policy that prohibits targeted runs altogether.

#### cloud/api/run

Screen Shot 2020-05-19 at 6 57 52 PM

#### cloud/sentinel/import/tfrun

Screen Shot 2020-05-19 at 6 56 29 PM

#### cloud/sentinel/import/tfplan

Screen Shot 2020-05-19 at 6 57 31 PM

#### cloud/sentinel/import/tfplan-v2

Screen Shot 2020-05-19 at 6 56 57 PM

### DIFF
--- a/content/source/docs/cloud/api/run.html.md
+++ b/content/source/docs/cloud/api/run.html.md
@@ -51,6 +51,7 @@ Key path                    | Type   | Default | Description
 ----------------------------|--------|---------|------------
 `data.attributes.is-destroy` | bool | false | Specifies if this plan is a destroy plan, which will destroy all provisioned resources.
 `data.attributes.message` | string | "Queued manually via the Terraform Enterprise API" | Specifies the message to be associated with this run.
+`data.attributes.target-addrs` | array[string] | | Specifies a list of resource addresses to be passed to the `-target` flag.
 `data.relationships.workspace.data.id` | string | | Specifies the workspace ID where the run will be executed.
 `data.relationships.configuration-version.data.id` | string | (nothing) | Specifies the configuration version to use for this run. If the `configuration-version` object is omitted, the run will be created using the workspace's latest configuration version.
 
@@ -68,8 +69,9 @@ Status  | Response                               | Reason
 {
   "data": {
     "attributes": {
-      "is-destroy":false,
-      "message": "Custom message"
+      "is-destroy": false,
+      "message": "Custom message",
+      "target-addrs": ["example.resource_address"]
     },
     "type":"runs",
     "relationships": {
@@ -120,6 +122,7 @@ curl \
       "terraform-version": "0.10.8",
       "created-at": "2017-11-29T19:56:15.205Z",
       "has-changes": false,
+      "target-addrs": ["example.resource_address"],
       "actions": {
         "is-cancelable": true,
         "is-confirmable": false,
--- a/content/source/docs/cloud/api/run.html.md
+++ b/content/source/docs/cloud/api/run.html.md
@@ -51,8 +51,8 @@ Key path                    | Type   | Default | Description
 ----------------------------|--------|---------|------------
 `data.attributes.is-destroy` | bool | false | Specifies if this plan is a destroy plan, which will destroy all provisioned resources.
 `data.attributes.message` | string | "Queued manually via the Terraform Enterprise API" | Specifies the message to be associated with this run.
-`data.attributes.target-addrs` | array[string] | | Specifies a list of resource addresses to be passed to the `-target` flag.
-`data.relationships.workspace.data.id` | string | | Specifies the workspace ID where the run will be executed.
+`data.attributes.target-addrs` | array[string] | (nothing) | Specifies an optional list of resource addresses to be passed to the `-target` flag.
+`data.relationships.workspace.data.id` | string | (nothing) | Specifies the workspace ID where the run will be executed.
 `data.relationships.configuration-version.data.id` | string | (nothing) | Specifies the configuration version to use for this run. If the `configuration-version` object is omitted, the run will be created using the workspace's latest configuration version.
 
 Status  | Response                               | Reason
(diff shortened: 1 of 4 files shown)


content/source/docs/cloud/api/run.html.md
```diff
@@ -51,6 +51,7 @@ Key path                    | Type   | Default | Description
 ----------------------------|--------|---------|------------
 `data.attributes.is-destroy` | bool | false | Specifies if this plan is a destroy plan, which will destroy all provisioned resources.
 `data.attributes.message` | string | "Queued manually via the Terraform Enterprise API" | Specifies the message to be associated with this run.
+`data.attributes.target-addrs` | array[string] | | Specifies a list of resource addresses to be passed to the `-target` flag.
```
mikenomitch, line 54: Do we add `| (nothing)` here to show it can be nil?

content/source/docs/cloud/sentinel/import/tfplan-v2.html.md
```diff
@@ -313,6 +313,8 @@ and data sources within this plan.
 This includes all resources that have been found in the configuration and state,
 regardless of whether or not they are changing.
 
+~> When [resource targeting](/docs/commands/plan.html#resource-targeting) is in effect, the `resource_changes` collection will only include the resources specified as targets for the run. This may lead to unexpected outcomes if a policy expects a resource to always exist. To prohibit targeted runs altogether, use [`tfrun.target_addrs`](./tfrun.html#value-target_addrs).
```
mikenomitch, line 316: Maybe just "use tfrun.target_addrs" is vague?

'Ensure `tfrun.target_addrs` is nil or empty'? might not be better?

content/source/docs/cloud/sentinel/import/tfplan.html.md
```diff
@@ -354,6 +354,8 @@ Some examples of multi-level access are below:
   `tfplan.resources.aws_instance`, with names being the next level down, and so
   on.
 
+~> When [resource targeting](/docs/commands/plan.html#resource-targeting) is in effect, `tfplan.resources` will only include the resources specified as targets for the run. This may lead to unexpected outcomes if a policy expects a resource to always exist. To prohibit targeted runs altogether, use [`tfrun.target_addrs`](/docs/cloud/sentinel/import/tfrun.html#value-target_addrs).
```
mikenomitch, line 357: "a policy expects a resource to always exist" - maybe "a policy expects a resource to be present in the plan"

Maybe that is less clear, but "to always exist" made me think it was talking about the underlying resource actually existing vs the value being handy for sentinel.

content/source/docs/cloud/sentinel/import/tfrun.html.md
```diff
@@ -94,6 +95,20 @@ variables (map of keys)
     └── sensitive (boolean)
 ```
 
+### Value: `target_addrs`
+
+* **Value Type:** An array of strings representing [resource addresses](/docs/internals/resource-addressing.html).
+
+Provides the targets specified using the [`-target`](/docs/commands/plan.html#resource-targeting) flag in the CLI or the `target-addrs` attribute in the API.
+
+To prohibit targeted runs altogether, check that the length of this collection is zero:
```
mikenomitch, line 104: I think this might be null or []. Will this go boom if its null?

(will try it out locally and return with an answer!)

mikenomitch: Yeah. I think this needs to account for null being a valid option. Here's the result for a non-targeted run with this rule:

Screen Shot 2020-05-20 at 10 15 54 AM

radditude: :sad_trombone: thanks for testing that!

content/source/docs/cloud/api/run.html.md
radditude: It doesn't seem like we have a consistent way to indicate optional attributes! I originally left it blank since that's what we're doing for `workspace_id` below, but I don't have a strong opinion one way or the other so happy to update both of those to `(nothing)`

content/source/docs/cloud/sentinel/import/tfplan-v2.html.md
radditude: I like it!

content/source/docs/cloud/sentinel/import/tfplan.html.md
radditude: Good point!

content/source/docs/cloud/sentinel/import/tfrun.html.md
radditude: The updated example should better account for `undefined`.

content/source/docs/cloud/sentinel/import/tfrun.html.md
```diff
@@ -94,6 +95,21 @@ variables (map of keys)
...
+* **Value Type:** An array of strings representing [resource addresses](/docs/internals/resource-addressing.html).
+
+Provides the targets specified using the [`-target`](/docs/commands/plan.html#resource-targeting) flag in the CLI or the `target-addrs` attribute in the API. Will be undefined if no resource targets are specified.
+
+To prohibit targeted runs altogether, make sure the `target_addrs` value is undefined or empty:
+
+```
+import "tfrun"
+
+main = (length(tfrun.target_addrs) else 0) == 0
```
mikenomitch, line 109: the length() function still seems to be failing with undefined

FWIW, I think asserting `tfrun.target_addrs is null` should work, though maybe `tfrun.target_addrs is null or tfrun.target_addrs is []` would work? (I also dont know if thats proper sentinel syntax)

mikenomitch: I misread this! Carry on. ✅ ing

PR #1266, opened 2020-05-22
Note about the Docker network address

sudomateo:
Adding a brief note to check if the Docker bridge network overlaps with any existing networks.


sudomateo:
@justinretzolk I would say it's outside the scope of these docs.

PR #1270, opened 2020-05-29
Remove provisioner plugin from docs

pselle:
## Description

Provisioner plugins are not supported in the current SDK, so remove mention of them from the docs.

### DIFF
--- a/content/source/docs/extend/index.html.md
+++ b/content/source/docs/extend/index.html.md
@@ -37,7 +37,7 @@ components of Terraform (Core vs. Plugins) and the basics of how they interact.
 
 ## [Plugin Types](/docs/extend/plugin-types.html)
 
-Learn about Terraform plugins, currently Providers.
+Learn about Terraform plugins that you can develop using the Terraform Plugin SDK (currently only Providers).
 
 ## [Writing Custom Providers](/docs/extend/writing-custom-providers.html)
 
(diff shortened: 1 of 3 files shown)


pselle:
Requesting review from @hashicorp/terraform-education, not sure what the protocol is at the moment for requesting review here!

content/source/docs/extend/plugin-types.html.md
```diff
@@ -11,13 +11,14 @@ description: |-
...
-only worry about the implementation of their specific Plugin behavior. 
+the AWS provider or the [cloud-init provider](/docs/providers/cloudinit/).
+Terraform Plugins are written in Go and are executable binaries executed as a separate 
+process and communicate with the main Terraform binary over an RPC interface.
+The network communication and RPC is handled automatically by higher-level Terraform
+libraries, so developers need only worry about the implementation of their specific
+Plugin behavior. 
 
-There are two types of plugins supported by Terraform:
+There is currently one type of plugin supported by Terraform:
```
judithpatudith, line 21: The Terraform Plugin SDK currently supports one type of plugin: providers.

content/source/docs/extend/index.html.md
```diff
@@ -39,7 +37,7 @@ components of Terraform (Core vs. Plugins) and the basics of how they interact.
 
 ## [Plugin Types](/docs/extend/plugin-types.html)
 
-Learn about the different types of Terraform plugins — Providers and Provisioners.
+Learn about Terraform plugins, currently Providers.
```
judithpatudith, line 40:
```suggestion
Learn about Terraform plugins that you can develop using the Terraform Plugin SDK (currently only Providers).
```

pselle:
Really good point, Judith! I believe #1 is the most accurate. In the long-run, that won't be true (you won't be able to write custom provisioners at all, SDK or no), but you're correct that you could _theoretically_ still do it (but we don't want to push people in that way, at all).

Thanks for the tip on the Schema page, updating there as well.

PR #1279, opened 2020-06-05
Favor explicitness for the two endpoints

sudomateo:
From a content organization standpoint, it was not clear that the "Show workspace" API could be accessed using two different API endpoints. This change helps separate the two endpoints when reading.


nfagerlund:
👍🏼 !

PR #1292, opened 2020-06-15
Fix name of application-settings.json to match other citations

ml4:
## PR Objective

- [x] Fixing inaccurate docs

## Description
Replicated settings has been changed in automating-the-installer from replicated-settings.json to just settings.json.  All good, except in the snippet for replicated.conf, the file path is to 'application-settings.json' which is misleading.  This makes all citations of the filename the same.


judithpatudith:
Sorry for the slow review! This looks good but I noticed a really cheesy banner on the homepage  of the build for this PR. I didn't see anything in your changes that should cause such a thing, but do you have any ideas what this might be? 

Screen Shot 2020-06-19 at 4 26 54 PM

ml4:
NP on the review @judithpatudith, thanks for reviewing!
Re the ad, do you have ABP on your browser?  That's def nothing to do with me.  I guess it might be worth checking after the merge.

PR #1297, opened 2020-06-17
Add recommendation around exposing ports

amy-hashi:
## PR Objective

Clarify TFE network port information.

- [x] Fixing inaccurate docs

## Description

Port 80 is unnecessary - removed it from the list entirely.

Ports 22 and 8800 should be considered high risk and limited access should be granted to these.

### DIFF
--- a/content/source/docs/enterprise/before-installing/network-requirements.html.md
+++ b/content/source/docs/enterprise/before-installing/network-requirements.html.md
@@ -34,6 +34,7 @@ If Terraform Enterprise is installed in online mode, it accesses the following h
 * `cdn.quay.io`
 * `quay-registry.s3.amazonaws.com`
 * `*.cloudfront.net`
+* `hub.docker.com`
 * `index.docker.io`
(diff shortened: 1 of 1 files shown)


amy-hashi:
> Looks good!
> 
> If we wanted to, we could also break the ingress section down into more descriptive categories:
> 
> * Ingress ‘user’ flows (443, 80), should be open to TFE clients
> * Incoming ‘admin’ flows (22, 8800): Should be restricted to TFE administrators. The use of a secure jumphost or limited source IP range is recommened for access to these ports.
> * ‘Intra-TFE’ flows between services (other ports listed): should be restricted to only the host (for the individual installation), or the subnet (clustered installation).

I like this a lot, actually. Lemme update...

sudomateo:
While we're in here we might want to add DockerHub to the egress connections since as of Replicated 2.45.0, the container images are hosted on DockerHub. Here are the [release notes](https://release-notes.replicated.com/release-notes/2.45.0/) showing that.

PR #1301, opened 2020-06-18
Adding SLA info around HashiCorp's commitment to pre-update IP range endpoint

vaficionado:
We are committing to give our customers 24h advance notice of any network changes via the ip-ranges API, under normal non-emergency circumstances. These two updates reflect that commitment. When we implement the API timestamps, we will want to add additional info around using/comparing those; so for now this is a process-only change

@nfagerlund hopefully I used the Note boxes properly, if not please feel free to reformat accordingly. I think this works but happy for your input on style.

### DIFF
--- a/content/source/docs/cloud/api/ip-ranges.html.md
+++ b/content/source/docs/cloud/api/ip-ranges.html.md
@@ -34,6 +34,8 @@ Name                             | Type   | Description
 
 -> **Note:** The IP ranges for each feature returned by the IP Ranges API may overlap. Additionally, these published ranges do not currently allow for execution of Terraform runs against local resources.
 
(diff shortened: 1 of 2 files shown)


vaficionado:
> Hi! I took quick look and the boxes look ok.
> 
> This might be naive but my one question is where we are going to pre-publish range changes, so that users who care about these changes can know that they're going to get the message. Other than that LGTM.

not at all naive! The IP endpoint itself will just contain the new information, and we will serve a timestamp/last changed date as a header so the user can be made programmatically aware of the update.

vaficionado:
@nfagerlund if you get a sec could you merge this one? I didn't realize it was still open

PR #1307, opened 2020-06-23
Add redirections for null and random

paultyng:
Redirections for the docs migration to the registry


paultyng:
I'm not sure if absolute redirections work, will just need to test once merged.

PR #1309, opened 2020-06-23
Migrate newrelic to registry

paultyng:


paultyng:
Will do this via the new Fastly TF instead.

PR #1310, opened 2020-06-23
Explain redirects.txt a bit better in its comments

nfagerlund:
## PR Objective

- [x] Fixing inaccurate docs

## Description

No one really expects redirects.txt to be limited to intra-site redirects only, but, it is. 🤷🏽‍♀️


paultyng:
Going to handle this in #1312

PR #1311, opened 2020-06-23
add terraform configuration for Fastly redirect dictionaries

alvin-huang:
## PR Objective

Add Terraform configuration for Fastly dictionaries used in redirects to the new equivalent registry.terraform.io domain.

## Description

The configuration of the actual Fastly service is not handled in this Terraform configuration. The VCL is published to Fastly directly. However, the VCL consumes dictionaries as shown in the code snippets in this PR. This Terraform configuration only modifies these dictionaries to make it easier on the team so they do not need to do it in the Fastly interface.

### DIFF
--- a/terraform/fastly.tf
+++ b/terraform/fastly.tf
@@ -1,0 +1,87 @@
+provider "fastly" {
+  version = "~> 0.17.0"
+}
+
+locals {
+  static_sites_service_id              = "7GrxRJP3PVBuqQbyxYQ0MV"
+  tf_registry_docs_type_dictionary_id  = "2meQ4j47Me81w1jSb5m9lh"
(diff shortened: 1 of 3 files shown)


terraform/fastly.tf
```diff
@@ -0,0 +1,87 @@
...
+
+# These should not need changing, these are static mappings of r,d,guides in the old docs site to the equivalents
+# in the Terraform Registry
+resource "fastly_service_dictionary_items_v1" "tf_registry_docs_type" {
+  service_id    = local.static_sites_service_id # we do not manage this service in Terraform at this time
+  dictionary_id = local.tf_registry_docs_type_dictionary_id
+  items = {
+    "r" : "resources"
+    "d" : "data-sources"
+    "guides" : "guides"
```
alvin-huang, line 80: this is not necessary but for completeness I added it

PR #1318, opened 2020-06-25
[Website] Add alert banner at top of page

EnMod:
## PR Objective

- [ ] Adding release banner to the site

## Description

This PR adds an alert banner to the site much like https://nomadproject.io. In fact, much of the banner code is inspired by the one used on that site.

### DIFF
--- a/content/source/assets/javascripts/alert-banner.js
+++ b/content/source/assets/javascripts/alert-banner.js
@@ -27,22 +27,6 @@ function onClose(banner) {
 
   // set the cookie so this banner doesn't show up anymore
   document.cookie = 'banner_terraform_io=1'
-
-  // TODO Check how to implement this
-  // trackEvent('close')
-}
-
-function trackEvent(type) {
-  if (window.analytics) {
-    const { tag, theme, text, linkText } = { ...this.props }
-
-    window.analytics.track(type.charAt(0).toUpperCase() + type.slice(1), {
-      category: 'Alert Banner',
-      label: `${text} - ${linkText} | ${type}`,
-      tag: tag,
-      theme: theme,
-    })
-  }
 }
 
 // Modified from js-cookie
--- a/content/source/assets/javascripts/alert-banner.js
+++ b/content/source/assets/javascripts/alert-banner.js
@@ -1,3 +1,49 @@
+//= require js-cookie
+
 document.addEventListener('turbolinks:load', function () {
-  // TODO Bring over the component logic
+  var qs = document.querySelector.bind(document)
+  var banner = qs('.g-alert-banner')
+  var closeButton = qs('.g-alert-banner .close')
+
+  // if cookie is set, don't show the component
+  if (cookie.get(`banner_terraform_io`)) {
+    closeButton.classList.remove('show')
+    return
+  }
+
+  closeButton.addEventListener(
+    'click',
+    () => {
+      onClose(banner)
+    },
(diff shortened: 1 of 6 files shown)


content/source/assets/javascripts/alert-banner.js
```diff
@@ -0,0 +1,65 @@
+document.addEventListener('turbolinks:load', function () {
+  var qs = document.querySelector.bind(document)
+  var banner = qs('.g-alert-banner')
+  var closeButton = qs('.g-alert-banner .close')
+  // if cookie is set, don't show the component
+  if (getCookie(`banner_terraform_io`) === '1') {
+    banner.classList.remove('show')
+    return
+  }
```
EnMod, line 9: I wonder, would this be better as "if cookie _isn't_ set, _show_ the component"? 

The major difference is which loading method is best:
- showing the alert by default and removing it if cookie is set (as above)
- hiding the alert by default and showing it if cookie _isn't_ set

Thoughts?

content/source/assets/javascripts/alert-banner.js
```diff
@@ -0,0 +1,65 @@
...
+  banner.style.height = `${banner.scrollHeight}px`
+  window.setTimeout(() => {
+    banner.style.height = '0'
+  }, 1)
+
+  // set the cookie so this banner doesn't show up anymore
+  document.cookie = 'banner_terraform_io=1'
+
+  // TODO Check how to implement this
+  // trackEvent('close')
```
mwickett, line 32: You can cut the analytics tracking from this, just to keep things simple.

content/source/assets/stylesheets/_alert_banner.scss
```diff
@@ -0,0 +1,134 @@
...
+    }
+  }
+
+  &.vault {
+    background: linear-gradient(90deg, #6a6d7a 1.56%, #373942 100%);
+
+    & svg path {
+      stroke: #e5e6eb;
+    }
+  }
```
mwickett, line 43: Let's just hard code the styles for Terraform since this is strictly for terraform.io

content/source/assets/stylesheets/application.scss
```diff
@@ -1,42 +1,42 @@
-@import "bootstrap-sprockets";
-@import "bootstrap";
+@import 'bootstrap-sprockets';
+@import 'bootstrap';
```
mwickett, line 2: Can you skip this linting change? Probably we're just out of step with the styles for this repo.

content/source/layouts/_alert_banner.erb
```diff
@@ -0,0 +1,17 @@
+<div class="g-alert-banner <% if theme %><%= theme %><% else %><%= ' default' %><% end %> <% if show %><%= ' show' %><% else %><%= '' %><% end %> <% if tag.length > 3 %><%= ' has-large-tag' %><% else %><%= '' %><% end %>">
```
mwickett, line 1: Same here, we can drop theme since we know we're on terraform.

content/source/assets/javascripts/alert-banner.js
mwickett: It's good to think this through - let's leave it as is. If someone has cookies disabled completely, then they wouldn't see the banner. Not at all a significant concern in this case, but I don't see a reason to change it here. Asking the question is solid though. 👊

PR #1319, opened 2020-06-26
Combine sidebars for TFC/TFE

nfagerlund:
## PR Objective

- [x] Making some docs easier to understand
- [x] Changing behavior or layout of website

## Description

TFC and TFE are the same application, but we currently don't have the tooling to
serve two totally separate versions of the application docs without an
unbearable maintenance burden. Our solution thus far has been to use the TFC
docs as the app docs, and the TFE docs as the deployment/operation docs for
running an instance.

OK in theory, but in practice it makes the docs frustrating for the TFE
application users. This change is intended to:

- Help with discoverability by making the entire TFC docs nav hierarchy visible
within the TFE deployment/ops docs.
- Help with confirmation/belief by changing the TFC sidebar header to
"Terraform Cloud / Terraform Enterprise", so TFE users can trust that
they're in the right place.

PR #1320, opened 2020-06-26
Include TFE in Terraform Cloud page titles

nfagerlund:
## PR Objective

## Description

This commit changes the trailing product name in page titles to "Terraform Cloud
and Terraform Enterprise", throughout the TFC docs.

I'm doing it as a separate PR from #1319 because it's an extremely noisy find-and-replace patch and I didn't want to drown out the actual changes over there.

**For reviewers:** This seemed fairly uncontroversial when I proposed it in the google doc that went around a while back, and it's fairly easy to revert, so I'd just like to get a thumbs-up from @judithpatudith or @glasner that they haven't heard any interesting objections to this.

### DIFF
--- a/content/source/docs/cloud/integrations/service-now/example-customizations.html.md
+++ b/content/source/docs/cloud/integrations/service-now/example-customizations.html.md
@@ -1,6 +1,6 @@
 ---
 layout: "cloud"
-page_title: "Example Customizations - ServiceNow Service Catalog Integration - Terraform Cloud"
+page_title: "Example Customizations - ServiceNow Service Catalog Integration - Terraform Cloud and Terraform Enterprise"
 ---
 
 # Example Customizations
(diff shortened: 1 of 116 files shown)


content/source/docs/cloud/integrations/service-now/example-customizations.html.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: "cloud"
-page_title: "Example Customizations - ServiceNow Service Catalog Integration - Terraform Cloud"
+page_title: "Example Customizations - ServiceNow Service Catalog Integration - Terraform Cloud and Terraform Enterprise"
```
judithpatudith, line 3: I don't think ServiceNow is meant to be used with Cloud, although I don't remember why exactly.

judithpatudith: Same comment for all the Service Now pages

nfagerlund: For that one, I think I'll just leave it like it is in the PR for now, since we'll be adjusting those docs at some point soon anyway. (ServiceNow can work with TFC too, but if you're at the scale where you need SN in the first place, you're probably an enterprise customer by default... is mostly what that's all about.)

Issue #1251, opened 2020-05-12
(Generic Linux docs) Clarify that Docker v19.x is unsupported at this time.


straubt1:
# Description

As outlined in the PR #1239, Docker 19.x is not supported and the appropriate documentation was updated for RHEL and CentOS to:

> Docker CE version 17.06 through 18.x 

However, the generic Linux section for **Standalone** does not match and instead states:

> The instance should run a supported version of Docker engine (1.7.1 or later, minimum 17.06.2-ce, maximum 18.09.2).

As a side note, this is also very confusing "(1.7.1 or later, minimum 17.06.2-ce, maximum 18.09.2)".
Is it 1.7.1+, or 17.06+?
Is this Docker CE, and/or EE?

# Proposed Solution

Update the generic linux documentation to match RHEL and CentOS, and clarify which versions of Docker are supported (including Docker, Docker CE, Docker EE).

PR #1321, opened 2020-06-26
Updating latest supported version for CentOS and RHEL

justinretzolk:
## PR Objective

- [x] Fixing inaccurate docs

## Description

Updating the latest supported Docker versions for RHEL and CentOS.

### DIFF
--- a/content/source/docs/enterprise/before-installing/index.html.md
+++ b/content/source/docs/enterprise/before-installing/index.html.md
@@ -126,7 +126,7 @@ Some operating systems have specific configuration requirements:
 
 For other Linux distributions, check Docker compatibility:
 
-* The instance should run a supported version of Docker engine (1.7.1 or later, minimum 17.06.2-ce, maximum 18.09.2). This also requires a 64-bit distribution with a minimum Linux Kernel version of 3.10.
+* The instance should run a supported version of Docker engine (1.7.1 or later, minimum 17.06.2-ce, maximum 19.03.8). This also requires a 64-bit distribution with a minimum Linux Kernel version of 3.10.
     * Replicated 2.32.0 and above required when running Docker 18+.
     * In Online mode, the installer will install Docker automatically.
     * In Airgapped mode, Docker should be installed before you begin.
--- a/content/source/docs/enterprise/before-installing/centos-requirements.html.md
+++ b/content/source/docs/enterprise/before-installing/centos-requirements.html.md
@@ -12,8 +12,8 @@ When installing Terraform Enterprise on CentOS Linux, ensure your OS and Docker 
 * A [supported version](/docs/enterprise/before-installing/index.html#operating-system-requirements) of CentOS.
 * A suitable version of Docker:
    * Docker 1.13.1 (available in the `extras` repository).
-   * [Docker CE](https://docs.docker.com/install/linux/docker-ce/centos/) version 17.06 through 18.x (Docker v19.x is currently unsupported.)
-   * [Docker EE](https://docs.docker.com/install/linux/docker-ee/centos/) version 17.06 through 18.x (Docker v19.x is currently unsupported.)
+   * [Docker CE](https://docs.docker.com/install/linux/docker-ce/centos/) version 17.06 through 19.03.8.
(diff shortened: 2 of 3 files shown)


straubt1:
Any chance we can resolve this Issue in the process #1251 ?

justinretzolk:
Nice catch @straubt1 - added that one as well.

straubt1:
Correct me if I am wrong, but isn't docker 19.x only supported when using the latest version of replicated?
We should be as specific as possible as this has been a sticking point for customers I have dealt with in the past.

justinretzolk:
Support was added in [2.44.0](https://release-notes.replicated.com/release-notes/2.44.0/). I can see the sentiment, but this is the version that the Replicated installer installs by default now. My view is that these docs are referenced by individuals who are installing Terraform Enterprise, and so probably wouldn't be relevant to anyone who's not installing the latest version of Replicated at the same time, but perhaps that's not quite right?

For what it's worth, I don't think we've ever specified what Replicated version was required for specific Docker versions before, but perhaps I've simply overlooked that in the past?

sudomateo:
Until we get truly versioned documentation, we won't be able to solve the issue of noting these caveats in a clean way. The installation documentation does specify using https://install.terraform.io/ptfe/stable to install Terraform Enterprise which uses the latest version of Replicated and is updated for each Terraform Enterprise release.

This type of caveat would probably be better suited for an upgrade guide, not an initial installation guide like these changes target. I think it's safe to assume a reader of the installation documentation is using the latest versions of Terraform Enterprise and Replicated.

content/source/docs/enterprise/before-installing/index.html.md
```diff
@@ -126,7 +126,7 @@ Some operating systems have specific configuration requirements:
 
 For other Linux distributions, check Docker compatibility:
 
-* The instance should run a supported version of Docker engine (1.7.1 or later, minimum 17.06.2-ce, maximum 18.09.2). This also requires a 64-bit distribution with a minimum Linux Kernel version of 3.10.
+* The instance should run a supported version of Docker engine (1.7.1 or later, minimum 17.06.2-ce, maximum 19.03.8). This also requires a 64-bit distribution with a minimum Linux Kernel version of 3.10.
     * Replicated 2.32.0 and above required when running Docker 18+.
```
straubt1, line 130: We already call out replicated versions here.

straubt1:
I would also call out that in airgap installations, we are artifacting specific versions of replicated and they are only upgrades intentionally (as well as TFE itself).

While I agree an upgrade guide would be great and probably would cover this in more detail, I don't see how being more explicit with the version of Replicated can do any harm?

content/source/docs/enterprise/before-installing/index.html.md
sudomateo: This call out didn't age well and is misleading in its current form. Reading this, you'd think Replicated 2.32.2 is all that's necessary to run Docker 18+, including Docker 19+. That is not the case. This is why I don't recommend notes like this for the installation guide. Something like "Support for Docker 18.y.z was added in Replicated 2.y.z" would have been better.

sudomateo:
@straubt1 I agree that we could describe the airgap installation and upgrade process in greater detail.

For the second point, I would welcome any suggestions that you have. Replicated themselves do not publish a matrix showing which Replicated versions support which Docker versions, making it difficult for us to do so as well. A note saying something like "this installation documentation assumes the latest version of Replicated is being used" would help. A similar note for upgrades would suffice too. Here's a [link](https://help.replicated.com/community/t/installing-docker-in-airgapped-environments/81) to Replicated's OS/Docker compatibility.

justinretzolk:
I agree with the sentiments in that there could be some better mapping for which versions are compatible, but I think we may be straying a bit from the point of this particular PR. We've gotten some questions about the install script trying to install `19.03.8` while the docs said this version is unsupported; this PR is to help clarify that this version is supported and the script installing it _is_ expected.

sudomateo:
I'm going to merge this pull request given that it's necessary for customer facing issues. I also believe it solved the linked issue #1251 in its current form. We can get another issue and/or pull request going for the Replicated references.

sudomateo:
Tracking the outstanding issues in https://github.com/hashicorp/terraform-website/issues/1323.

PR #1322, opened 2020-06-26
Update installer.html.md

larryebaum:
Add online installer step for RHEL that specifies `no-docker` option flag.

## PR Objective

- [ ] Fixing inaccurate docs

### DIFF
--- a/content/source/docs/enterprise/install/installer.html.md
+++ b/content/source/docs/enterprise/install/installer.html.md
@@ -192,6 +192,7 @@ If your instance can access the Internet, use the Online install mode.
 1. From a shell on your instance:
     * To execute the installer directly, run `curl https://install.terraform.io/ptfe/stable | sudo bash`
     * To inspect the script locally before running, run `curl https://install.terraform.io/ptfe/stable > install.sh` and, once you are satisfied with the script's content, execute it with `sudo bash install.sh`.
+    * To install on RedHat Enterprise Linux where pre-installation of Docker is necessary, run `curl https://install.terraform.io/ptfe/stable > install.sh` and, once you are satisfied with the script's content, execute it with `sudo bash install.sh no-docker`. 
 1. The installation will take a few minutes and you'll be presented with a message
     about how and where to access the rest of the setup via the web. This will be
     `https://<TFE HOSTNAME>:8800`.
--- a/content/source/docs/enterprise/install/installer.html.md
+++ b/content/source/docs/enterprise/install/installer.html.md
@@ -192,7 +192,7 @@ If your instance can access the Internet, use the Online install mode.
 1. From a shell on your instance:
     * To execute the installer directly, run `curl https://install.terraform.io/ptfe/stable | sudo bash`.
     * To inspect the script locally before running, run `curl https://install.terraform.io/ptfe/stable > install.sh` and, once you are satisfied with the script's content, execute it with `sudo bash install.sh`.
-    * To install on RedHat Enterprise Linux where pre-installation of Docker is necessary, run `curl https://install.terraform.io/ptfe/stable > install.sh` and, once you are satisfied with the script's content, execute it with `sudo bash install.sh no-docker`. 
+      * RedHat Enterprise Linux requires Docker to be pre-installed. As such, execute the installer script using `sudo bash install.sh no-docker` to prevent the installer script from automatically installing Docker.
 1. The installation will take a few minutes and you'll be presented with a message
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/install/installer.html.md
```diff
@@ -192,6 +192,7 @@ If your instance can access the Internet, use the Online install mode.
 1. From a shell on your instance:
     * To execute the installer directly, run `curl https://install.terraform.io/ptfe/stable | sudo bash`
     * To inspect the script locally before running, run `curl https://install.terraform.io/ptfe/stable > install.sh` and, once you are satisfied with the script's content, execute it with `sudo bash install.sh`.
+    * To install on RedHat Enterprise Linux where pre-installation of Docker is necessary, run `curl https://install.terraform.io/ptfe/stable > install.sh` and, once you are satisfied with the script's content, execute it with `sudo bash install.sh no-docker`.
```
justinretzolk, line 195: I think we can combine the two lines so as to minimize the amount we repeat ourselves.

```suggestion
    * To inspect the script locally before running, run `curl https://install.terraform.io/ptfe/stable > install.sh` and, once you are satisfied with the script's content, execute it with `sudo bash install.sh`. If preferred, the automated installation of Docker can by bypassed by executing the script as `sudo bash install.sh no-docker`.
```

larryebaum: I disagree. RHEL requires specific steps that differ from other platforms. 'If preferred' infers not required, which is incorrect, and doesn't specifically identify the step as required for RHEL. The confusion/issues customers experience stems from us not being explicitly clear in our docs for RHEL, as well as with other areas (I have other forthcoming docs edits).

sudomateo, line 195: @larryebaum Technically speaking, letting the installer install Docker CE on RHEL _will_ work, it just won't be officially supported by Red Hat. That is why we have users install Docker from `rhel-extras` or Docker EE.

I think the below suggestion will suffice. Let me know your thoughts.

```suggestion
    * To inspect the script locally before running, run `curl https://install.terraform.io/ptfe/stable > install.sh` and, once you are satisfied with the script's content, execute it with `sudo bash install.sh`. If Docker is already installed on the instance, which is usually the case when running RHEL, the automated installation of Docker can by bypassed by executing the script as `sudo bash install.sh no-docker`.
```

larryebaum: @sudomateo Wonder if something recent broke the installer behavior you indicate, as my customer/prospect experience last week differed. The installer fully disrespects that fact that Docker is installed, and overwrites it with a later version that isn't supported and contradicts our docs with state 1.13.1 is **required.** Further, this appears to have caused a variety of other issues (in their PoV) and required a wipe/reinstall. I've validated the experience on a test instance. Perhaps Replicated broke the installer? As we call our RHEL requirements elsewhere, distinct from other steps, I think its appropriate to continue to call out this RHEL step so it's clear and not potentially confusing. If the argument is saving 2 lines of info and being potentially confusing, vs 2 extra lines but being clear and a better customer experience, I'd favor the better customer experience. Zoom for discussion?

sudomateo, line 195: This suggestion adds a sub-bullet detailing the RHEL specific instructions.

```suggestion
      * RedHat Enterprise Linux requires Docker to be pre-installed. As such, execute the installer script using `sudo bash install.sh no-docker` to prevent the installer script from automatically installing Docker.
```

content/source/docs/enterprise/install/installer.html.md
```diff
@@ -192,6 +192,7 @@ If your instance can access the Internet, use the Online install mode.
 1. From a shell on your instance:
     * To execute the installer directly, run `curl https://install.terraform.io/ptfe/stable | sudo bash`
```
sudomateo, line 193: Minor syntax adjustment.

```suggestion
    * To execute the installer directly, run `curl https://install.terraform.io/ptfe/stable | sudo bash`.
```

content/source/docs/enterprise/install/installer.html.md
sudomateo, line 195: This suggestion adds a sub-bullet detailing the RHEL specific instructions.

```suggestion
      * RedHat Enterprise Linux requires Docker to be pre-installed. As such, execute the installer script using `sudo bash install.sh no-docker` to prevent the installer script from automatically installing Docker.
```

PR #1337, opened 2020-07-08
Updating docs for enforcement levels

hcrhall:
## PR Objective

- [ ] Fixing inaccurate docs

## Description

This PR includes a minor change to the Manage Policies section in the Terraform Cloud documentation. In the docs, we explain the differences between the different enforcement levels that are available in Sentinel and which personas have the ability to override certain behaviour. This can be confusing for the reader so we are changing the messaging to move away from the concept of a persona (i.e administrator) and explicitly call out the permissions that are required instead.

### DIFF
--- a/content/source/docs/cloud/sentinel/manage-policies.html.md
+++ b/content/source/docs/cloud/sentinel/manage-policies.html.md
@@ -33,7 +33,7 @@ Policy sets are managed at an organization level; viewing and modifying them req
 Enforcement levels in Sentinel are used for defining behavior when policies fail to evaluate successfully. Sentinel provides three enforcement modes:
 
 * `hard-mandatory` requires that the policy passes. If a policy fails, the run is halted and may not be applied until the failure is resolved.
-* `soft-mandatory` is much like `hard-mandatory`, but allows an administrator to override policy failures on a case-by-case basis.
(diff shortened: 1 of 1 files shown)


content/source/docs/cloud/sentinel/manage-policies.html.md
```diff
@@ -33,7 +33,7 @@ Policy sets are managed at an organization level; viewing and modifying them req
 Enforcement levels in Sentinel are used for defining behavior when policies fail to evaluate successfully. Sentinel provides three enforcement modes:
 
 * `hard-mandatory` requires that the policy passes. If a policy fails, the run is halted and may not be applied until the failure is resolved.
-* `soft-mandatory` is much like `hard-mandatory`, but allows an administrator to override policy failures on a case-by-case basis.
+* `soft-mandatory` is much like `hard-mandatory`, but allows any user with the [Manage Policies](../users-teams-organizations/permissions.html#manage-policies) permission to override policy failures on a case-by-case basis.
```
dylanegan, line 36: It is also applicable to override if they're an organisation owner.

hcrhall: @dylanegan An Organization Owner is a persona which is assigned the Manage Policies permission. I wasn't sure if we needed to explicitly call this out or not. Do you think it is necessary?

Juansbulbisful:
-h

PR #1348, opened 2020-07-13
added until block in script examples

andybaran:
## PR Objective

- [ ] Fixing inaccurate docs

## Description

Working with @sudomateo  I've added "until" blocks in the example bash scripts for automated recovery of TFE using Replicated snapshots. Without this, it is a common occurrence for the script to fail because it issues replicatedctl before Replicated is in the ready state.  While this could have been accomplished using 'jq', or other more elegant means, I felt it was important to keep it simple and stick with 'grep' as the script is most likely to be used during a high stress recovery situation.

### DIFF
--- a/content/source/docs/enterprise/admin/automated-recovery.html.md
+++ b/content/source/docs/enterprise/admin/automated-recovery.html.md
@@ -88,11 +88,11 @@ apt-get install -y jq
 curl https://install.terraform.io/ptfe/stable | bash -s fast-timeouts
 
 # Wait for replicated to start before proceeding
-until replicatedctl system status | grep -q -e '"Replicated": "ready",i' -e '"Retraced": "ready"'; do :
-  sleep 1;
-  echo "Replicated is not yet ready";
-done;
-echo 'Replicated is ready.'
+until replicatedctl system status --template '{{and (eq .Replicated "ready") (eq .Retraced "ready")}}' | grep -q true; do
+  sleep 1
+  echo "Replicated is not yet ready."
+done
+echo "Replicated is ready."
 
 # This retrieves a list of all the snapshots currently available.
 replicatedctl snapshot ls $access -o json > /tmp/snapshots.json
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/admin/automated-recovery.html.md
```diff
@@ -87,6 +87,13 @@ apt-get install -y jq
 
 curl https://install.terraform.io/ptfe/stable | bash -s fast-timeouts
 
+# Wait for replicated to start before proceeding
+until replicatedctl system status | grep -q -e '"Replicated": "ready",i' -e '"Retraced": "ready"'; do :
+  sleep 1;
+  echo "Replicated is not yet ready";
+done;
+echo 'Replicated is ready.'
```
sudomateo, line 95: Since the installation script already uses `jq` before this block of code, it might be better to use `jq` instead of `grep`. That way we're parsing JSON and not lines of text.

```suggestion
until [ "$(replicatedctl system status | jq '.Replicated == "ready" and .Retraced == "ready"')" == "true" ]; do
  sleep 1
  echo "Replicated is not ready."
done
echo "Replicated is ready."
```

andybaran: True.  If someone's having problems running this during a "high stress event" they really should be on the phone with our support anyways :)

sudomateo, line 95:
```suggestion
until replicatedctl system status --template '{{and (eq .Replicated "ready") (eq .Retraced "ready")}}' | grep -q true; do
  sleep 1
  echo "Replicated is not yet ready."
done
echo "Replicated is ready."
```

content/source/docs/enterprise/admin/automated-recovery.html.md
```diff
@@ -141,6 +148,13 @@ apt-get install -y jq
...
```
sudomateo, line 156:
```suggestion
until replicatedctl system status --template '{{and (eq .Replicated "ready") (eq .Retraced "ready")}}' | grep -q true; do
  sleep 1
  echo "Replicated is not yet ready."
done
echo "Replicated is ready."
```

content/source/docs/enterprise/admin/automated-recovery.html.md
```diff
@@ -194,6 +208,13 @@ apt-get install -y jq
...
```
sudomateo, line 216:
```suggestion
until replicatedctl system status --template '{{and (eq .Replicated "ready") (eq .Retraced "ready")}}' | grep -q true; do
  sleep 1
  echo "Replicated is not yet ready."
done
echo "Replicated is ready."
```

andybaran:
@sudomateo 's changes above have been committed :)

PR #1351, opened 2020-07-15
Update gitlab-eece.html.md

susiesolis:
customer requested that docs be updated to include "PEM-encoded "


sudomateo:
@susiesolis I think it's best to close this pull request and instead combine these changes with #1349.

susiesolis:
see https://github.com/hashicorp/terraform-website/pull/1349

PR #1353, opened 2020-07-16
Migrate hcloud to registry

paultyng:


paultyng:
Already terraform applied the fastly updates.

PR #1364, opened 2020-07-23
removed misleading note regarding OAuth

andybaran:

## Description

I'd like to remove this note all together as it can be mis-interpreted in a way that seems like we don't cover the OAuth setup.  We do, it's part of the Application Link creation.


sudomateo:
Instead of removing this note, why not just update it to something like this?

> - Note that Bitbucket Server requires both OAuth authentication and an SSH key, both of which are covered in the instructions below.

andybaran:
Brilliant!

PR #1370, opened 2020-07-29
Retry Speculative Plan Documentation

jondavidjohn:
## PR Objective

Adding/updating docs for a new feature

## Description

These docs are for the "Retry Speculative Plans" feature.


content/source/docs/cloud/run/index.html.md
```diff
@@ -73,6 +73,19 @@ There are three ways to run speculative plans:
...
+Sometimes a speculative plan might fail due to an external factor, and you just need to try to run it again.  Retrying a speculative plan is allowed under the following conditions:
+
+- The user viewing the run has [workspace level permissions](https://www.terraform.io/docs/cloud/users-teams-organizations/permissions.html#workspace-permissions) to be able to queue plans.
+- The Plan itself has failed, through an error or being canceled.
+
+When these conditions are met, you will see a "Retry Run" button as shown below.
+
+![Screenshot: Clicking on the "Retry Run" button to trigger a new run](./images/retry.gif)
+
+Retrying the run will create a new run with the same configuration version. If it is a VCS-backed workspace, the pull request interface will receive the status of the new run, along with a link to the new run.
```
nfagerlund, line 87:
```suggestion
If a speculative plan fails due to an external factor, you can run it again using the "Retry Run" button on its page:

![Screenshot: Clicking on the "Retry Run" button to trigger a new run](./images/retry.gif)

Retrying a plan requires permission to queue plans for that workspace. ([More about permissions.](../users-teams-organizations/permissions.html)) Only failed or canceled plans can be retried.

[permissions-citation]: #intentionally-unused---keep-for-maintainers

Retrying the run will create a new run with the same configuration version. If it is a VCS-backed workspace, the pull request interface will receive the status of the new run, along with a link to the new run.
```

- Changing order of info so readers can bail out earlier if they feel like it. 
- Making shorter.
- Formatting info about permissions to match the way we do it elsewhere in TFC docs.
    - The weird link reference to nowhere is so we can do mass updates easier, if we ever change the permissions model in a major way (like we did that one time). 
    - No need to refer to workspace-level perms, especially since there are org-level perms that should also allow this. (Right?)
- No proper-noun caps on "plan."

nfagerlund:
also, bonus points for the gif. :)

PR #1371, opened 2020-07-30
Removing "Gitlab Subgroups unsupported" note from TFC PMR index page

achuchulev:
## Description

As per Asana [10908 TFE modules and GitLab subgroups](https://app.asana.com/0/852068279639184/862570025625495) _Gitlab subgroups_ are now supported in TFC PMR, and the note indicating they aren't was removed from [Publishing Modules to the Terraform Cloud Private Module Registry](https://www.terraform.io/docs/cloud/registry/publish.html) page with the following [PR](https://github.com/hashicorp/terraform-website/pull/1215/files/31110d70fb254b2887c95e77bc6fd6ce9ba0aa7b) however the `unsupported note` is still present on the [index](https://www.terraform.io/docs/cloud/registry/index.html) page, so removing the statement from there will fix the inconsistency in the documentation.


nfagerlund:
Excellent, thank you!

PR #1396, opened 2020-08-12
Adding further context to Policies and Policy Sets docs

hcrhall:
## PR Objective

## Description

This PR expands the contents of the Sentinel documentation, in particular, the [Policy and Policy Sets](https://www.terraform.io/docs/cloud/sentinel/manage-policies.html#policies-and-policy-sets) section. This change will provide the reader with an understanding of the differences between individually managed policies and versioned policy sets. 

I have also explicitly calls out that individually managed policies via the UI are a deprecated feature and recommends the use of versioned policy sets instead.

### DIFF
--- a/content/source/docs/cloud/sentinel/manage-policies.html.md
+++ b/content/source/docs/cloud/sentinel/manage-policies.html.md
@@ -20,11 +20,11 @@ Sentinel Policies are rules which are enforced on Terraform runs to validate tha
 [users]: ../users-teams-organizations/users.html
 [workspaces]: ../workspaces/index.html
 
-**Policies** are written using the Sentinel [language](https://docs.hashicorp.com/sentinel/concepts/language). Policies are the guardrails that prevent Terraform runs from performing dangerous actions. Upon evaluation, policies will adhere to a predefined [enforcement level](#enforcement-levels). Policies are either managed individually or as part of a versioned policy set. 
+**Policies** are written using the Sentinel [language](https://docs.hashicorp.com/sentinel/concepts/language). Policies are the guardrails that prevent Terraform runs from performing dangerous actions. Upon evaluation, policies will adhere to a predefined [enforcement level](#enforcement-levels). 
 
-Individual policies allow the policy to be edited directly in the Terraform UI, whereas versioned policy sets are defined as individual policy files and are stored in a supported VCS provider or uploaded via the Terraform API.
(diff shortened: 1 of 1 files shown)


dylanegan:
@hcrhall reads well to me so far. One thing that might be good to call out specifically because the UI does it for the latter, is extending "individually managed policies" to "individually managed policies and policy sets".

content/source/docs/cloud/sentinel/manage-policies.html.md
```diff
@@ -20,7 +20,11 @@ Sentinel Policies are rules which are enforced on Terraform runs to validate tha
 [users]: ../users-teams-organizations/users.html
 [workspaces]: ../workspaces/index.html
 
-**Policies** consist of a Sentinel policy file and an enforcement level.
+**Policies** are written using the Sentinel [language](https://docs.hashicorp.com/sentinel/concepts/language). Policies are the guardrails that prevent Terraform runs from performing dangerous actions. Upon evaluation, policies will adhere to a predefined [enforcement level](#enforcement-levels). Policies are either managed individually or as part of a versioned policy set. 
+
+Individual policies allow the policy to be edited directly in the Terraform UI, whereas versioned policy sets are defined as individual policy files and are stored in a supported VCS provider or uploaded via the Terraform API.
+
+-> **Note:** Individually managed policies is a deprecated feature in Terraform Cloud and recommend versioned policy sets as the primary method of enforcing policies.
```
nfagerlund, line 27: How about making this recommendation stronger, and mentioning the correct interface before the incorrect interface? 

```suggestion
**Policies** are written using the Sentinel [language](https://docs.hashicorp.com/sentinel/concepts/language). Policies are the guardrails that prevent Terraform runs from performing dangerous actions. Upon evaluation, policies will adhere to a predefined [enforcement level](#enforcement-levels). 

Policies are managed as parts of versioned policy sets, which allow individual policy files to be stored in a supported VCS provider or uploaded via the Terraform Cloud API.

-> **Note:** It's also possible to manage policies and policy sets individually. However, this is a deprecated feature in Terraform Cloud, and we recommend always using versioned policy sets to manage policies. 
```

hcrhall: Thank @nfagerlund this is a much better approach

dylanegan (approved):

PR #1403, opened 2020-08-12
Add analytics to navigation

jescalan:
## PR Objective

## Description

This PR anonymously tracks what parts of the navigation are being used more/less, in order to get a better idea of what links users are clicking vs not.

### DIFF
--- a/content/source/assets/javascripts/analytics.js
+++ b/content/source/assets/javascripts/analytics.js
@@ -29,7 +29,7 @@ document.addEventListener('turbolinks:load', function() {
   Array.prototype.slice
     .call(
       document.querySelectorAll(
-        '.navbar-static-top .navigation-links a[data-track]'
+        '.navbar-static-top .navigation-links a[data-track], .sidebar-nav a[data-track]'
       )
     )
     .map(function(el) {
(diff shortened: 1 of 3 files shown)


content/source/assets/javascripts/analytics.js
```diff
@@ -24,4 +24,17 @@ document.addEventListener('turbolinks:load', function() {
...
   })
+
+  // track navigation link clicks
+  Array.prototype.slice
+    .call(
+      document.querySelectorAll(
+        '.navbar-static-top .navigation-links a[data-track]'
+      )
+    )
+    .map((el) => {
```
nfagerlund, line 35: UNFORTUNATELY, ES6 lambda syntax will cause the "sprockets" build pipeline to barf. 😞 It might not explode in local dev mode, but it's caused breakages in the past for prod builds.

nfagerlund:
Other than those, this looks good! Question -- are these values for the `data-track` attribute hardcoded in our analytics? I'm planning to refactor the top nav sections a bit (after an RFC), so I'm curious what all I'll need to update.

content/source/assets/javascripts/analytics.js
jescalan: Ah man I forgot about that! Good callout here, I'll revert those back to normal functions. Thank you!

jescalan:
@nfagerlund I just made the identifiers up for this PR, so if they change that is ok I think - if we're pulling any statistics here we'll just have to take note of the day they change and divide our stats up that way!

jescalan:
Thank you @nfagerlund for the review! I think I covered everything with these last two changes, let me know if this looks ok

PR #1409, opened 2020-08-15
Extend/Resource/Import: Fix Code Block Escape Format

Tensho:
## Before

https://www.terraform.io/docs/extend/resources/import.html#resource-documentation-implementation

Screenshot 2020-08-15 at 11 24 44

## After

https://github.com/hashicorp/terraform-website/blob/f2fe498f4500de35b1459ac29e8b0a9fad52d8ca/content/source/docs/extend/resources/import.html.md#resource-documentation-implementation

Screenshot 2020-08-15 at 11 16 11

## PR Objective

- [ ] Fixing inaccurate docs


nfagerlund:
SIGH. Commonmark spec says the original code SHOULD work, but I guess whatever old version of redcarpet we're using can't handle it. 😑 Thanks for the fix!

PR #1414, opened 2020-08-19
Add warning about overwriting a default Terraform version

justinretzolk:
## PR Objective

- [x] Clarifying behavior

## Description

We often see users attempt to add a customized Terraform version (say, from using `terraform-bundle`) by overwriting the default Terraform versions with the custom download URL and checksum. These values are regularly overwritten, and thus result in undesirable behavior. This PR is intended to more clearly indicate that overwriting the default versions is not supported, and that a custom version with a different name should instead be used.


content/source/docs/enterprise/admin/resources.html.md
```diff
@@ -89,6 +89,8 @@ Terraform Enterprise ships with a default list of Terraform versions. However, t
 
 To add a new version of Terraform, click **Terraform Versions** in the left menu and **Add Terraform Version** in the upper right, then provide the version number, Linux 64-bit download URL, and SHA256 checksum of the binary. Set the status to Beta to make the version available to site administrators, or Enabled to add it for everyone.
 
+~> **Important:** The versions of Terraform that are shipped with Terraform Enterprise releases are fixed and regularly checked to validate that they have not been altered. To ensure that the download URL and checksum are not overwritten when using an alternative to a default Terraform version, add a new version with a different name.
```
sudomateo, line 92: Some wording changes. @nfagerlund can probably make this even better.

```suggestion
~> **Important:** Terraform Enterprise ships with a default list of Terraform versions. Any modifications to these default Terraform versions will be overwritten. As such, it is recommended to create new Terraform versions instead of modifying the default Terraform versions.
```

content/source/docs/enterprise/admin/resources.html.md
```diff
@@ -89,6 +89,8 @@ Terraform Enterprise ships with a default list of Terraform versions. However, t
 
 To add a new version of Terraform, click **Terraform Versions** in the left menu and **Add Terraform Version** in the upper right, then provide the version number, Linux 64-bit download URL, and SHA256 checksum of the binary. Set the status to Beta to make the version available to site administrators, or Enabled to add it for everyone.
 
+~> **Important:** The versions of Terraform that are shipped with Terraform Enterprise releases are fixed and regularly checked to validate that they have not been altered. To ensure that the download URL and checksum are not overwritten when using an alternative to a default Terraform version, add a new version with a different name.  
+
 ![screenshot: the Add Terraform Version page](./images/admin-version-add.png)
```
sudomateo, line 94: Since you're in here, would you mind updating both this screenshot and the other `./images/admin-versions.png` in this section? They show an older user interface.

justinretzolk: Updated!

PR #1417, opened 2020-08-21
Add registry link to nav, consolidate sign up and sign in buttons

jescalan:
## PR Objective

## Description

This PR aims to improve two things in the navigation:

1. Link to the registry at the top level - it wasn't linked to before anywhere in the nav and is an important part of the Terraform ecosystem.
2. Consolidate and correct the behavior of the "sign up" and "sign in" buttons. These buttons directly imply that you can sign in to the current website, terraform.io, but in reality take you to the terraform cloud website instead. Additionally, the two buttons were taking up a lot of real estate in the nav bar, and according to analytics, the sign up button was getting very little traffic relative to any of the other links. We can both free up valuable real estate required to add the "registry" link, and make the behavior more clear by consolidating both buttons to a single button that makes it clear that the user would be heading over to terraform cloud. If they are signed in, it would bring them to the dashboard, and if not, they would arrive at the "sign in" page, which also has a clear link to sign up if you don't have an account.

Here's how the new navigation bar looks:

I have retained this structure in the responsive and mobile versions of the nav, and retained the UTM parameter on the terraform cloud external link.

### DIFF
--- a/content/source/layouts/_sidebar.erb
+++ b/content/source/layouts/_sidebar.erb
@@ -8,6 +8,7 @@
   <ul class="nav sidebar-nav">
     <li><a data-track='intro' href="/intro/index.html">Intro</a></li>
     <li><a data-track='learn' href="https://learn.hashicorp.com/terraform/">Learn</a></li>
+    <li><a data-track='registry' href="https://registry.terraform.io">Registry</a></li>
     <li><a data-track='community' href="/community.html">Community</a></li>
     <li><a data-track='enterprise' href="https://www.hashicorp.com/products/terraform/?utm_source=oss&utm_medium=header-nav&utm_campaign=terraform">Enterprise</a></li>
   </ul>
--- a/content/source/layouts/layout.erb
+++ b/content/source/layouts/layout.erb
@@ -78,7 +78,7 @@
 
             <%
               docs_items = <<-EOT
-                <li><a data-track='docs-home' href="/docs/index.html">Docs Home</a></li>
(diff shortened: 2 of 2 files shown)


nfagerlund:
Super happy about this! 

Buuuuuut I have some questions about how we should behave at different sizes. 

With this PR as-is, the Terraform Cloud link/button disappears entirely at "medium" viewport sizes (which is where I usually have my window set). I think it's a pretty important control, so I'd like to keep it always available in _some_ capacity... but we probably need to poke it a bit to figure out exactly _how._ 

I did some poking of my own, and here are the fragments of ideas I came up with: 

- We could shrink the button by making the full word "Terraform" collapse once we pass below the "large" cutoff — either remove it entirely so it's just "Cloud ->", or replace it with a chiclet of the Terraform logo ("**T** Cloud ->"). 
- But that still leaves things too wide for "medium" with the new registry link included. So, maybe we could collapse some other nav items into menus once we hit "medium"? We already squish a bunch of things into an "info" menu once we hit "small" (but before the "extra-small" that swaps the top nav for a burger), so that's a semi-established pattern... maybe we could turn "download" into a menu at that point, and move "github" into there? (Possibly "Enterprise" as well? It's not really a download option per se, but the page you go to seems meant to be part of that purchase journey.)

I'm not sure yet. Thoughts?

nfagerlund:
This ended up happening IRL, in a different PR! And my new team's stale review reminder pinged me about it, so I'm gonna just close it. 😆

PR #1421, opened 2020-08-24
Task/cost estimation update handled priceable resources

anamahalo:
## PR Objective

- [x] Fixing inaccurate docs

## Screenshots

Each resolver will look like the following:

CleanShot 2020-08-24 at 16 06 06

### DIFF
--- a/content/source/docs/cloud/cost-estimation/aws.html.md
+++ b/content/source/docs/cloud/cost-estimation/aws.html.md
@@ -13,7 +13,7 @@ Terraform Cloud can estimate monthly costs for many AWS Terraform resources.
 
 ## Handled Resources
 
-The following are priceable resources:
+Below is a list of resources that cost estimation supports:
 
 * aws_alb
 * aws_autoscaling_group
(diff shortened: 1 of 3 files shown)


content/source/docs/cloud/cost-estimation/aws.html.md
```diff
@@ -11,142 +11,20 @@ Terraform Cloud can estimate monthly costs for many AWS Terraform resources.
...
-| aws_vpc_endpoint_route_table_association | |
-| aws_vpc_endpoint_service | |
-| aws_vpc_ipv4_cidr_block_association | |
-| aws_vpc_peering_connection_accepter | |
-| aws_vpc_peering_connection_options | |
-| aws_vpn_connection_route | |
-| aws_waf_ipset | |
-| aws_waf_rule | |
-| aws_waf_web_acl | |
+The following are priceable resources:
```
mwhooker, line 16: The "priceable resource" term is kind of confusing and something we want to get away from. I think the following is fine:

```suggestion
Below is a list of resources that cost estimation supports:
```

PR #1437, opened 2020-09-03
Remove errant single quotes

justinretzolk:
## PR Objective

- [x] Fixing inaccurate docs

## Description

Fairly straightforward! The version of the Automated Recovery sample script that we provide had incorrect single quote, which would lead to a failure to interpolate the variable. This PR removes said single quotes.

### DIFF
--- a/content/source/docs/enterprise/admin/automated-recovery.html.md
+++ b/content/source/docs/enterprise/admin/automated-recovery.html.md
@@ -196,7 +196,7 @@ set -e -u -o pipefail
 
 path=absolute_path_to_directory_of_snapshots
 
-access="--store local --path '$path'"
+access="--store local --path $path"
 
 # jq is used by this script, so install it. For other Linux distros, either preinstall jq
 # and remove these lines, or change to the mechanism your distro uses to install jq.


nfagerlund:
Have you confirmed that the old line breaks and the new one works? On what systems?

If so, then 👍🏼

justinretzolk:
I'll defer to @hashitop for his testing of the script itself, as he noticed this when doing some testing/KB writing. I just happened to be around and had enough free time to actually open this PR 😅.

Generally speaking though, the single quotes will stop interpolation, which is why I figured it was safe enough to get this opened without _personally_ testing it:

```
➜ export test="this_is_a_test"
➜ echo '$test'
$test
➜ echo $test
this_is_a_test
```

justinretzolk:
Ah, yes! That lines up with what Top had pointed out - I don't want to speak out of turn here, but I believe the issue was that the single quotes being around the interpolated value still caused an issue. When you're around, mind expanding on what you saw, @hashitop ?

(just to show that you're _completely_ right and I _should_ have done some additional testing on my own!):

```
➜ echo "Is this a test? '$test'"
Is this a test? 'this_is_a_test'
```

justinretzolk:
I was able to do some testing on this and the single quotes do _not_ appear to have any ill effect. Closing this PR.

hashitop:
Yes, it will pass the string with quotes to replicated and the restoration stalled, when open up UI and click browse to try to find snapshot with the string surrounded by the single quotes, the snapshots won't show up.

I can share you the screenshot if that helps

hashitop:
Screen Shot 2020-09-07 at 11 02 55 am

@justinretzolk , I took those few lines from the existing script on the document and ran them with/without the single quotes

nfagerlund:
@hashitop Thanks for the screenshot and the confirmation! If that's what it is, that's what it is. 👍🏼

PR #1441, opened 2020-09-08
Fix list item numbers in okta.html

nfagerlund:
## PR Objective

## Description

For multi-paragraph list items, every paragraph after the first one must be
indented exactly four spaces from the starting point of that list item;
otherwise redcarpet loses track of the numbering.

<details><summary>screenshot</summary>

image

</details>


dylanegan (approved):
👨‍🍳 😗 🖐️

Juansbulbisful:
Same format just noticible

PR #1446, opened 2020-09-10
Move the upgrade guide under our guides section.

paddycarver:
Our guide on upgrading to the standalone Terraform SDK was hanging out
on its own in the sidebar. This just moves it under our guides section,
both in the sidebar and in URL. A redirect is added to keep the old link
working.


nfagerlund (approved):
Seems good. 👍🏼 

(Obligatory reminder that sidebar_current doesn't do anything anymore and can be safely omitted!)

PR #1448, opened 2020-09-12
Update Providers section of testcase docs

gnarlex:
The `terraform.ResourceProvider` no longer exists in v2 of the terraform-plugin-sdk and can therefore not be used anymore for the providers field.

Use `*schema.Provider` instead (as used by the [AWS plugin](https://github.com/terraform-providers/terraform-provider-aws/blob/4ac98ce9911f302cb7181db28d4a8772b5e67112/aws/provider_test.go#L34) for example).

## Labels

- [x] inaccuracy
- [ ] clarification
- [x] new docs

### DIFF
--- a/content/source/docs/extend/testing/acceptance-tests/testcase.html.md
+++ b/content/source/docs/extend/testing/acceptance-tests/testcase.html.md
@@ -144,10 +144,10 @@ func testAccPreCheck(t *testing.T) {
 ### Providers 
 
 **Type:**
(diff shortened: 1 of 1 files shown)


judithpatudith (approved):
Builds fine and the link works :)

judithpatudith:
@im2nguyen can you validate that that this is a good change to make?

im2nguyen (approved):
This looks great, thanks for creating this PR @gnarlex!

PR #1455, opened 2020-09-25
Sentinel: Add details on updated provider_name field for TF 0.13

vancluever:
Starting with Terraform 0.13, the provider_name field in the resource
collections in the tfplan/v2 and tfstate/v2 imports will contain the
fully qualified path to the resource in the Terraform Registry. Example:
registry.terraform.io/hashicorp/null for the null provider.

This change adds a note for this in the tfplan/v2 and tfstate/v2
documentation pages.

### DIFF
--- a/content/source/docs/cloud/sentinel/import/tfplan-v2.html.md
+++ b/content/source/docs/cloud/sentinel/import/tfplan-v2.html.md
@@ -334,9 +334,9 @@ An element contains the following fields:
   where a provider offers a resource type whose name does not start with its own
   name, such as the `googlebeta` provider offering `google_compute_instance`.
 
--> Starting with Terraform 0.13, the `provider_name` field will contain the
-_full_ source address to the provider in the Terraform Registry. Example:
-`registry.terraform.io/hashicorp/null` for the null provider.
+    -> **Note:** Starting with Terraform 0.13, the `provider_name` field contains the
+    _full_ source address to the provider in the Terraform Registry. Example:
+    `registry.terraform.io/hashicorp/null` for the null provider.
 
 * `deposed` - An identifier used during replacement operations, and can be used
   to identify the exact resource being replaced in state.
--- a/content/source/docs/cloud/sentinel/import/tfplan-v2.html.md
+++ b/content/source/docs/cloud/sentinel/import/tfplan-v2.html.md
@@ -333,6 +333,11 @@ An element contains the following fields:
   allows the provider to be interpreted unambiguously in the unusual situation
   where a provider offers a resource type whose name does not start with its own
   name, such as the `googlebeta` provider offering `google_compute_instance`.
+
+-> Starting with Terraform 0.13, the `provider_name` field will contain the
+_full_ source address to the provider in the Terraform Registry. Example:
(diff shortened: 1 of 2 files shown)


rberlind (commented):
I think the notes need to be directly inside the text for `provider_name` to make sure people see it.

vancluever:
This would be rather wordy for a bullet and needs to be visually notable so that it's seen and not just passed over if someone is perusing documentation quickly. The markup here will ensure that it shows up in a blue info box.

We could move the box right below the bullet for `provider_name` instead?

rberlind:
Yes, I think putting the box right after `provider_name` would be better to make sure it is seen.

nfagerlund (commented):
In the Terraform language docs, we try to consistently refer to these registry paths as "source addresses." (Cf. https://www.terraform.io/docs/configuration/provider-requirements.html#source-addresses) I'd recommend doing the same here. 

Otherwise, lgtm.

content/source/docs/cloud/sentinel/import/tfplan-v2.html.md
```diff
@@ -333,6 +333,11 @@ An element contains the following fields:
   allows the provider to be interpreted unambiguously in the unusual situation
   where a provider offers a resource type whose name does not start with its own
   name, such as the `googlebeta` provider offering `google_compute_instance`.
+
+-> Starting with Terraform 0.13, the `provider_name` field will contain the
+_full_ source address to the provider in the Terraform Registry. Example:
+`registry.terraform.io/hashicorp/null` for the null provider.
```
nfagerlund, line 339:
```suggestion
    -> **Note:** Starting with Terraform 0.13, the `provider_name` field contains the
    _full_ source address to the provider in the Terraform Registry. Example:
    `registry.terraform.io/hashicorp/null` for the null provider.
```

- Adding the conventional "**Note:**" prefix we use for blue callout boxes. 
- Present tense, since 0.13 is _on the loose._
- Fixing formatting: indenting by exactly four spaces will keep the note associated with its parent list item, instead of breaking this into two separate unordered lists with a paragraph between them.

content/source/docs/cloud/sentinel/import/tfstate-v2.html.md
```diff
@@ -142,6 +142,11 @@ An element in the collection has the following values:
   allows the provider to be interpreted unambiguously in the unusual situation
   where a provider offers a resource type whose name does not start with its own
   name, such as the `googlebeta` provider offering `google_compute_instance`.
+
+-> Starting with Terraform 0.13, the `provider_name` field will contain the
+_full_ source address to the provider in the Terraform Registry. Example:
+`registry.terraform.io/hashicorp/null` for the null provider.
```
nfagerlund, line 148:
```suggestion
    -> **Note:** Starting with Terraform 0.13, the `provider_name` field contains the
    _full_ source address to the provider in the Terraform Registry. Example:
    `registry.terraform.io/hashicorp/null` for the null provider.
```

nfagerlund:
image

Local build screenshot of my markup tweaks, just to keep myself honest about whether I'm remembering our markdown renderer's behavior right. 😝

PR #1458, opened 2020-09-30
Clarify overriding Sentinel policy

rberlind:
A large customer was uncertain whether overriding a Sentinel policy only affected the current run or all future runs. I actually found the current language clear enough since it said "and continue the execution of the run". But I added sentence "This will not have any impact on future runs." to satisfy them.

## Labels
- [ ] inaccuracy
- [x ] clarification
- [ ] new docs


nfagerlund (approved):
Sure, why not.

PR #1472, opened 2020-10-19
add instructions for using custom providers in TFC/E with TF 0.13

radditude:
The instructions for installing and using custom providers in Terraform Cloud and Enterprise needed some updating in this brave new world of Terraform 0.13+. I've added a "Terraform 0.13 and later" section and left the old docs mostly unchanged under the "Terraform 0.12 and earlier" heading - there's some repetition between the two, but that seems like an acceptable price to pay in order to make the page easier to skim.

Additionally, the directory to use when including custom providers within a VCS repo was incorrect; it should be `terraform.d`, not `.terraform`.

localhost_4567_docs_cloud_run_install-software html


chrisarcand (approved):
The only thing I'd add is that any consideration of network mirrors and specifying [Source addresses](https://www.terraform.io/docs/configuration/provider-requirements.html#source-addresses) should be added for 0.13 - whether you can do it, or that you explicitly cannot with TFC at this time.

radditude:
Good call! Did some more testing and added notes on both.

PR #1477, opened 2020-10-22
Add custom Segment event for 404

mwickett:
Converts the `404.html.md` to `erb` to allow for a `script` tag.

The script fires a Segment `track` event to capture the path and referrer (if any) when the 404 page is displayed. This will improve our ability to fixup any 404s. This will provide a baseline as we move towards migration and make spotting regressions with redirects easier.

Testing this is a bit challenging, but I did see the event showing up in the Segment debugger when hitting `localhost:4567/404.html`.


jescalan (approved):
LGTM

content/source/404.html.erb
```diff
@@ -0,0 +1,18 @@
...
+
+<h1>Page Not Found</h1>
+
+<p>Sorry, the page you tried to visit does not exist. This could be our fault, and if so we will fix that up right away.</p>
+
+<p>Please go back to get back on track.</p>
+
+<script>
+  // Fire a custom event to Segment to surface paths that are causing 404s
+  window && window.analytics && window.analytics.track('404 Response')
```
mwickett, line 17: Segment captures the relevant info like referrer, path etc. that we want here. This event name (`404 Response`) is consistent with what we're using with our other web properties.

nfagerlund:
@mwickett OH NO, I filed (and merged) https://github.com/hashicorp/terraform-website/pull/1482 before seeing that you had this in the hopper. 😱 Did it as part of another project I was heads-down on, so I wasn't staying on top of PRs. 

Which implementation do you and the team prefer? I'm perfectly happy to overwrite my version with yours if you want.

kylemac:
@mwickett - lets close or resolve this puppy so it doesn't linger around

mwickett:
Ah, thanks @kylemac  - sorry I missed your comment @nfagerlund!

No worries - I'm working with @BrandonRomano to standardize on the event signature here. The issue with passing the URL as the event name is that it creates table bloat in our data warehouse.

I'll solidify our approach and update this PR to reflect our approach going forward.

mwickett:
Closing this out in favor of Nick's #1482 for now. @BrandonRomano is going to be taking a look at how we handle this tracking and very likely propose some changes in the future.

PR #1478, opened 2020-10-22
Fix broken link in recommended practices doc

da3mon-01:
The examples were moved to terraform providers aws repository, so the documentation here points at a broken link.

## Labels

- [ ] inaccuracy
- [ ] clarification
- [ ] new docs


nfagerlund:
Good catch, thank you. 👍🏼

PR #1481, opened 2020-10-28
Icons: Replace "outbound link" icon for aesthetic reasons

nfagerlund:
This SVG file is a one-off, but it came from the standard library of icons used
in the new web platform (which we aren't on yet).

image


judithpatudith (approved):
So great!!

PR #1482, opened 2020-10-28
Site behavior: Add GA tracking segment for 404 referrers

nfagerlund:
Copied largely from
https://github.com/hashicorp/nomad/blob/master/website/pages/not-found/index.jsx#L13

This should track where links that 404 are coming from, which will hopefully help reduce the risk of moving pages in the Terraform docs. (We already redirect, but this is an extra safety net.)


jescalan (approved):
👌

PR #1483, opened 2020-10-28
Styles: Make code spans usable in sidebar navs

nfagerlund:
This is in preparation for https://github.com/hashicorp/terraform/pull/26723, which will use code spans to style certain sidebar nav items (mostly `terraform` command pages). 

Without this, they were getting some goofy Bootstrap defaults. With it, they
look like normal links in a different font.

BEFORE: 😑

image

AFTER: 😄

image

### DIFF
--- a/content/source/assets/stylesheets/_docs.scss
+++ b/content/source/assets/stylesheets/_docs.scss
@@ -20,6 +20,15 @@
     a { display: inline; }
   }
 
+  // Code spans in nav links shouldn't get bootstrap default code span styles;
+  // make them just inherit everything.
+  code {
+    padding: unset;
+    color: unset;
+    background-color: unset;
+    border-radius: unset;
+  }
+
   // hide an outdated implementation of expand/collapse link
   a.subnav-toggle {
     display: none;


im2nguyen (approved):

robin-norwood (commented):
This is valid CSS and looks like it works two questions, though:- it seems a little weird to me to have text styles in a side nav, though?

robin-norwood (approved):

nfagerlund:
@robin-norwood 

For 2: This is SCSS rather than plain CSS, so the block it's nested in (hidden above the fold on the diff) becomes part of the selector via a descendant combinator (the one where there's just a space between two selectors). So the real selector this compiles to is `#docs-sidebar code`. 

For 1: Well, it's what made sense at the time. 😅 I've got some pages whose names are just a verbatim computer word (`apply`, e.g.), and they looked weird alongside page titles that were in title case. We might iterate on this, and look for other ways to communicate that this is the name of a command, but I liked this better than any of the other approaches I messed with (most of which added more words to the title and ended up noisy).

PR #1485, opened 2020-10-29
Add support for how to debug gRPC terraform plugins

remyleone:
I've tried to contribute this content first to https://github.com/hashicorp/terraform/pull/26582 and was redirected here.

I don't clearly understand how I can debug terraform communications between terraform core and the provider at a low level. I would like to add content about how to get a better grasp of the gRPC communication between terraform and the providers.

Plus I would also like to enhance the migration guide https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html
with an example about how to use the debug mode with a provider. It is not clear for me at the moment how to get a setup working. Could we start working on something together?

## Labels

- [ ] inaccuracy
- [ ] clarification
- [X] new docs

### DIFF
--- a/content/source/docs/extend/how-terraform-works.html.md
+++ b/content/source/docs/extend/how-terraform-works.html.md
@@ -151,3 +151,28 @@ acceptable version of a given provider is installed elsewhere,
 
 [0]: https://en.wikipedia.org/wiki/Static_build#Static_building
 [1]: https://golang.org/
+
+### Debug gRPC
+
+When a given provider is running for instance with: `go run main.go` it will start a process that will listen on a UNIX socket.
+
+```
+$ go run main.go -debuggable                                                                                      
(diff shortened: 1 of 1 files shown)


paddycarver:
I think this is superseded by https://www.terraform.io/docs/extend/debugging.html#debugger-based-debugging and no longer necessary, so I'm going to close it out. If you believe it's still necessary, feel free to open an issue and tag me in it and we can talk about what you're looking for and how the current documentation doesn't provide it. Thanks for the contribution and writing all this out!

remyleone:
I think it would still be useful to have a word about how gRPC based debugging works.

paddycarver:
I agree with you, and I would love an issue on https://github.com/hashicorp/terraform-plugin-sdk/issues/new about gRPC debugging: what you want from it, the situations you'd reach for it in, what you'd be trying to accomplish, etc. I'm super interested in this.

My concern here is that the gRPC debugging experience you've documented is complicated and fiddly and relies on implementation details--knowing the socket path, for example--and the config, state, and plan are going to be msgpack-encoded and unreadable, anyways, which kinda limits the utility of it. I'd love to enable gRPC debugging, I think we just want to do it properly, if that makes sense.

PR #1488, opened 2020-10-31
Use docker mirror

mdeggies:
CHANGELOG: no impact

Dockerhub is going to rate limit unauthenticated pulls on November 1st. IPs from CI machines (with the exception of github actions) will be near their limit all the time. We're moving projects to use a public un-rate-limited Mirror of these images instead. Let me know if you have any q's, otherwise feel free to merge when you can! 

## Labels

- [ ] inaccuracy
- [ ] clarification
- [ ] new docs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 references:
   images:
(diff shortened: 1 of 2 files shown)


nfagerlund:
Excellent, thank you! I thought I had a question about this, but you already answered it in your description. 

I'm gonna go ahead and go for it.

PR #1496, opened 2020-11-11
Layout: Add outbound link to registry in Docs menu

nfagerlund:
This proposal could really use a better alternative to the CSS hackery that fixes the colors of the outbound link icon, so I'm filing it as a draft in the hopes that someone can suggest something. 😅

Context: 

- You can't see it in this PR, but we're in the process of renaming the current "Registry" link in the docs menu to "Publishing Modules & Providers" (because it goes to the docs about publishing modules and providers on the registry). So that won't be a name conflict. 
- Similar idea as https://github.com/hashicorp/terraform-website/pull/1417, except the link is going in the docs menu. That's not really the correct place for it, but it's temporarily expedient until we can get some kind of consensus about what to do with the massive TFC links on the far right. 
- I got that "outbound link" SVG from @BrandonRomano, and I think it came from a collection of similar icons on our Figma. 
- The whole problem here is that the "download", "github", and "menu dropdown thingie" SVGs in the header all get their visible shapes from a fill property, but the outbound link SVG uses a stroke property, which we need to override to match the colors on both the homepage and the rest of the site. I couldn't see a way to do that when using an img tag, so I used the inline SVG helper, and that brings us to where we are today. 

It looks fine, I just hate the code.

Screen Shot 2020-11-11 at 3 07 38 PM Screen Shot 2020-11-11 at 3 07 50 PM

### DIFF
--- a/content/source/assets/stylesheets/_global.scss
+++ b/content/source/assets/stylesheets/_global.scss
@@ -29,3 +29,10 @@ body {
 .sidebar .sidebar-nav li a {
   line-height: 18px;
 }
+
+.sidebar .sidebar-nav li a .hack-outbound-svg {
+  fill: none !important;
+  path {
+    stroke: $sidebar-link-color !important;
+  }
+}
--- a/content/source/assets/stylesheets/_header.scss
+++ b/content/source/assets/stylesheets/_header.scss
@@ -62,6 +62,13 @@
           svg {
             fill: $header-link-color-hover;
           }
+
+          .hack-outbound-svg {
+            fill: none !important;
+            path {
+              stroke: $header-link-color-hover !important;
+            }
+          }
         }
       }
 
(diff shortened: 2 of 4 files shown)


jescalan:
I would be curious to get @swarnap 's take on this one, last we spoke I think she wanted this as a top-level nav item

swarnap:
We are already working on the Terraform io content refresh (which will go live with 0.14 GA) which will bring Registry to the top-nav.  I will defer it to web dev team if it makes sense to make the UI change (indicator for an outbound link) in the meantime.

nfagerlund:
@swarnap If you're getting a top-level registry link into the 0.14 GA, then that's good enough for me — that's when I was planning this stopgap to go into place, but if you're that close then I'll just leave it be. 👍🏼

PR #1497, opened 2020-11-12
Fix changed Cost Estimation value in API docs

chrisarcand:
We recently made a change where Cost Estimation is now enabled by default in organizations; this has the added caveat that on Terraform Cloud, the organization must also be entitled to use Cost Estimation. On Terraform Cloud, it is always enabled.

## Labels

- [x] inaccuracy
- [x] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/cloud/api/organizations.html.md
+++ b/content/source/docs/cloud/api/organizations.html.md
@@ -54,7 +54,7 @@ curl \
       "type": "organizations",
       "attributes": {
         "name": "hashicorp",
-        "cost-estimation-enabled": false,
+        "cost-estimation-enabled": true,
         "created-at": "2017-09-07T14:34:40.492Z",
         "email": "user@example.com",
         "session-timeout": null,
@@ -114,7 +114,7 @@ curl \
     "type": "organizations",
     "attributes": {
       "name": "hashicorp",
-      "cost-estimation-enabled": false,
+      "cost-estimation-enabled": true,
       "created-at": "2017-09-07T14:34:40.492Z",
       "email": "user@example.com",
       "session-timeout": null,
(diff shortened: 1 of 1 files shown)


anamahalo (approved):
Thanks - I think the first portion of the sentence can also be taken out as it's confusing with what is following and what is being added in the parens clarifies (it made more sense standalone defaulted to false).

It might be a good idea to change the sample payloads as well to match as they presently say `false`.

content/source/docs/cloud/api/organizations.html.md
```diff
@@ -165,7 +165,7 @@ Key path                                   | Type    | Default   | Description
 `data.attributes.session-timeout`          | integer |    20160  | Session timeout after inactivity (minutes)
 `data.attributes.session-remember`         | integer |    20160  | Session expiration (minutes)
 `data.attributes.collaborator-auth-policy` | string  | password  | Authentication policy (`password` or `two_factor_mandatory`)
-`data.attributes.cost-estimation-enabled`  | boolean | false     | Whether or not the cost estimation feature is enabled for all workspaces in the organization
+`data.attributes.cost-estimation-enabled`  | boolean | true      | Whether or not the cost estimation feature is enabled for all workspaces in the organization. Defaults to true on Terraform Cloud when the organization is on the Teams & Governance tier (else, the value is false and cannot be changed) and always true on Terraform Enterprise.
```
nfagerlund, line 168: Hmm, yeah, the phrasing of this is tricky. What I _think_ I'm understanding here is: 

- TFC, Team&Gov or Business: defaults to true, can be optionally set to false. 
- TFC, Team or free: locked to false. 
- TFE: locked to true.

If that's the case, then maybe something like this? 

```suggestion
`data.attributes.cost-estimation-enabled`  | boolean | true      | Whether the cost estimation feature is enabled for the organization. If the organization is on the Teams & Governance paid tier or higher, this defaults to `true` but can be optionally disabled. Otherwise, the value (`false` on Terraform Cloud tiers without cost estimation, `true` on Terraform Enterprise) cannot be changed.
```

HMM.

nfagerlund (commented):
I do kind of want to tweak the wording on that, but this was the best I could come up with today.

content/source/docs/cloud/api/organizations.html.md
chrisarcand: The last point isn't true - in TFE, the _default_ is true but can be set to false.

Here's a second attempt to make things a little more clear; the key point is that the default _is_ true, it just has added caveats depending on your platform (TFC must have the right tier, and TFE must be enabled at a higher level as well).

> Whether or not the cost estimation feature is enabled for all workspaces in the organization. Defaults to true. In a Terraform Cloud organization which does not have Teams & Governance features, this value is always false and cannot be changed. In Terraform Enterprise, Cost Estimation must also be enabled in Site Administration.

chrisarcand: I'm going to go with this to proceed, but absolutely feel free to tweak as desired!

nfagerlund: Belatedly: this works for me!

PR #1501, opened 2020-11-13
Add deprecation notice for Bitbucket Server v5.3.0 and below

acespacecase:
Support for Bitbucket Server v5.3.0 and below is getting phased out and adding a deprecation notice is the current step. Soon, information for Bitbucket Server v5.3.0 and below will be removed from docs completely.

Screen Shot 2020-11-13 at 9 40 04 AM

Screen Shot 2020-11-13 at 9 40 12 AM


content/source/docs/cloud/vcs/bitbucket-server.html.md
```diff
@@ -18,13 +18,13 @@ Configuring a new VCS provider requires permission to manage VCS settings for th
...
+-> **Version note:** Terraform Cloud supports Bitbucket Server versions 4.9.1 and above, and Bitbucket Data Center versions 5.4.0 and above. HashiCorp does not test older versions of Bitbucket Server with Terraform Cloud, and they might not work as expected. Also note that, although we do not deliberately remove support for versions that have reached end of life (per the [Atlassian Support End of Life Policy](https://confluence.atlassian.com/support/atlassian-support-end-of-life-policy-201851003.html)), our ability to resolve customer issues with end of life versions might be limited.<br /><br />Terraform Cloud will soon deprecate support for Bitbucket Server versions 5.3.0 and below.
```
nfagerlund, line 21: This takes a bit of thinking to understand, but I think we can make it easier by just changing the order of some of this info:

```suggestion
-> **Version note:** Terraform Cloud supports Bitbucket Server and Bitbucket Data Center versions 5.4.0 and above. <br /><br />Support for Bitbucket Server versions between 4.9.1 and 5.3.7 is deprecated. Terraform Cloud currently works with these versions, but will stop supporting them in the near future. <br /><br />HashiCorp does not test older versions of Bitbucket Server with Terraform Cloud, and they might not work as expected. Also note that, although we do not proactively remove support for versions that have reached end of life (per the [Atlassian Support End of Life Policy](https://confluence.atlassian.com/support/atlassian-support-end-of-life-policy-201851003.html)), our ability to resolve customer issues with end of life versions might be limited.
```

acespacecase: 🙌 This is a great improvement, thanks for your help!

lafentres (approved):

PR #1505, opened 2020-11-16
Multipool agents

caseylang:
Updating the docs for Agents to include information about multipool support.

- Adds edits to the agents page for multiple pools
- Adds edits to the agents API page for multiple pools
- Moves "Terraform Cloud Agents" out from underneath "Workspaces" in the sidebar. I found that to be really confusing and buried, when it's supposedly a major feature; it also makes just as much sense as Splunk/Service Now/etc integrations at that level. Added a redirect to not break existing links out in the wild.
- Updated screenshots throughout


lafentres (approved):

nfagerlund (approved):
Hey, thanks for this! I had two issues with it, but I pushed a commit that addresses them -- they were markdown structure things that would have been way too annoying to show in the "suggestions" review interface. Namely:

- List item indentation was messed up so that every step was step 1. To fix this, I made sure every image that was supposed to be part of a numbered step was indented by exactly four spaces. (An out-dented image that starts at character 0 of its line will reset list item numbering.) 
- Header levels were confusing and overly deep. Make sure to take a look at this one before merging; basically, I decided a top-level "configuration" section wasn't desirable, and split it into multiple top-level sections instead. The division I used was:
    - "Managing pools" is stuff you do in your org settings. Moved token revocation under here. 
    - "Managing agents" happens on your linux VM or docker host. 
    - "Configuring workspaces to use the agent" happens in each workspace's settings.

Here's the new outline of header hierarchy, from the quick nav menu: 

image

nfagerlund:
Oh, two more things: 

- You'll need to fix the merge conflict in redirects.txt, but that's easy, it's just because two people added stuff to the very end of the file. Just keep both parts. 
- I noticed our table of network access doesn't mention the Registry for provider downloads -- that's outdated, right? I won't block this PR on it because that text wasn't touched, but it'd be good to clear that up soonish. 

image

caseylang:
Thanks a lot @nfagerlund! Chris pushed both those fixes. We'll let the tests finish and get this merged

PR #1508, opened 2020-11-17
Mention overriding soft-mandatory policies in CLI

rberlind:
Users can override soft-mandatory Sentinel policies when using the Terraform CLI with the remote backend. They are shown the following:
```
Do you want to override the soft failed policy check?
  Only 'override' will be accepted to override.

  Enter a value:
```

They must type "override" to override.  Note that "yes" will not work.

## Labels

- [ ] inaccuracy
- [x] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/cloud/run/cli.html.md
+++ b/content/source/docs/cloud/run/cli.html.md
@@ -212,9 +212,6 @@ Do you want to override the soft failed policy check?
 
   Enter a value: override
 ```
-If you do override, you will then be prompted to apply the run. 
-
-Hard mandatory checks cannot be overridden and they prevent `terraform apply` from applying changes.
 
 ```
 $ terraform apply
--- a/content/source/docs/cloud/run/cli.html.md
+++ b/content/source/docs/cloud/run/cli.html.md
@@ -182,7 +182,14 @@ Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
 
 ## Sentinel Policies
 
-If the specified workspace uses Sentinel policies, those policies will run against all speculative plans and remote applies in that workspace. The policy output will be available in the terminal. Soft mandatory checks can be overridden by users with the [Manage Policies](docs/cloud/users-teams-organizations/permissions.html#manage-policies) permission by typing "override" in response to the prompt. (Note that "yes" will not work.)
+If the specified workspace uses Sentinel policies, those policies will run against all speculative plans and remote applies in that workspace. Policy output is shown in the terminal. 
+
+Failed policies can pause or prevent an apply, depending on the enforcement level:
+
+- Hard mandatory checks cannot be overridden and they prevent `terraform apply` from applying changes.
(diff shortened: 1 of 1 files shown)


content/source/docs/cloud/run/cli.html.md
```diff
@@ -182,7 +182,39 @@ Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
 
 ## Sentinel Policies
 
-If the specified workspace uses Sentinel policies, those policies will run against all speculative plans and remote applies in that workspace. The policy output will be available in the terminal. Hard mandatory checks cannot be overridden and they prevent `terraform apply` from applying changes.
+If the specified workspace uses Sentinel policies, those policies will run against all speculative plans and remote applies in that workspace. The policy output will be available in the terminal. Soft mandatory checks can be overridden by users with the [Manage Policies](docs/cloud/users-teams-organizations/permissions.html#manage-policies) permission by typing "override" in response to the prompt. (Note that "yes" will not work.)
```
nfagerlund, line 185:
```suggestion
If the specified workspace uses Sentinel policies, those policies will run against all speculative plans and remote applies in that workspace. Policy output is shown in the terminal. 

Failed policies can pause or prevent an apply, depending on the enforcement level:

- Hard mandatory checks cannot be overridden and they prevent `terraform apply` from applying changes.
- Soft mandatory checks can be overridden by users with permission to [manage policies](docs/cloud/users-teams-organizations/permissions.html#manage-policies). If your account can override a failed check, Terraform will prompt you to type "override" to confirm. (Note that typing "yes" will not work.) If you override the check, you will be prompted to apply the run (unless auto-apply is enabled).

[permissions-citation]: #intentionally-unused---keep-for-maintainers
```

content/source/docs/cloud/run/cli.html.md
```diff
@@ -182,7 +182,39 @@ Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
...
+FALSE - my-policy.sentinel:1:1 - Rule "main"
+
+Do you want to override the soft failed policy check?
+  Only 'override' will be accepted to override.
+
+  Enter a value: override
+```
+If you do override, you will then be prompted to apply the run. 
+
+Hard mandatory checks cannot be overridden and they prevent `terraform apply` from applying changes.
```
nfagerlund, line 217:
```suggestion
```

nfagerlund (changes requested):
Thanks for this! 

I went into the preview to evaluate this. The two code blocks that show policy check output are too similar to both be valuable, and they make the section too visually noisy. 

Could you please remove the second code block? That way we'll show the more complicated case (the override prompt), and readers can just intuit that the simpler case looks much the same but without the option to interact.

I also have some suggestions about changing the ordering of information a bit, to reduce the amount of stuff the reader needs to remember while they skim the code block.

rberlind:
@nfagerlund : I committed your proposed revisions and removed the second Sentinel policy failure example.  I think that should cover all your proposed changes.

nfagerlund (approved):
Indeed!

PR #1517, opened 2020-11-18
Adopt registry docs into terraform-website repo

nfagerlund:
In recent months Terraform core has had a somewhat complicated branching and backporting strategy, which has had occasional bad effects on teams that need to maintain docs that have no real connection to Terraform core's release cadence. In particular, the Registry team has been hitting some "disappearing docs" issues due to the requirement to backport fixes to three different branches upon merge to main (`v0.13`, `v0.14`, and `stable-website`). 

Well, the registry docs don't properly belong in the hashicorp/terraform repo in the first place; they were just there because terraform-website initially didn't have any real textual content when we first split the provider repos out. Now that we have a convention of maintaining docs for web services (Terraform Cloud) in this repo, we should do the same for the Registry. 

After this PR is merged, we'll be free to delete the registry docs from the terraform repo itself.


justincampbell:
This is great! Thank you so much.

alisdair (approved):

PR #1520, opened 2020-11-18
0.14 release task: ship language and cli docs

nfagerlund:
**Draft status: This should be merged immediately after Terraform 0.14 is shipped. It's marked as a draft to prevent early merge.**

This PR adjusts several nav elements to accommodate the newly split Terraform Language and Terraform CLI docs, which will go live on 0.14 ship day. 

- Merging this early won't break the site, but will result in confusing/broken top-level navs. 
- Same deal for merging it late.

So, it basically needs to be merged right after the release CI process finishes, but it's not like we're trying to escape a rolling boulder of doom. Within the same hour should be fine.

### DIFF
--- a/content/source/docs/cloud/guides/recommended-practices/index.html.md
+++ b/content/source/docs/cloud/guides/recommended-practices/index.html.md
@@ -1,6 +1,6 @@
 ---
 page_title: "Index - Terraform Recommended Practices"
-layout: "guides"
+layout: "intro"
 ---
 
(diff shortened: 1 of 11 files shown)


nfagerlund:
@pkolyvas Heya! This PR is part of the 0.14 ship process, so I should make sure it appears in your github mentions in case you need to find it later.

nfagerlund (commented):
Petros will merge this on launch morning!

PR #1536, opened 2020-12-02
Remove submodules for some redirected providers

nfagerlund:
Redirects for all of these provider docs are already live on the site, which
means the contents of these submodules are no longer reachable at all. So let's
yank em.


sudomateo (approved):
Consider them yanked!

paultyng:
These were left because in some cases builds depended on them. Let me double check with the teams.

nfagerlund:
Oh, oops. Thank you for checking. 😬

PR #1538, opened 2020-12-03
Fix typo in gitlab-eece documentation

p0pr0ck5:
Fix double-typed "a" in heading.


nfagerlund:
Ah thanks, good catch.

PR #1541, opened 2020-12-05
Fix persistent nasty bug with navigating to anchor links

nfagerlund:
Or, How I Learned to Stop Worrying and Love the Turbolinks.

Turbolinks changes the lifecycle of a normal web page in completely bizarre ways, and I didn't have a good handle on it when I originally implemented some of our features. That had some wild knock-on effects, the most visible of which was that navigating to an in-page anchor was outrageously unreliable. 

Petros reminded me of that earlier today, and I resolved to finally figure out what the heck. So the upshot is: 

- It's not safe to modify the page's layout in any meaningful way within a `turbolinks:load` event handler. The docs encourage you to do that, but it's a TRAP. There's a point earlier in the process where it's safe to move things around, but then you also need to register separate handlers for initial page load and for turbolinks transitions. 
    - So I did that. That fixed the anchor link problem.
- I was accidentally exempting some links from turbolinks behavior, so links from the sidebar felt slower than links from body text. 
- Added a _deliberate_ exemption for anchor links within the same page; turbolinks doesn't need to interact with those at all, the browser can just handle it. This makes anchor links somewhat more resilient, but mostly just saves some wasted HTTP calls. 
- The whole point of turbolinks is to re-use all the work you did when initially setting up a page, rather than throwing it out and starting fresh. So I refactored the JS to not re-create all my event handlers on every page transition.

Anyway, everything should feel a bit less janky now.

### DIFF
--- a/content/source/assets/javascripts/application.js
+++ b/content/source/assets/javascripts/application.js
@@ -51,7 +51,6 @@ document.addEventListener("turbolinks:load", function() {
         if (e.target == this) {
             $(this).toggleClass("active");
         }
-        e.stopPropagation();
     });
     // If the subnav header doesn't link to a different page, use it as a toggle.
     docsSidebar.find("a[href^='#']").on("click", function(e) {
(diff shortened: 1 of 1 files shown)


jescalan (approved):
This looks great. A little tough to read the diff, but I see what happened here and your explanation is clear. Pulled it down and tested locally and everything that I ran through was working perfectly well. Good sleuthing! 💯

PR #1543, opened 2020-12-08
Suggested edit

assareh:
I'd like to suggest that we remove this note from the documentation, as the pricing page for TFCB only has a button to Contact Sales. I feel like having it here is likely to create more frustration than not having it. 

## Labels

- [x] inaccuracy
- [] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/cloud/agents/index.html.md
+++ b/content/source/docs/cloud/agents/index.html.md
@@ -209,7 +209,7 @@ Agent status appear on the **Organization Settings > Agents** page and will cont
 
 Agents are considered active and count towards the usage of the organization's total capacity if they are in the **Idle**, **Busy**, or **Unknown** state. Agents which are in the **Errored** or **Exited** state do not count towards the usage of the total capacity.
 
-The number of active agents you are eligible to deploy is determined by the number of Concurrent Runs your organization is entitled to. [Learn more about Terraform Cloud pricing here](https://www.hashicorp.com/products/terraform/pricing/).
+The number of active agents you are eligible to deploy is determined by the number of Concurrent Runs your organization is entitled to. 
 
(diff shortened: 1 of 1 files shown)


content/source/docs/cloud/agents/index.html.md
```diff
@@ -209,7 +209,7 @@ Agent status appear on the **Organization Settings > Agents** page and will cont
 
 Agents are considered active and count towards the usage of the organization's total capacity if they are in the **Idle**, **Busy**, or **Unknown** state. Agents which are in the **Errored** or **Exited** state do not count towards the usage of the total capacity.
 
-The number of active agents you are eligible to deploy is determined by the number of Concurrent Runs your organization is entitled to. [Learn more about Terraform Cloud pricing here](https://www.hashicorp.com/products/terraform/pricing/).
+The number of active agents you are eligible to deploy is determined by the number of Concurrent Runs your organization is entitled to.
```
vaficionado, line 212:
```suggestion
The number of active agents you are eligible to deploy is determined by the number of Concurrent Runs your organization is entitled to. Agents are available as part of the [Terraform Cloud Business] (https://www.hashicorp.com/products/terraform/pricing/) tier.
```

vaficionado (changes requested):
We need some kind of a tickler on this page that indicates the agents are not available at just any tier. I agree that the wording could be improved here, per Andy's suggestion, but I don't think we can just remove any link to the TFCB collateral. Since we don't have a better comparative landing page that I can think of, the link should stay as-is, but maybe the text can be made a little clearer?

content/source/docs/cloud/agents/index.html.md
assareh: thanks Jon, i agree with your suggested wording.

PR #1546, opened 2020-12-09
Fix some broken links

nfagerlund:


jbonhag (approved):
👍 Thanks, nice catch on those 404s. I saw that all the azuread/azurerm links have redirects in production, so I assume those are being handled by the provider repo.

PR #1551, opened 2020-12-10
CI: Separate the upload and cache warming steps to distinguish broken links from broken builds

nfagerlund:
Our current website deployment CI doesn't distinguish between broken links
(trivial, to be fixed at our convenience) and hosed builds (apocalyptic,
preempts everything). So, let's fix that by displaying their results as separate
steps in the deploy workflow. 

I'm not sure yet whether configuring `linkcheck` (for PR branches) and `website-deploy` (for master) as separate _workflows_ was the right move; could use input from people who spend more time in Circle! 

For context, this PR is probably the first in a series of improvements. I'd also like to: 

- Replace the `wget` spidering with a different linkchecker, which will still perform the task of cache warming but will give much more actionable output. 
- Improve the linkchecking for PRs, possibly moving the build step out into a Vercel preview and then running some form of linkchecker (in a less heavyweight container) against it. 
    - Ideally I'd like PR linkchecking to only report on broken links added _in that PR,_ so that we can insist people take immediate action when they see red; right now, 99% of the red PR checks are non-actionable for the person who submitted the PR, and might not even be coming from the same repository. But I'm still designing how that'll work.


.circleci/config.yml
```diff
@@ -56,15 +56,38 @@ jobs:
...
+    # only run on main branch
+    jobs:
+      - website-build-and-upload:
+          context: static-sites
+          filters:
+            branches:
+              only: master
+      - website-warm-cache-check-links:
           context: static-sites
           filters:
```
alvin-huang, line 89: You can remove the filter here if you want since the dependency on the `website-build-and-upload` step will prevent this job from running on non-master branches.

.circleci/config.yml
```diff
@@ -56,15 +56,38 @@ jobs:
...
+          command: ./scripts/upload.sh
+
+  website-warm-cache-check-links:
+    docker:
+      - image: *MIDDLEMAN_IMAGE
+    steps:
+      # Don't need gems or anything
+      - run:
+          name: Warm cache and check for broken links
+          command: wget --delete-after --level inf --no-directories --no-host-directories --no-verbose --page-requisites --recursive --spider "https://www.terraform.io/"
```
alvin-huang, line 68: Do you want to call `./scripts/warm-cache.sh` here instead?

alvin-huang (approved):
A few minor things but LGTM!

.circleci/config.yml
nfagerlund: Ah, excellent! thank you.

.circleci/config.yml
nfagerlund: I changed that in the latest commit. The initial version of this job ran checkout, then ran the script from the repository... but then I realized we were checking out the entire repository for literally just one `wget` command, and decided we could probably just inline it. That might have been premature optimization, since I'm not sure what Circle does to make checkouts performant! But, since I'm hoping to replace wget with something else here soonish... seems fine either way.

PR #1554, opened 2020-12-12
Fix testAccCheckExampleResourceExists in example codes

budougumi0617:
Hi, I am implementing my terraform provider while referencing community's docs. I appreciate helpful documents, but, I found strange code in below page.

- https://www.terraform.io/docs/extend/best-practices/testing.html#basic-test-to-verify-attributes

The strange code is below function.

```go
// testAccCheckExampleResourceExists queries the API and retrieves the matching Widget.
func testAccCheckExampleResourceExists(n string, widget *example.Widget) resource.TestCheckFunc {
    return func(s *terraform.State) error {
        // find the corresponding state object
        rs, ok := s.RootModule().Resources[n]
        if !ok {
            return fmt.Errorf("Not found: %s", n)
        }

        // retrieve the configured client from the test setup
        conn := testAccProvider.Meta().(*ExampleClient)
        resp, err := conn.DescribeWidget(&example.DescribeWidgetsInput{
            WidgetIdentifier: rs.Primary.ID,
        })

        if err != nil {
            return err
        }

        // If no error, assign the response Widget attribute to the widget pointer
        *widget = *resp.Widget

        return fmt.Errorf("Widget (%s) not found", rs.Primary.ID)
    }
}
```

`testAccCheckExampleResourceExists` is TestCheckFunc, bud it method always returns `error`.
I believe that `testAccCheckExampleResourceExists` returns nil if there is `widget`.
I would appreciate it if you could check this changes.

Regards,

## Labels

- [x] inaccuracy
- [ ] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/extend/best-practices/testing.html.md
+++ b/content/source/docs/extend/best-practices/testing.html.md
@@ -159,8 +159,11 @@ func testAccCheckExampleResourceExists(n string, widget *example.Widget) resourc
 
 		// If no error, assign the response Widget attribute to the widget pointer
 		*widget = *resp.Widget
+		if *widget == nil {
+			return fmt.Errorf("Widget (%s) not found", rs.Primary.ID)
+		}
 
-		return fmt.Errorf("Widget (%s) not found", rs.Primary.ID)
+		return nil
 	}
 }
 
--- a/content/source/docs/extend/best-practices/testing.html.md
+++ b/content/source/docs/extend/best-practices/testing.html.md
@@ -157,11 +157,12 @@ func testAccCheckExampleResourceExists(n string, widget *example.Widget) resourc
 			return err
 		}
 
-		// If no error, assign the response Widget attribute to the widget pointer
-		*widget = *resp.Widget
-		if *widget == nil {
+		if resp.Widget == nil {
 			return fmt.Errorf("Widget (%s) not found", rs.Primary.ID)
 		}
+
+		// assign the response Widget attribute to the widget pointer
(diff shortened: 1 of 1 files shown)


laurapacilio:
@apparentlymart I was wondering if you could either review, or maybe assign someone to check this change? Thank you!

content/source/docs/extend/best-practices/testing.html.md
```diff
@@ -159,8 +159,11 @@ func testAccCheckExampleResourceExists(n string, widget *example.Widget) resourc
 
 		// If no error, assign the response Widget attribute to the widget pointer
 		*widget = *resp.Widget
+		if *widget == nil {
```
paddycarver, line 162: `*widget` is dereferencing the pointer, so it can never be `nil`. This would actually panic on line 161 if `resp.Widget` were `nil`, as `*resp.Widget` is trying to dereference it. The `nil` check and error is probably better located before that, on line 159.

paddycarver (changes requested):
Thanks for working on this! That is indeed some strange code. Unfortunately, I think your fix isn't quite right, either. I left a comment inline about it.

content/source/docs/extend/best-practices/testing.html.md
budougumi0617: @paddycarver Thanks for the helpful comments. I have changed the PR, please check if the changes reflect your comments correctly.

paddycarver (approved):
Thanks for cleaning this up!

PR #1555, opened 2020-12-14
Updating unknown values changes section to incl. interpolated values

hcrhall:
This PR includes a minor change to the [Using Sentinel with Terraform 0.12](https://www.terraform.io/docs/cloud/sentinel/sentinel-tf-012.html) section to call out that when a Terraform value is computed, we will no longer provide an interpolated value. This was the default behavior prior to Terraform 0.12.

## Labels

- [ ] inaccuracy
- [x] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/cloud/sentinel/sentinel-tf-012.html.md
+++ b/content/source/docs/cloud/sentinel/sentinel-tf-012.html.md
@@ -172,7 +172,7 @@ following exceptions:
 
 Unknown values within `applied` in the
 [resource namespace](./import/tfplan.html#namespace-resources-data-sources)
-no longer return interpolated values, or the magic UUID (defined as
+no longer return values with interpolation sequences, or the magic UUID (defined as
 `74D93920-ED26-11E3-AC10-0800200C9A66` in Terraform 0.11 or earlier). Instead,
 unknown values are now returned as `undefined`.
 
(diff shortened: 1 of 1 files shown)


nfagerlund:
Just to make sure I understand what you mean by "interpolated values:" does this mean that both...

- `some_resource.foo.computed_attribute`
- `"String with ${some_resource.foo.computed_attribute}"`

... would appear as `undefined` in the applied map, as opposed to showing the first one as `undefined` and the second one as (something else)? 

Also, what would the (something else) have been, if we weren't blanking it out to `undefined`? `"String with undefined"`, or a string that still includes the literal un-substituted interpolation token, or...?

hcrhall:
> Just to make sure I understand what you mean by "interpolated values:" does this mean that both...
> 
> * `some_resource.foo.computed_attribute`
> * `"String with ${some_resource.foo.computed_attribute}"`
> 
> ... would appear as `undefined` in the applied map, as opposed to showing the first one as `undefined` and the second one as (something else)?
> 
@nfagerlund no, in the above example both would be known values, even though they are computed. What I mean is that previously, when using Terraform 11.x a `new` value in a diff would be represented as an interpolated value:
```
"diff": {
  "name": {
     "computed": true,
     "new": "${random_pet.master.id}",
     "old": "",
}
```
Whereas in later versions of Terraform (>= 0.12.x) the diff would appear as follows:
```
"diff": {
  "name": {
     "computed": true,
     "new": "",
     "old": "",
}
```

nfagerlund (approved):
Okay, thanks for explaining! I have a suggestion for how to tweak the wording; other than that, lgtm.

PR #1559, opened 2020-12-15
CircleCI: Reconfigure Slack notifications with orb

nfagerlund:
Circle's built-in Slack notifications (which are apparently deprecated anyhow)
only notify once at the end of the workflow, so we don't get the benefit from
splitting the deploy linkcheck jobs -- it's always just one cryptic red block in
our channel.

But the Slack orb is more configurable, and lets us give more informative
messages on a per-job basis! In fact, we can even do _multiple_  messages per
job, which will be nice for letting people know that their deploy is WIP.

## Considerations:

I'll need to use the Circle API to delete the legacy Slack configs, because they disabled the UI for that. I'm pretty sure I have the necessary permissions, I'll just go ahead and do it once this PR is approved. 

## Screenz:

The old notifications: (note that for this one, the deploy succeeded but the whole notification is red because of broken links.) 

Screen Shot 2020-12-14 at 4 43 24 PM

Testing the new notifications: (I tweaked some of the text after taking this screenshot, and added an `@here` mention to the deploy failure since it's a fairly dire circumstance that deserves broad notice.)

Screen Shot 2020-12-14 at 4 42 47 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,9 @@
 version: 2.1
 
+orbs:
+  # Slack orb v. 3 uses a webhook URL for auth. If upgrading to v. 4+, we'll need a token instead.
+  slack: circleci/slack@3.4.2
+
 references:
   images:
     middleman: &MIDDLEMAN_IMAGE docker.mirror.hashicorp.services/hashicorp/middleman-hashicorp:0.3.44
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -58,6 +70,10 @@ jobs:
           working_directory: content
           command: ./scripts/upload.sh
 
+      - slack/status:
+          success_message: ":terraformda: Successfully deployed to terraform.io."
+          failure_message: ":boom: Failed to deploy terraform.io! Urgent attention needed! <!here>"
```
nfagerlund, line 75: `<!here>` is API-ese for `@here`. https://api.slack.com/reference/surfaces/formatting#special-mentions

mdeggies (approved):
Emoji game is on point! Looks great to me.

nfagerlund:
And, legacy notifs successfully removed.

PR #1561, opened 2020-12-16
Circle configs: replace wget with linkcheck

nfagerlund:
(This commit also updates a couple links in the site header navs. Just a no-op to skip an unnecessary redirect.)

After deploys for terraform.io, we've been using `wget` in spider mode to
simultaneously warm up the Fastly cache and report on any broken links. It's
okay at the cache thing, but its reporting of broken links is abysmal -- most
notably, it doesn't tell you what pages the broken links CAME from, only the
failed destination URL.

[filiph/linkcheck](https://github.com/filiph/linkcheck) (Dart) gives much better
reporting, and still visits every page on the site to warm the cache.

We need to use a beta version of the Dart SDK for running it, because we need a
recent fix for recognizing 308 codes (sent by Vercel, currently just for the
marketing pages) as redirects. Details are in the yaml file.

We considered some other options:

- [linkchecker](https://github.com/linkchecker/linkchecker) (Python) gives
  similar quality of results, but is an order of magnitude slower.
- muffet (Go) (and our wrapper around it) can't be configured to stay within a
  single site, so we get noise from external hosts that don't want to be
  spidered. (Yeah, it doesn't recurse into external sites, but at the frequency
  we're running this task, I don't even really want to hit the top-level links.)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,6 +8,9 @@ references:
   images:
     middleman: &MIDDLEMAN_IMAGE docker.mirror.hashicorp.services/hashicorp/middleman-hashicorp:0.3.44
     ubuntu: &UBUNTU_IMAGE ubuntu-1604:201903-01
+    # We rely on dart-lang/sdk commit 7e7c01e804179782f884b174706ed0c80fb2ef71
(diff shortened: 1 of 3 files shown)


mdeggies (approved):
Looks great!

PR #1568, opened 2020-12-18
merge master to provider sunset branch

nfagerlund:
Just keeping updated.


nfagerlund:
OK! I'm going to self-merge this because it adds nothing new, and TFC confirmed that it has no effect on the relevant workspace due to no changes in the terraform dir.

PR #1569, opened 2020-12-18
Provider namespace redirects: github, ecl, and skytap

nfagerlund:
I think this is the right way to do these, but there was a comment down in the `hashicorp` namespace that gave me pause.


nfagerlund:
OK, looks like paul did the ECL and skytap ones on master instead, and I'm not sure yet what's up with github

PR #1575, opened 2020-12-22
Update mounted disk application configuration with `pg_password`

hashitop:
## Labels

- [x] inaccuracy
- [x] clarification
- [ ] new docs


justinretzolk:
LGTM based on [this](https://hashicorp.slack.com/archives/C7HG09LBV/p1605825421206200)

content/source/docs/enterprise/install/automating-the-installer.html.md
```diff
@@ -180,9 +180,10 @@ If you have chosen `production` for the `installation_type`, `production_type` i
 
 - `production_type` — One of `external` or `disk`.
 
-If you have chosen `disk` for `production_type`, `disk_path` is required:
+If you have chosen `disk` for `production_type`, `disk_path` and `pg_password` are required:
```
sudomateo, line 183: The `pg_password` setting is not currently required. One will be generated if not provided. However, I do agree that it should be documented and that we should recommend that the user explicitly set this value so that it is known for re-installations. I do not believe that we should call it required until it actually becomes, if ever, required in the settings.

content/source/docs/enterprise/install/automating-the-installer.html.md
```diff
@@ -180,9 +180,10 @@ If you have chosen `production` for the `installation_type`, `production_type` i
...
 
 - `disk_path` — Path on instance to persistent storage.
+- `pg_password` — The password for the internal PostgreSQL access.
```
sudomateo, line 186: I would add another sentence detailing that it is auto-generated if not provided and that the user should consider explicitly setting their own value or retrieving the auto-generated value prior to re-installation.

sudomateo (changes requested):

content/source/docs/enterprise/install/automating-the-installer.html.md
hashitop: Updated, thank you!

content/source/docs/enterprise/install/automating-the-installer.html.md
hashitop: Added, thank you!

sudomateo (approved):

content/source/docs/enterprise/install/automating-the-installer.html.md
```diff
@@ -183,6 +183,7 @@ If you have chosen `production` for the `installation_type`, `production_type` i
 If you have chosen `disk` for `production_type`, `disk_path` is required:
 
 - `disk_path` — Path on instance to persistent storage.
+- `pg_password` — The password for the internal PostgreSQL access, the password will be auto-generated for each installation if not provided.
```
sudomateo, line 186:
```suggestion
- `pg_password` — The password for the internal PostgreSQL access. The password will be auto-generated for each installation if not provided.
```

sudomateo (changes requested):
Minor grammar change.

sudomateo (approved):

PR #1580, opened 2021-01-07
Update PRM docs with submodules and examples dropdown

bethanyr:
## Labels

- [ ] inaccuracy
- [ ] clarification
- [ ] new docs


content/source/docs/cloud/registry/using.html.md
```diff
@@ -55,6 +55,16 @@ Click a module's "Details" button to view its details page. Use the "Versions" d
 
 ![Terraform Cloud screenshot: a module details page](./images/publish-module-details.png)
 
+### Viewing Nested Modules
```
nfagerlund, line 58: Let's put a blank line between the header and the paragraph. Markdown processors and editors handle things more consistently if you make sure there's space between block-level elements. 

```suggestion
### Viewing Nested Modules

```

content/source/docs/cloud/registry/using.html.md
```diff
@@ -55,6 +55,16 @@ Click a module's "Details" button to view its details page. Use the "Versions" d
...
+If a module contains nested modules following the [standard module structure](/docs/modules/structure.html), then a "Submodules" dropdown will appear below the module source information.
```
nfagerlund, line 59: I have some questions! 

- It looks like the convention is that a nested module can be used independently if it contains a `README`/`README.md` file. Is that how the registry decides which submodules to let you browse? Or does it include other submodules as well? 
- Do nested modules get their own independent source addresses in the registry? As in, if I paste the snippet from the "provision instructions" block for a submodule, will it just work without me needing to also include the parent module in my configuration? 
    - If so, what's the structure of a source address for a nested module?

content/source/docs/cloud/registry/using.html.md
```diff
@@ -55,6 +55,16 @@ Click a module's "Details" button to view its details page. Use the "Versions" d
 
 ![Terraform Cloud screenshot: a module details page](./images/publish-module-details.png)
 
+### Viewing Nested Modules
+If a module contains nested modules following the [standard module structure](/docs/modules/structure.html), then a "Submodules" dropdown will appear below the module source information. 
+
+![Terraform Cloud screenshot: a module submodules button](./images/publish-submodules-dropdown.png)
```
nfagerlund, line 61: I thought it was weird that these images would have `publish-` in the filename, since this is the "using modules" page rather than the "publishing modules" page, and then I noticed that the last image file referenced above in the existing content has `publish-`! 

_90% chance that was my fault_ — I probably wrote this page second, and just re-used an existing screenshot that was generic enough to use in either context. Sorry about that! 

Still, it's probably worth renaming these two files to start with `using-` instead, just so that _most_ of these images stay clearly associated with a particular page.

nfagerlund (commented):
Thanks for this! I have one adjustment to the markup, and then I have a couple of questions about the implementation, most of which are inline. Depending on the answers, I might suggest some changes to the text.

Also: Do you think it's worth tweaking some of the explanations on the module _publishing_ page, as well? I'm looking at this section here: 

https://www.terraform.io/docs/cloud/registry/publish.html#preparing-a-module-repository

...and a couple of things stand out: 

> **One module per repository.** The registry cannot use combined repositories with multiple modules.

It sounds like this is now something more like one _top-level_ module per repository, right? Nested modules work fine, but sibling modules are a no-no. 

> **Standard module structure.** The module must adhere to the standard module structure. This allows the registry to inspect your module and generate documentation, track resource usage, and more.

Depending on the answers to those implementation questions below, this would be a good place to explicitly point out that if you put a README in one of your nested modules, _we're gonna hold you to that_ and will allow users in your organization to consume it as an independent module!

content/source/docs/cloud/registry/using.html.md
bethanyr: I have added the blank line.

content/source/docs/cloud/registry/using.html.md
bethanyr: ah, good catch. I have updated the file names. Hopefully the publish page isn't confusing because I updated the publish-modules-details file to match the current prod UI. The images above it still say `beta` on the publish page.

content/source/docs/cloud/registry/using.html.md
bethanyr: there is no checking by the registry if you are using nested modules with or without a readme. They are just sorted differently in the dropdown:
image

Basically it works the same as the public registry does today (which I couldn't find any documentation for the public registry)

bethanyr: Also the source is different for submodules and examples:
```
module "project-factory_fabric-project" {
  source  = "tfe-zone-1a04f900.ngrok.io/example_corp/project-factory/google//modules/fabric-project"
  version = "9.2.0"
  # insert required variables here
}
```

bethanyr:
@nfagerlund - I have updated to include `examples` which we are in the process of exposing via the UI. Using the submodules and examples should work the same as they do in the public registry - which I couldn't find any documentation that I could leverage for the PMR.

I'm going to follow up on your other questions about updating the module _publishing page_: https://www.terraform.io/docs/cloud/registry/publish.html#preparing-a-module-repository

bethanyr:
@nfagerlund - after looking through existing modules documentation I found: https://learn.hashicorp.com/tutorials/terraform/pattern-module-creation - which talks more about nested modules - would it make sense to include a link to this doc?

We are planning to make more updates to the PMR this year - it might make sense to create a separate ticket to do more docs cleanup in the PMR.

nfagerlund (approved):

PR #1585, opened 2021-01-12
Policy checks API: add show endpoint, include run relationship

nfagerlund:
This recent API change is now live for TFC.


content/source/docs/cloud/api/policy-checks.html.md
```diff
@@ -86,19 +92,83 @@ curl \
 }
 ```
 
+## Show Policy Check
+
+This endpoint gets information about a specific policy check ID. Policy check IDs can appear in audit logs.
+
+-> **Note**: The `sentinel` hash in the `result` attribute structure represents low-level Sentinel details generated by the policy engine. The keys or structure may change over time. Use the data in this hash at your own risk.
```
radditude, line 99: This is a bit out of scope of this PR, but I'm curious about this line and its twin in the "List Policy Checks" section above!

It's my understanding that the stuff under the `sentinel` key here is roughly the same as the Sentinel CLI output when running [`sentinel apply -json`](https://docs.hashicorp.com/sentinel/writing/tracing#json-output) - if so, it would probably be handy to add a link to those Sentinel docs here for context (cc @chrisarcand to check me on that)

radditude (approved):

content/source/docs/cloud/api/policy-checks.html.md
nfagerlund: Yeah, good call! I wandered slightly out of scope anyway because I spotted some inconsistencies in the formatting through this page, so I could certainly tweak this too. I guess it's a question of how much we want to discourage / tacitly-encourage use of that info.

chrisarcand: This note is appropriate and doesn't merit a link, IMO.

It roughly is the JSON you link, but the point is plainly stated. These are details generated by the policy engine (Sentinel) and are exposed in the API as a TFC integration point, their contents are not subject to TFC's API stability policy and it's not the API's concern (as it's not the API who determines its contents).

nfagerlund:
This API change is deployed! Since CJ approved and Chris answered the outstanding question about the Sentinel results, I'm gonna merge.

PR #1592, opened 2021-01-14
Weekly incoming links check

nfagerlund:
I think this is the last bit of the link checking work, aside from the still in-flight PR in terraform core. 

The old wget spidering link check also checked a list of specific popular URLs that was, at some point, exported from google analytics; the goal was just to make sure we had redirects in place if we deleted any popular pages. 

This PR:

- Fixes a bug in the implementation. 
- Splits it out into its own thing.
- Makes a weekly CircleCI job to run that check and report it to our slack channel, since it would be a bad fit for the per-deploy workflow. 
- Deletes the rest of the infrastructure around the old wget link checking system.


alvin-huang (approved):
LGTM!

PR #1593, opened 2021-01-14
PR link check: add special casing for Vercel routes

nfagerlund:
This should help avoid false positives for the path prefixes that get served by
the Next/Vercel app in production. Unfortunately there's not a good way to
discover which paths that currently applies to, but at the time of writing it's
just /cloud and /community, and by the time they expand it much further, we'll
probably be looking to throw this script away anyhow.

This PR also includes a fix to ignore non-HTTP(S) URL protocols like mailto:.

### DIFF
--- a/content/scripts/check-pr-links.rb
+++ b/content/scripts/check-pr-links.rb
@@ -42,6 +42,11 @@ errors = {}
 
 # takes a `URI` object, returns [ok, html-or-error]
 def check_link(url)
+  # Ignore non-web protocols like mailto:
+  unless url.scheme == 'https' || url.scheme == 'http'
+    return [true, 'Not an HTTP(S) URL']
+  end
+
   # Special case for Vercel routes: can't check them against a local build, so
   # change URL to check prod.
   if url.path =~ VERCEL_REGEXP && url.to_s =~ /^#{ROOT_URL}/
--- a/content/scripts/check-pr-links.rb
+++ b/content/scripts/check-pr-links.rb
@@ -16,15 +16,22 @@ require 'erb'
 
 # Only checking files in website content.
 # Main content dir is "content/source/" for terraform-website, "website/" for terraform.
-site_root_paths = %r{^(content/source/|website/)}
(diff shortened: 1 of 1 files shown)


content/scripts/check-pr-links.rb
```diff
@@ -122,8 +142,7 @@ def find_links(html)
   puts '=== No broken links! ==='
 else
   puts "=== Found broken links! ==="
-  puts "Fix before merging... or if they're not really broken, explain why."
-  puts "(NOTE: This script reports false positives for Vercel app routes like /cloud. Sorry!)\n\n"
+  puts "Fix before merging... or if they're not really broken, explain why.\n\n"
```
acidprime, line 145:
```suggestion
  puts 'Fix before merging... or if they're not really broken, explain why.\n\n'
```
No interpolation so no need for double quotes

content/scripts/check-pr-links.rb
```diff
@@ -35,6 +42,19 @@
 
 # takes a `URI` object, returns [ok, html-or-error]
 def check_link(url)
+  # Ignore non-web protocols like mailto:
+  unless url.scheme == 'https' || url.scheme == 'http'
+    return [true, 'Not an HTTP(S) URL']
+  end
```
acidprime, line 48:
```suggestion
   return [true, 'Not an HTTP(S) URL'] unless url.scheme == 'https' || url.scheme == 'http'
```
This makes it more clear to me but just a thought

acidprime (approved):
Did not run ; but overall looks good, strictly speaking you should `.freeze` the constants when assigned to mutable objects but I doubt that is an issue here.

content/scripts/check-pr-links.rb
nfagerlund: Alas, the `\n` newlines require doubles.

content/scripts/check-pr-links.rb
nfagerlund: Works for me!

PR #1600, opened 2021-01-20
add docs for all billing resources

radditude:
We never actually documented most of the billing API endpoints for Terraform Cloud. Here they are!

Also updates the organization API responses and org entitlement breakdown to reflect the latest.

### DIFF
--- a/content/source/docs/cloud/api/feature-sets.html.md
+++ b/content/source/docs/cloud/api/feature-sets.html.md
@@ -91,6 +91,28 @@ curl \
       }
     },
     {
+      "id": "fs-4ec3b3RVJWsBkst3",
+      "type": "feature-sets",
+      "attributes": {
+        "comparison-description": "Gain additional features built for larger teams and enterprise environments.",
+        "cost": 0,
+        "description": "Full access to Terraform Cloud's enterprise features",
+        "identifier": "business",
+        "is-current": true,
+        "is-free-tier": true,
+        "name": "Business",
+        "plan": null,
+        "self-serve-billing": false,
+        "cost-estimation": true,
+        "sentinel": true,
+        "teams": true,
+        "user-limit": null,
+        "audit-logging": true,
+        "sso": true,
+        "private-networking": true
+      }
+    },
+    {
       "id": "fs-T9BCyZi3KJyWHebk",
       "type": "feature-sets",
       "attributes": {
@@ -140,7 +162,7 @@ curl \
 
 ## List Feature Sets for Organization
 
-This endpoint lists the feature sets available to a particular organization. The results may differ from the previous global endpoint - for instance, if the organization has already had a free trial, the trial feature set will not appear in this list.
+This endpoint lists the feature sets a particular organization is eligible to access. The results may differ from the previous global endpoint - for instance, if the organization has already had a free trial, the trial feature set will not appear in this list.
 
 `GET /organizations/:organization_name/feature-sets`
 
@@ -204,6 +226,28 @@ curl \
         "audit-logging": false,
         "sso": false,
         "private-networking": false
+      }
+    },
+    {
+      "id": "fs-4ec3b3RVJWsBkst3",
+      "type": "feature-sets",
+      "attributes": {
+        "comparison-description": "Gain additional features built for larger teams and enterprise environments.",
+        "cost": 0,
+        "description": "Full access to Terraform Cloud's enterprise features",
+        "identifier": "business",
+        "is-current": true,
+        "is-free-tier": true,
+        "name": "Business",
+        "plan": null,
+        "self-serve-billing": false,
+        "cost-estimation": true,
+        "sentinel": true,
+        "teams": true,
+        "user-limit": null,
+        "audit-logging": true,
+        "sso": true,
+        "private-networking": true
       }
     },
     {
--- a/content/source/docs/cloud/api/feature-sets.html.md
+++ b/content/source/docs/cloud/api/feature-sets.html.md
@@ -1,0 +1,233 @@
+---
+layout: "cloud"
+page_title: "Feature Sets - API Docs - Terraform Cloud and Terraform Enterprise"
+---
+
+[200]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200
+[201]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201
+[202]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202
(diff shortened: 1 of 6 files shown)


radditude:
The link checker is very unhappy about the `mailto` link in `content/source/docs/cloud/api/index.html.md`. I think it's probably ok to ignore in this case!

nfagerlund (commented):
I'm still seeing traffic on commits, so I'll wait a little while before doing a line-by-line, but one thing I noticed: the source filenames for the new pages should use the `.html.md` "double extension." If I remember right, that's actually not _required_ by middleman, but I feel like I ran into some weird rare edge cases when omitting it, and consistency is also nice for its own sake. 👍🏼

nfagerlund:
Oh HELLO, it looks like my little link checker was NOT expecting mailto: to exist. 😑 I'd better go make it exclude anything but http(s).

radditude:
Ah, thank you!

radditude:
@nfagerlund this is ready for eyes now when you get a chance to circle back!

content/source/docs/cloud/api/feature-sets.html.md
```diff
@@ -0,0 +1,233 @@
...
+        "private-networking": false
+      }
+    }
+  ]
+}
+```
+
+## List Feature Sets for Organization
+
+This endpoint lists the feature sets available to a particular organization. The results may differ from the previous global endpoint - for instance, if the organization has already had a free trial, the trial feature set will not appear in this list.
```
nfagerlund, line 143: Question: How's this interact with the Business tier? My expectation would be that it'd appear in the global list but NOT the org list (usually), because you can't just swipe a credit card to get it... but I'm not seeing it in the global list either.

content/source/docs/cloud/api/index.html.md
```diff
@@ -113,14 +123,14 @@ For example, if the API endpoint documentation defines the path `/runs` then the
 
 ## JSON API Formatting
 
-The Terraform Cloud endpoints use the [JSON API specification](http://jsonapi.org/), which specifies key aspects of the API. Most notably:
+The Terraform Cloud endpoints use the [JSON API specification](https://jsonapi.org/), which specifies key aspects of the API. Most notably:
```
nfagerlund, line 126: Ah, good catches, thank you. 🙏

content/source/docs/cloud/api/invoices.html.md
```diff
@@ -0,0 +1,113 @@
...
+
+# Invoices API
+
+-> **Note:** The invoices API is only available in Terraform Cloud.
+
+Organizations on credit-card-billed plans may view their previous and upcoming invoices.
+
+## List Invoices
+
+This endpoint lists the previous invoices for an organization.
```
nfagerlund, line 31: Is this endpoint paginated? (I'm guessing the invoices pile up, after a while... or at least we'd hope so 😅)

If so let's add the pagination boilerplate, should be able to just grab it from one of the other pages that uses it.

content/source/docs/cloud/api/index.html.md
```diff
@@ -63,35 +63,45 @@ The [show entitlement set](./organizations.html#show-the-entitlement-set) endpoi
 
 The following entitlements are available:
 
-- `state-storage` — Allows an organization to store state versions in its workspaces, which enables local Terraform runs with the remote backend. Affects the [state versions][] endpoints.
+- `agents` — Allows isolated, private or on-premises infrastructure to communicate with an organization in Terraform Cloud. Affects the [agent pools][], [agents][], and [agent tokens][] endpoints.
```
nfagerlund, line 66: aaa, thanks for updating this!

content/source/docs/cloud/api/subscriptions.html.md
```diff
@@ -0,0 +1,197 @@
...
+[500]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500
+[504]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504
+[JSON API document]: /docs/cloud/api/index.html#json-api-documents
+[JSON API error object]: http://jsonapi.org/format/#error-objects
+
+# Subscriptions API
+
+-> **Note:** The subscriptions API is only available in Terraform Cloud.
+
+An organization can subscribe to different [feature sets](./feature-sets.html), which represent the [pricing plans](/docs/cloud/paid.html) available in Terraform Cloud. An organization's [entitlement set](./index.html#feature-entitlements) is calculated using its subscription and feature set.
```
nfagerlund, line 27:
```suggestion
An organization can subscribe to different [feature sets](./feature-sets.html), which represent the [pricing plans](/docs/cloud/paid.html) available in Terraform Cloud. An organization's [entitlement set](./index.html#feature-entitlements) is calculated using its subscription and feature set.

To change the subscription for an organization, use the billing settings in the Terraform Cloud UI. Subscriptions cannot be changed via the API. 
```

Figured we should just say it before someone asks.

nfagerlund (approved):
I have a few questions, but this is in great shape, so I'm going to pre-approve — feel free to merge once you've determined: 

- Whether there's anything we need to say about Business tier
- Whether invoices needs the pagination bit

By the way, bonus thanks for updating the semi-related output in the orgs endpoints!

content/source/docs/cloud/api/index.html.md
radditude: I give the link checker full credit on these - jsonapi.org returns 301s for http urls now 😆

content/source/docs/cloud/api/subscriptions.html.md
radditude: Excellent addition! Technically there is an API endpoint for updating subscriptions, and I was going back and forth a bit on whether to include it here - it's _almost_ useless when you're not using the UI because of some Stripe magic on the frontend.

content/source/docs/cloud/api/feature-sets.html.md
radditude: This is a weird thing that I wasn't 100% sure how best to represent here! The Business tier should  appear in both, so I must have copy-pasted something wrong at some point, but as you note, organizations can't actually subscribe to it directly.

Maybe better phrasing would be:

```suggestion
This endpoint lists the feature sets a particular organization is eligible to access. The results may differ from the previous global endpoint - for instance, if the organization has already had a free trial, the trial feature set will not appear in this list.
```

content/source/docs/cloud/api/invoices.html.md
radditude: Good catch! This endpoint is indeed paginated, but it uses a weird bespoke pagination scheme for reasons that boil down to "Stripe". Added.

PR #1601, opened 2021-01-20
update network requirements for active/active

p0pr0ck5:


nfagerlund:
Hey! It looks like the broken link checker noticed two problems with external links on this page (see the circleci build details). 

- The GCP one just moved to a new location (https://cloud.google.com/network-connectivity/docs/vpn/concepts/mtu-considerations).
- The Docker one is just gone!!! I don't know enough about Docker networking to know where the new best source to point to is; possibly https://docs.docker.com/network/bridge/? 

I know this PR doesn't actually touch these links, but I'd be grateful if you could catch them anyway while you're in the page. 🌻

nfagerlund:
Also, I don't have the context to approve this PR, because I don't understand the active/active architecture thoroughly enough to know whether all that network traffic actually _goes away_ now. You'll need approval from someone else who knows the TFE guts better.

p0pr0ck5:
Thanks @nfagerlund! I've cleaned up those links :+1: with regard to the Docker link, the current way things are setup is a bit legacy; pointing to the generic Docker networking docs seems like the best approach to me.

@sudomateo this is ready for your :eyes: following our chat. Thanks!

sudomateo (approved):

PR #1604, opened 2021-01-21
Registry/provider sunset transition

nfagerlund:
This PR will merge all of the updated Fastly redirect rules to master. The `registry/provider-sunset-transition` branch will receive NO further commits, so it's safe to land!

## Context

The provider docs redirects are part of the site's Fastly configuration, and they're managed by a Terraform configuration contained right here in this repo. That configuration gets applied by a TFC workspace. 

Since TFC wasn't able to handle the ridiculous number of submodules we used to have (200+), we couldn't use it to run the configuration from the master branch. We wanted to share responsibility for runs without having to hand out Fastly credentials, so we switched the workspace to do remote operations, but pointed at a special branch (`registry/provider-sunset-transition`) where all of the provider submodules were prematurely removed. 

To redirect a new provider (at time of writing we're down to eight), we merge Terraform changes to that special branch. 

Once we're done, and all the provider submodules have been removed, we want to make sure the Terraform configuration gets updated in the master branch. We'll probably want to leave this special branch abandoned and just bring over the changes to fastly.tf.


laurapacilio (approved):
I trust you on this that this is correct :-) Thank you!

nfagerlund:
Oh gdi, github auto-deleted the branch. 😅 I'll fix.

nfagerlund:
That branch should _probably_ stick around for a while, just because we don't necessarily want that TFC workspace to plan a run for EVERY unrelated docs merge. But I don't know that we want it eternally... Hmm. Sounds like a question for another day.

PR #1605, opened 2021-01-22
Add supplementary redirects for implied index URLs

nfagerlund:
I'll just let the comments at the top of the redirects.txt file explain this for
me:

> The live terraform.io website handles index.html files in the normal fashion:
> - /docs/cloud/free/ will actually serve the content at /docs/cloud/free/index.html.
> - /docs/cloud/free (without the trailing slash) will go to /docs/cloud/free/
>   if there's a directory at "free" instead of a normal file.
>
> However, these redirect rules aren't that clever. This means that if you
> redirect an index.html file, you must also redirect its parent path, both with
> and without the trailing slash; otherwise, they'll continue to 404 instead of
> finding the new location.

Bites me every time, sigh.

### DIFF
--- a/content/redirects.txt
+++ b/content/redirects.txt
@@ -602,13 +602,21 @@
 
 # Removing some ghost pages
 /docs/backends/index.html           /docs/language/settings/backends/index.html
+/docs/backends/                     /docs/language/settings/backends/index.html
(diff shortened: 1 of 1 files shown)


judithpatudith (approved):
LGTM!

PR #1609, opened 2021-01-25
active/active support bundle addition

mikegreen:
## Labels

- [ ] inaccuracy
- [ x] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/enterprise/support/index.html.md
+++ b/content/source/docs/enterprise/support/index.html.md
@@ -27,6 +27,12 @@ Terraform Enterprise can automatically generate a support bundle including logs 
 
 Diagnostic information is available via in the installer dashboard on port 8800 of your installation.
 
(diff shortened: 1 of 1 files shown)


mikegreen:
build fail due to this - 
  - Broken link: https://support.hashicorp.com
    [301 Moved Permanently] https://support.hashicorp.com

which is acceptable

practicalint (commented):
Looks good as far as information communicated. The sentence is kinda long. May want to break it up something like multiple sentences and putting the command separate and maybe even in a code box.

practicalint (changes requested):
-> **Note**: For [Terraform Enterprise Active/Active](https://www.terraform.io/docs/enterprise/install/active-active.html) implementations the installer dashboard is disabled. Support bundles can instead be generated and saved to the configured object storage location by executing this command from the command line on a node instance:
 
```
tfe-admin support-bundle
```
This can and should be done individually for every node in the cluster.

practicalint (approved):
LGTM

sudomateo (commented):
Thank you for submitting this!

PR #1610, opened 2021-01-26
Cut old content in HCL1 from AWS RA

ml4:
As the ever-perspicacious @alexbasista found, the example linked in the AWS RA points to an example of deployment which is in HCL1 format which has not been updated in 3 years.  I vote that, like the other RAs which do not have the equivalent old content linked, this should be removed, especially as suitable replacements are being worked on.

- [x] inaccuracy
- [ ] clarification
- [ ] new docs


practicalint (approved):
Looks good!

PR #1622, opened 2021-02-09
Update index.html.md

Galser:
We had several customers stepping on S3 test problem in ReplicatedUI dashboard during config phase , perhaps it should be outlined  here, in our manual.

The source of the information is here : https://help.replicated.com/docs/config-screen/test-procs/#aws-auth

> Feel free to decline if you feel that this is more suited for KB artilce in Support

### DIFF
--- a/content/source/docs/enterprise/before-installing/index.html.md
+++ b/content/source/docs/enterprise/before-installing/index.html.md
@@ -163,6 +163,8 @@ At a minimum, Terraform Enterprise requires the following S3 permissions:
     ]
 }
 ```
+-> **Note:** During the interactive installation Replicated (TFE setup dashboard) can execute S3 access tests that will be perfomed using `ListAllMyBuckets` action. And, as a result - require additional permission added *temporary* to the policy above - **ListAllMyBuckets**. Later, in operations Terraform Enterprise not require this permission.  
+
 
 #### KMS Policy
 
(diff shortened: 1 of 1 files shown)


nervda (changes requested):
**Note:** During the interactive installation Replicated (TFE setup dashboard) can execute S3 access tests that will be performed using `ListAllMyBuckets` action. And, as a result - require additional permission added *temporarily* to the policy above - **ListAllMyBuckets**. Later, in operations Terraform Enterprise not require this permission.

content/source/docs/enterprise/before-installing/index.html.md
```diff
@@ -163,6 +163,8 @@ At a minimum, Terraform Enterprise requires the following S3 permissions:
     ]
 }
 ```
+-> **Note:** During the interactive installation Replicated (TFE setup dashboard) can execute S3 access tests that will be perfomed using `ListAllMyBuckets` action. And, as a result - require additional permission added *temporary* to the policy above - **ListAllMyBuckets**. Later, in operations Terraform Enterprise not require this permission.
```
sudomateo, line 166: We can be a bit more brief here.

```suggestion
-> **Note:** The `s3:ListAllMyBuckets` permission is necessary when testing authentication via the Replicated web console. However, the permission is not required for Terraform Enterprise to function and can be removed once the authentication is successfully tested.
```

sudomateo (changes requested):
Thank you for working on this! Submitted a suggestion.

sudomateo:
@Galser This should not have been merged without an approval. Just an FYI for next time.

PR #1623, opened 2021-02-10
Add docs about renaming organizations

nfagerlund:
This is now shipped, as of Feb 10.

Here's the deal with renaming organizations: Don't. Except, if you've got a brand new test organization with a semi-random name and you want to give it a more sensible permanent name, actually go ahead because that's fine. 

image


judithpatudith (approved):
LGTM!

PR #1634, opened 2021-02-23
docs: add dlv flags in debugging

remyleone:
add continue and accept-multiclient to dlv command

The specified command was not working for me

## Labels

- [x] inaccuracy
- [x] clarification
- [x] new docs

### DIFF
--- a/content/source/docs/extend/debugging.html.md
+++ b/content/source/docs/extend/debugging.html.md
@@ -147,18 +147,19 @@ run, specifying whatever flags, environment variables, or other input is
 necessary to start your provider in debug mode:
 
 ```sh
-$ dlv exec --headless ./terraform-provider-my-provider -- --debug
+$ dlv exec --headless --accept-multiclient --continue ./terraform-provider-my-provider -- --debug
 ```
 
-Connect your debugger (whether it's your IDE or the debugger client) to the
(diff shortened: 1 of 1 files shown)


paddycarver:
Can you expand on what "not working for me" means?

remyleone:
I was not able to connect from my IDE (goland) to the binary running in debug mode.

paddycarver:
Did it give you any errors?

laurapacilio:
Hi all - just poking this to see if we can get it resolved :-) @remyleone did you have any additional comments here?

remyleone:
When I was trying the command without the options I added, the debugging process just crashed. I've looked around on the internet and adding the following options made it work on my setup.

paddycarver:
I'd love to see crash output if this is reproducible.

laurapacilio:
@remyleone Thank you for submitting this! Do you have crash output? I think we just want to make _sure_ that this is something that should be documented for everyone in the official docs. I also think it'd be great for us to understand what happened so that we can document further and/or make fixes. 

Let us know, thank you!

laurapacilio:
Hi @remyleone We appreciate your submission! I think in the absence of crash output to clarify, I suggest that we close this PR and perhaps you could submit an issue instead that we could dig into on the backend? Let us know what you think about this, thank you!

PR #1635, opened 2021-02-24
Add uuid and secret field to the POST definition

samm-git:
## Labels

- [x] inaccuracy
- [x] clarification
- [ ] new docs


chrisarcand:
Hello again! I've explained why `secret` and `uuid` are attributes used primarily for the Terraform Cloud user interface and how general API clients should instead provide a pre-generated OAuth token string in [the associated client PR](https://github.com/hashicorp/go-tfe/pull/182).

That's mentioned here in the docs:

image

As far as `display-name`, that's not a valid creation parameter - the existing `name` attribute is correct. You may be thinking of `service-provider` and `service-provider-display-name` in the response object.

Thanks!

samm-git:
@chrisarcand thank you for reply. Not sure if i understood you correctly. My usecase is simple:

1. Admin user creating Oath APP in the target github org and configuring Oauth VCS connection (but not establishing it). He gets secret and id from github.com.
2. User goes to target org and establishing connection using UI. He dont know oauth secret.

How i can make it without using fields i mentioned?

samm-git:
@chrisarcand my intention is that user will will generate ouath token, so i think TFE must know secret and oauth id to do this,

samm-git:
got it working, please ignore prev. message :)

PR #1636, opened 2021-02-24
Plugin installation is never skipped, remove from extend docs

pselle:
Plugin installation is never skipped as of 0.13+, and so this line is misleading.


bflad (approved):
Nice fix 🚀

judithpatudith (approved):

PR #1643, opened 2021-03-06
Update some links in TFC docs, adjust PR link checker

nfagerlund:
Did a tiny docs PR, and the checker caught a broken link. But there were a bunch of other links throughout the docs that matched that pattern, so I tried fixing all instances of them... then it kept snowballing and catching more outbound links with issues. :| 

(For the record, I do NOT endorse doing that if you're doing a normal docs PR! Just catch whatever turns up in the document you're currently messing with. But since I don't always follow my own advice, here's a PR.) 

The CI will be red for this job anyway, unfortunately. Here's the "explain why"s: 

- The registry docs and the GCP docs don't render right until a bunch of javascript rehydrates, so my dumb little link checker can't find anchors in them. But the links are good!
- The API docs template isn't a real page. The link checker thinks it was probably intended to be and thus something's probably awfully wrong. But really it's fine, and it's intentional that it fails to render.

### DIFF
--- a/content/scripts/check-pr-links.rb
+++ b/content/scripts/check-pr-links.rb
@@ -28,6 +28,13 @@ ROOT_URL = 'http://localhost:4567/'
 VERCEL_ROUTES = ['/community', '/cloud']
 VERCEL_REGEXP = /^(#{VERCEL_ROUTES.join('|')})/
 
+IGNORE_LIST = [
+  # The front page of TFC redirects to /session, but we *don't* want to update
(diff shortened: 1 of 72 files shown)


nfagerlund:
@judithpatudith I think this is mostly the same thing as #1650, so I'm gonna close it... but we should maybe talk more about this redirects thing, because I'm starting to think it's not ideal to report them as errors. (Redirects in our own educational properties are one thing, but out in the world? Maybe not so much.)

PR #1650, opened 2021-03-10
Correct jsonapi links to use https

judithpatudith:
Link checks on PRs to some Terraform docs are failing due to redirects. I was trying to find and replace the URLs that needed fixing but fell down a bit of a rabbit hole. I'd still like to get this PR merged to fix the links that I did get, but I'll follow up with the webdev team for a better way to address these later on. 

I was hoping to get _this_ PR to go green but the two errors that are still showing up are actually ok to keep. One is to Terraform Cloud, and shouldn't specify the landing page. The other isn't registering an anchor link right, but the link itself seems to be fine 🤔 

## Labels

- [ ] inaccuracy
- [ ] clarification
- [ ] new docs


im2nguyen (approved):

robin-norwood (approved):
Spot checked most of the links. They work. :)

PR #1660, opened 2021-03-19
Add note for cloning tfe vm

hashitop:
## Labels

- [X] inaccuracy
- [X] clarification
- [ ] new docs


content/source/docs/enterprise/install/installer.html.md
```diff
@@ -12,6 +12,8 @@ installer to set up Terraform Enterprise on a customer-controlled machine.
 
 -> **Important:** Before you begin, consult the [Pre-Install Checklist](../before-installing/index.html) for prerequisites. You'll need to prepare data files and a Linux instance.
 
+-> **Important:** Cloning an already installed Terraform Enterprise host machine can be done with extra cares. It is not allowed to have more than one Terraform Enterprise instances connecting to the same database unless the installation is performed for the [Active/Active architecture](active-active.html). When cloning Terraform Enterprise host machine, these following files contain the private IP address and the hostname of the previous installed environment, `/etc/sysconfig/replicated` or `/etc/default/replicated` and `/etc/sysconfig/replicated-operator` or `/etc/default/replicated-operator`. It is required to ensure that those configuration files maintain the correct private IP address and the hostname of the Terraform Enterprise host machine at the restoration from the image. The `replicated` and `replicated-operator` services should also be restarted when the configuration files are updated.
```
judithpatudith, line 15: Thanks for this addition! I had a couple of clarifying questions. 

This seems like too much information to put into a callout box, since it might be hard for the reader to visually parse. Does the pre-installation checklist previously reference not exist any more, or is there a reason these items shouldn't be a bulleted list in the regular body of the doc? 

I took a guess at some possible edits, but please check them thoroughly; I'm not sure I preserved the meaning. I'm specifically hesitant about a couple of things: 

> at the restoration from the image

Could replace at in the above sentence with "before", "during", or "after"? (I guessed "before").

> services should also be restarted

Does the operator need to manually restart these services, or do they just need to verify that the services have restarted automatically? (I guessed that the operator needed to restart them.)

> when the configuration files are updated

Does the operator update these files or are they automatically updated somehow? (I guessed that the operator updates them.)

```suggestion
-> **Important:** Be careful when cloning an already installed Terraform Enterprise host machine. You cannot have more than one Terraform Enterprise instance connect to the same database unless the installation is uses the [Active/Active architecture](active-active.html). When cloning a Terraform Enterprise host machine, you must make sure that the `/etc/sysconfig/replicated` or `/etc/default/replicated`, and `/etc/sysconfig/replicated-operator` or `/etc/default/replicated-operator` maintain the correct private IP address and the hostname of the Terraform Enterprise host machine before the restoration from the image. You should also restart the `replicated` and `replicated-operator` services when you update the configuration files.
```

hashitop: @judithpatudith as far as I know we currently don't have a `How To` document for cloning an existing VM with TFE pre-installed.

All the existing documents we have for TFE installation are to install TFE from scratch, there are customers already using the clone method and we are aware that Replicated also supports it. I am not certain that this installation method requires a dedicated section or it should be only an additional note on any existing document so at this stage I am just trying to add this note at the top of the installation page as surprisingly some customers have been doing what we never documented and failed then raised support tickets with us. 

Would it make more sense to relocate the note to this file below? I think the pre-install checklist is still something for fresh install rather than cloning VM.

https://github.com/hashicorp/terraform-website/blob/31655560ccc856ee3fc4e2c05b439edf522b1827/content/source/docs/enterprise/install/index.html.md

laurapacilio:
Hi @hashitop Wanted to check in on this - is this still something we'd like to try to get merged in somewhere in our docs, or should we close this out?  Thank you!

content/source/docs/enterprise/install/installer.html.md
hashitop: I don't see this is related to Active/Active architecture. The use case could be users clone dev environment into staging or from staging into production

hashitop (commented):
Committed the suggestion, I think it looks good

content/source/docs/enterprise/install/installer.html.md
```diff
@@ -12,6 +12,8 @@ installer to set up Terraform Enterprise on a customer-controlled machine.
 
 -> **Important:** Before you begin, consult the [Pre-Install Checklist](../before-installing/index.html) for prerequisites. You'll need to prepare data files and a Linux instance.
 
+-> **Important:** Be careful when cloning an already installed Terraform Enterprise host machine. You cannot have more than one Terraform Enterprise instance connect to the same database unless the installation is uses the [Active/Active architecture](active-active.html). When cloning a Terraform Enterprise host machine, you must make sure that the `/etc/sysconfig/replicated` or `/etc/default/replicated`, and `/etc/sysconfig/replicated-operator` or `/etc/default/replicated-operator` maintain the correct private IP address and the hostname of the Terraform Enterprise host machine before the restoration from the image. You should also restart the `replicated` and `replicated-operator` services when you update the configuration files.
+
```
laurapacilio, line 16: @hashitop I have another thought here. I agree with Judith that this is a lot to put in a callout box at the top of the page. Instead, I wonder if we can just make it a heading, like the others. I think that at least this way, users will be more likely to see it and the bullets will make it easier to parse. Thoughts? 

```suggestion
## Cloning a Host Machine

You cannot have more than one Terraform Enterprise instance connect to the same database unless the installation is uses the [Active/Active architecture](active-active.html). When cloning an existing Terraform Enterprise host machine:

-  Make sure that the `/etc/sysconfig/replicated` or `/etc/default/replicated`, and `/etc/sysconfig/replicated-operator` or `/etc/default/replicated-operator` maintain the correct private IP address and the hostname of the Terraform Enterprise host machine before the restoration from the image. 
- Restart the `replicated` and `replicated-operator` services when you update the configuration files.
```

laurapacilio (changes requested):
Top - I had another thought about how to handle this - see what you think! Also, I think that we should delete "Delivery" at the top of this page and start with "This page explains how to use the Terraform Enterprise installer to set up Terraform Enterprise on a customer-controlled machine."

content/source/docs/enterprise/install/installer.html.md
hashitop: :thinking:  Having them that way, it becomes more like a suggestion/option which I don't believe that it is our recommendation or suggestion to clone TFE host machine, it is kind of something that can be done but with some extra cautions. It might be best to check with other teams on this as well but the content looks ok to me.

laurapacilio: @sudomateo Do you have thoughts on this/do you know anyone who would?

sudomateo: From a technical perspective, cloning the instance can be a dangerous operation and only really acceptable in the following scenarios:
- When replacing an errant instance with a new instance that is as close of a replica as possible to the original instance. In this case both instances cannot be started at the same time due to database access issues.
- When creating an image template that can be used to recover the instance upon failure via automating scaling features. This type of cloning requires that the Terraform Enterprise application is not already installed among other cleanup items such as MAC address changes, etc.

With that being said, @hashitop what is the motivation or goal for this change? What information do we want the reader to be aware? Also, was there an issue or something that this information could have prevented?

hashitop: My initial thought would be just to provide a note for extra caution for whoever wants to implement this.
I don't think editing those files is a common thing that we would normally do. If we would not recommend this method at all then perhaps we would rather explicitly state that we don't support it. 

And as you mentioned that it is a dangerous operation, we could provide a separate set of instructions for backup and restore data into another machine or a different method for people to clone TFE instance? I do see that we have a page for backup/restore API but we also have other methods in TFE as well and for mounted disk, perhaps the CLI maybe faster and more reliable? When thinking about this, I am actually not sure why some people did this, perhaps we did not provide sufficient details on how to use the database backup/restore command on mounted disk mode? Maybe we need to rethink again but the initial thought was only to provide a set of cautions and a way to avoid the error providing that we did not explicitly state anything to this method currently.

laurapacilio: Thank you for the context, both!  If some non-trivial number of people are doing this/will want to do it, we should address it and then give our recommendations. Maybe we should add what @sudomateo wrote at the top of this new section along with an explicit warning that we don't recommend this? 

But I'm open to whatever y'all think is best here, as you have much more experience with the customers than I do. thoughts?

laurapacilio:
@hashitop per our conversations offline, do you want to just close this PR and return to this issue in a larger rework of the enterprise docs?

hashitop:
@laurapacilio agreed, let's close off the PR, the topic is more related to backup/restore IMO

PR #1665, opened 2021-03-23
Improve docs for automatic speculative plans, document new setting in UI

nfagerlund:
This setting already existed in the API, but we're adding it to the UI too. Then, because of the added complexity around explaining the behavior, I took the chance to rearrange that whole subheading. 

This was previously approved in private review by @radditude. ~I'm in the process of shipping that new feature this afternoon.~ This is now shipped to prod.


nfagerlund:
Approved on private and shipped the change, so I'm gonna merge.

PR #1666, opened 2021-03-24
Remove documentation about workspace performance attributes

jondavidjohn:
These were recently added to the workspace resource in the API and have since created some performance problems we'd like to address before committing to documenting them for long-term support.

We will be re-introducing these as part of the public API in the future, but would like to de-document them for the time being.


nfagerlund (approved):

jondavidjohn:
We were actually able to correct the performance issues in place 🎉 

No need for this now.

PR #1670, opened 2021-03-30
remove single module comments

rberlind:
PMR modules are no longer restricted to a single module.

## Labels

- [x ] inaccuracy
- [ ] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/cloud/registry/publish.html.md
+++ b/content/source/docs/cloud/registry/publish.html.md
@@ -23,7 +23,7 @@ Only members of the "owners" team can publish new modules. Once a module is publ
 
 The private module registry is designed to be as automatic as possible, so it defers to your VCS provider for most management tasks. The only manual tasks are adding a new module and deleting versions.
 
-After configuring at least one [connection to a VCS provider][vcs], you can publish a new module by specifying a properly formatted VCS repository (one module per repo, with an expected name and tag format; see below for details). The registry automatically detects the rest of the information it needs, including the module's name and its available versions.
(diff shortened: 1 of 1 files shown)


rberlind:
Compare the doc I modified to https://www.terraform.io/docs/registry/modules/publish.html#requirements.  But note that while the latter does require a registry description, I just tested that the TFC/E PMR will accept a registry that does not have one when publishing a module to the PMR and that it even creates a description for the module like "Terraform module for s3-bucket on provider aws." based on the name of the module.

The Terraform docs team might want to check whether registry descriptions are really needed in the public registry.  If not, that requirement could be dropped from https://www.terraform.io/docs/registry/modules/publish.html#requirements. But I did not have time to do a test in the public registry.

judithpatudith:
@bethanyr Do you know who can verify this for us? 

@rberlind sorry this sat for so long,  taking a look now. Looks like a link check is failing; do you mind if I rebase if necessary?

bethanyr:
the module description field doesn't appear to be required in the public registry - but I think we want module publishers to include it to provide more information about the module.

rberlind:
@bethanyr : you can handle it anyway you see fit.  Thanks.

judithpatudith (approved):

rberlind:
I'd prefer you do it, @judithpatudith . Thanks.

laurapacilio:
Merging this - there was one failing link check but a) it's not in anything that was changed here and b) it doesn't appear to actually be broken. So I'm merging anyway.

PR #1671, opened 2021-04-01
Update Active Active configuration values

bnferguson:
New in TFE v202104-1 will be the greatly simplified token requirements for Active/Active. This is a WIP PR to begin updating that documentation to show that enc_password is the only required token now. 

(Also I seem to not have repo access so this is on a fork for now!)

## Labels

- [ ] inaccuracy
- [x] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/enterprise/install/active-active.html.md
+++ b/content/source/docs/enterprise/install/active-active.html.md
@@ -268,8 +268,7 @@ For example:
 
 ```
 
-
--> **Note**: This value must match across nodes. In some previous versions there may have been other values such as `cookie_hash` and `archivist_token`, these are no longer needed but will still work if they are still set by your installation configuration. 
+-> **Note**: In versions prior to `v202104-1` the following values were also required to be set: `install_id`, `root_secret`, `user_token`, `cookie_hash`, `archivist_token`, `internal_api_token`, `registry_session_encryption_key` (HEX), and `registry_session_encryption_key` (HEX). These varlues are no longer required but will still work if they are still set by your configuration.  
 
 
 ### Step 3: Connect to External Redis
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/install/active-active.html.md
```diff
@@ -230,9 +230,9 @@ For example:
 
 
 
-##### Add Common Configuration
+##### Add Common Encryption Password
```
aaron-lane, line 233: Since the encryption password is now the only token covered in this section, I think removing "Common" will help clarify the topic.

```suggestion
##### Add Encryption Password
```

content/source/docs/enterprise/install/active-active.html.md
```diff
@@ -265,139 +255,21 @@ For example:
...
-resource "random_id" "registry_session_secret_key" {
-  byte_length = 16
-}
-
-locals {
-  registry_session_secret_key = random_id.registry_session_secret_key.hex
-}
-```
-
+-> **Note**: This value must match across nodes. In some previous versions there may have been other values such as `cookie_hash` and `archivist_token`, these are no longer needed but will still work if they are still set by your installation configuration.
```
aaron-lane, line 272: I think it will help with clarity if we provide the full list of deprecated values.

aaron-lane (changes requested):
Thanks for taking this on! A couple of thoughts follow.

aaron-lane (approved):
LGTM 🌬️

bnferguson:
`v20210401-1` has been released so this needs to be merged and published! 🎉

amy-hashi (approved):

PR #1674, opened 2021-04-07
Reference Arch: AWS: Be more explicit about regional outages

USA-RedDragon:
The wording of "Multi-Region Deployment to Address Region Failure" seems to imply the need to replicate the TFE EC2 instance as the warm standby ("An identical infrastructure should be provisioned in a secondary AWS
Region."), but it's the data plane that requires cross-region replication.

Recently had a support ticket where this wording was a source of confusion.

## Labels

- [ ] inaccuracy
- [x] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/enterprise/before-installing/reference-architecture/aws.md
+++ b/content/source/docs/enterprise/before-installing/reference-architecture/aws.md
@@ -340,9 +340,9 @@ failure on a regional AWS service. In this section, implementation patterns to s
 We recommend provisioning an identical infrastructure in a secondary AWS
 Region. Depending on recovery time objectives and tolerances for
 additional cost to support AWS Region failure, the infrastructure can be
-running (Warm Standby) or stopped (Cold Standby). Please note that with _Standalone_ implementation mode, only one Terraform Enterprise instance can be running against the same database. This deployment acts to minimize the Mean Time To Recovery (MTTR) in the event of a regional failure, avoiding the need to replicate and stand up the data plane infrastructure during an outage. In the event of the primary AWS Region hosting the Terraform Enterprise application fail, the secondary
-AWS Region will require some configuration before traffic is directed to
-it along with some global services such as DNS.
+running (Warm Standby) or stopped (Cold Standby). Please note that with _Standalone_ implementation mode, only one Terraform Enterprise instance can be running against the same database. 
+
+This deployment acts to minimize the Mean Time To Recovery (MTTR) in the event of a regional failure, avoiding the need to replicate and stand up the data plane infrastructure during an outage. If the primary AWS Region hosting the Terraform Enterprise application fails, you will need to perform some configuration before traffic is directed to the secondary AWS Region:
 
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/before-installing/reference-architecture/aws.md
```diff
@@ -345,6 +345,8 @@ primary AWS Region hosting the Terraform Enterprise application failing, the sec
 AWS Region will require some configuration before traffic is directed to
 it along with some global services such as DNS.
 
+Please note that with _Standalone_ implementation mode, only one Terraform Enterprise instance can be running against the same database. This deployment acts to minimize the Mean Time To Recovery (MTTR) in the event of a regional failure, avoiding the need to replicate and stand up the data plane infrastructure during an outage.
```
laurapacilio, line 348: I think it might be helpful to move this up so that it comes right after the warm and cold standby sentence. Right now, it interrupts the flow because it comes in between the sentence about configuration and the following examples.

laurapacilio (changes requested):

content/source/docs/enterprise/before-installing/reference-architecture/aws.md
```diff
@@ -340,8 +340,7 @@ failure on a regional AWS service. In this section, implementation patterns to s
 An identical infrastructure should be provisioned in a secondary AWS
 Region. Depending on recovery time objectives and tolerances for
```
laurapacilio, line 341:
```suggestion
We recommend provisioning an identical infrastructure in a secondary AWS
Region. Depending on recovery time objectives and tolerances for
```

content/source/docs/enterprise/before-installing/reference-architecture/aws.md
```diff
@@ -340,8 +340,7 @@ failure on a regional AWS service. In this section, implementation patterns to s
...
 additional cost to support AWS Region failure, the infrastructure can be
-running (Warm Standby) or stopped (Cold Standby). In the event of the
-primary AWS Region hosting the Terraform Enterprise application failing, the secondary
+running (Warm Standby) or stopped (Cold Standby). Please note that with _Standalone_ implementation mode, only one Terraform Enterprise instance can be running against the same database. This deployment acts to minimize the Mean Time To Recovery (MTTR) in the event of a regional failure, avoiding the need to replicate and stand up the data plane infrastructure during an outage. In the event of the primary AWS Region hosting the Terraform Enterprise application fail, the secondary
 AWS Region will require some configuration before traffic is directed to
 it along with some global services such as DNS.
```
laurapacilio, line 345:
```suggestion
running (Warm Standby) or stopped (Cold Standby). Please note that with _Standalone_ implementation mode, only one Terraform Enterprise instance can be running against the same database. 

This deployment acts to minimize the Mean Time To Recovery (MTTR) in the event of a regional failure, avoiding the need to replicate and stand up the data plane infrastructure during an outage. If the primary AWS Region hosting the Terraform Enterprise application fails, you will need to perform some configuration before traffic is directed to the secondary AWS Region:
```

laurapacilio (approved):
I don't want to block you, so I'm approving, but I suggested a couple more tweaks for flow. I changed a few things to active voice (easier to read) and also added some space between that note and the explanation. I think this makes the content easier to digest. Thank you for considering my suggestions :-)

USA-RedDragon:
Thanks a ton! Words are hard haha :100:

PR #1676, opened 2021-04-09
Update users.html.md - remove suggestion to use private API

kikitux:
This PR is to remove the the hint to use an internal API for admin.


radditude (approved):
Makes sense to me! Thanks for taking care of this.

PR #1679, opened 2021-04-13
Update agent documentation to clarify shutdown

ryanuber:
Adding a section about proper `tfc-agent` shutdown procedures, and fixing the resilience section to ensure it is not misleading.

### DIFF
--- a/content/source/docs/cloud/agents/index.html.md
+++ b/content/source/docs/cloud/agents/index.html.md
@@ -155,6 +155,17 @@ docker run -e TFC_AGENT_TOKEN=your-token -e
 TFC_AGENT_NAME=your-agent-name hashicorp/tfc-agent
 ```
 
+### Stopping the Agent
+
+The agent maintains a registration and a liveness indicator within Terraform Cloud during the entire course of its runtime. When an agent is to be retired, it must deregister itself from Terraform Cloud. The agent performs deregistration automatically as part of its shutdown procedure in the following scenarios:
+
+* If using an interactive terminal, Ctrl-C is pressed.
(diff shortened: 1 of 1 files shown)


content/source/docs/cloud/agents/index.html.md
```diff
@@ -195,7 +206,7 @@ You may choose to run multiple agents within your network, up to the organizatio
 
 #### Resilience
 
-If an agent terminates abnormally, any running jobs can be restarted where they left off by resubmitting them. We strongly recommend pairing the agent with a process supervisor to ensure that it is automatically restarted in case of an error.
+It is possible that an agent process could be terminated unexpectedly (due to killing the process forcefully, power cycling the host machine, etc.). We strongly recommend pairing the agent with a process supervisor to ensure that it is automatically restarted in case of an error.
```
vaficionado, line 209:
```suggestion
It is possible that an agent process could be terminated unexpectedly (due to killing the process forcefully, power cycling the host machine, etc.). We strongly recommend pairing the agent with a process supervisor to ensure that it is automatically restarted in case of an error. If this happens, any running jobs can be restarted where they left off by resubmitting them.
```

This seemed like an important bit of info that was no longer included, otherwise LGTM!

vaficionado (changes requested):
If you left off the sentence I added back for a reason, just LMK

content/source/docs/cloud/agents/index.html.md
ryanuber: That is the part that isn't true. You can't resume work that was in-flight when an agent is terminated unexpectedely. Runs will need to be canceled and enqueued again, which is a new run, not a restart. This is actually the specific bit that felt misleading to me. Does that make sense?

vaficionado: Aha, yes! That's info I was missing from the resolution yesterday. I should've known there was a reason you left it out. Makes sense to me, everything else looks good.

vaficionado (approved):

PR #1682, opened 2021-04-14
Add section for includes to Run Triggers API

jondavidjohn:
Per the upcoming Run Triggers addition to the workspace overviews, we've added a few includible resources.

Also noting the permission implications of the include behavior.

2021-04-14 at 1 21 PM

### DIFF
--- a/content/source/docs/cloud/api/run-triggers.html.md
+++ b/content/source/docs/cloud/api/run-triggers.html.md
@@ -319,4 +319,13 @@ curl \
   --header "Authorization: Bearer $TOKEN" \
   --header "Content-Type: application/vnd.api+json" \
   https://app.terraform.io/api/v2/run-triggers/rt-3yVQZvHzf5j3WRJ1
-```
+  ```
+
+## Available Related Resources
+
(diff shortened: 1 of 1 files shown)


content/source/docs/cloud/api/run-triggers.html.md
```diff
@@ -319,4 +319,13 @@ curl \
   --header "Authorization: Bearer $TOKEN" \
   --header "Content-Type: application/vnd.api+json" \
   https://app.terraform.io/api/v2/run-triggers/rt-3yVQZvHzf5j3WRJ1
-```
+  ```
```
nfagerlund, line 322:
````suggestion
```
````

I don't think you want to indent this?

nfagerlund (approved):
Rad, thank you. Feel free to merge at your convenience! 

Tangentially, now that I've been using includes sometimes for my own work, I'm really dissatisfied with the way I originally organized include info in these docs. Maybe someday I'll take a stab at reorganizing that. :| But that's not relevant here; for this PR, matching the rest of the API docs is the move. 👍🏼

PR #1683, opened 2021-04-16
update path for provider binaries

mildwonkey:
The directory layout shown in this doc is outdated, so removing the inaccurate `<OS>_<ARCH>` part and adding a note that providers need to be in the correct subdirectory under that path (the full installation info is already linked right above this, so i didn't go into further detail).

### DIFF
--- a/content/source/docs/extend/how-terraform-works.html.md
+++ b/content/source/docs/extend/how-terraform-works.html.md
@@ -101,26 +101,27 @@ configuration's [version constraints](/docs/language/providers/configuration.htm
 and chooses a version for each plugin as follows:
 
 - If any acceptable versions are installed, Terraform uses the newest
-  _installed_ version that meets the constraint (even if releases.hashicorp.com
+  _installed_ version that meets the constraint (even if the
+[Terraform Registry](https://registry.terraform.io/)
   has a newer acceptable version).
 - If no acceptable versions are installed and the plugin is one of the
-  [providers distributed by HashiCorp](/docs/providers/index.html),
-  Terraform downloads the newest acceptable version from
-  releases.hashicorp.com and saves it in `.terraform/plugins/<OS>_<ARCH>`.
-- If no acceptable versions are installed and the plugin is not distributed
-  by HashiCorp, initialization fails and the user must manually install an
-  appropriate version.
+  [providers distributed by HashiCorp](/docs/providers/index.html), Terraform
+  downloads the newest acceptable version from the [Terraform Registry](https://registry.terraform.io/) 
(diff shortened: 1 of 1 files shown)


content/source/docs/extend/how-terraform-works.html.md
```diff
@@ -118,9 +118,9 @@ releases.hashicorp.com for newer acceptable provider versions and downloads them
 if available.
 
 This behavior only applies to providers whose _only_ acceptable versions are in
-`.terraform/plugins/<OS>_<ARCH>` (the automatic downloads directory); if any
-acceptable version of a given provider is installed elsewhere,
-`terraform init -upgrade` will not download a newer version of it.
+the correct subdirectories under `.terraform/plugins/` (the automatic downloads
```
alisdair, line 121: Isn't this `.terraform/providers` now? Am I misunderstanding the context of these changes?

alisdair (commented):
There's another mention of `<OS>_<ARCH>` in the previous section. Should that also be removed?

content/source/docs/extend/how-terraform-works.html.md
mildwonkey: Nope - the new _install_ dir is `.terraform/providers`; we left `.terraform/plugins` alone since it was (widely) documented as the directory to drop provider binaries in: 
https://github.com/hashicorp/terraform/blob/23800438ab40890cac41440aa4fd6c798265fb71/command/meta_providers.go#L110-L113

I'll go check that other ref to OS_ARCH, thank you!

alisdair: Ah! Maybe "the automatic downloads directory" is now misleading, then?

mildwonkey: OH HRM that's a very good point - the directory referred to in this section IS `.providers`. 

arg sorry 🙃 I haven't looked closely at this in too long (and did not read this page closely enough)

mildwonkey: oh heck this whole page is very out of date
i have regrets

mildwonkey:
Ok, take 3: I replaced references to releases.hashicorp.com with a link to the registry, and fixed these references: they are referring to where terraform installs provider binaries (`.terraform/providers`), not the legacy `.terraform/plugins` directory where users can place binaries. Sorry!

mildwonkey:
And take 4, removing the entire table under `Discovery`: I left the helpful intro paragraph, but the link right below that has *all* the information.

alisdair (approved):
🎉

PR #1686, opened 2021-04-19
Additional note for the bootstrap file permissions for TFE installation

hashitop:
## Labels

- [X] inaccuracy
- [X] clarification
- [ ] new docs

Additional note for the bootstrap file permissions for TFE installation

### DIFF
--- a/content/source/docs/enterprise/install/automating-the-installer.html.md
+++ b/content/source/docs/enterprise/install/automating-the-installer.html.md
@@ -18,6 +18,8 @@ You may also need to provide additional flags (such as the instance's public and
 
 This document expects that the user is already familiar with how to do a [manual install](./installer.html#installation).
 
+-> **Note**: The `/etc/replicated.conf`, the license file, and TLS certificate files need to allow read access from other groups.
+
 ## Application settings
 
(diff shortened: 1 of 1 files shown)


content/source/docs/enterprise/install/automating-the-installer.html.md
```diff
@@ -18,6 +18,8 @@ You may also need to provide additional flags (such as the instance's public and
 
 This document expects that the user is already familiar with how to do a [manual install](./installer.html#installation).
 
+-> **Note**: The `/etc/replicated.conf`, the license file, and TLS certificate files need to allow read access from other groups.
```
laurapacilio, line 21:
```suggestion
-> **Note**: The `/etc/replicated.conf`, license file, and TLS certificates need to allow read access from other groups.
```

laurapacilio (approved):
Reviewing for style and flow only (assuming content is correct). Tiny tiny nit - otherwise, you're good to merge :-)

laurapacilio:
Per conversation with @hashitop we've decided to close this PR as it's not needed anymore. Top raised this issue with Replicated and they actually fixed that across all versions of the installation script. So, only customers who are using old versions of the installation script (as in they downloaded it long ago and never revisited the latest version of the installation) would need to do this.

PR #1684, opened 2021-04-16
clean up custom docker image example

p0pr0ck5:

PR #1687, opened 2021-04-19
ignore known false positives in broken link checker

p0pr0ck5:
Some documents contain links that, by design, fail the broken link check logic (e.g., linking to resources that we control, that return an HTTP 301 or such). We should silence these false positives, so that we get a better signal-to-noise ration in CI output.

### DIFF
--- a/content/scripts/check-pr-links.rb
+++ b/content/scripts/check-pr-links.rb
@@ -28,6 +28,13 @@ ROOT_URL = 'http://localhost:4567/'
 VERCEL_ROUTES = ['/community', '/cloud']
 VERCEL_REGEXP = /^(#{VERCEL_ROUTES.join('|')})/
 
+# A few documents intentionally contain links that would be flagged by this script,
+# but that we don't want (or can't) "fix". Rather than continually failing CI checks,
+# let's silence those false positives so failures don't become ignored noise.
+IGNORE_PROBLEMS = {
+  "content/source/docs/enterprise/install/installer.html.md" => ["https://install.terraform.io/airgap/latest.tar.gz"],
+}
+
 ARGF.set_encoding('utf-8')
 input = ARGF.read
 input_files = input.split(/\x00|\n/)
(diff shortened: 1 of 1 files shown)


p0pr0ck5:
The first entry is a known "broken" link- the TFE installation docs point to a script that, by design, returns an HTTP 301. Since this is the desired behavior, we should silence this rather than continually see failed CI runs. See also https://github.com/hashicorp/terraform-website/pull/1684

jescalan:
I'm mostly confused by why a 301 is reported as a broken link. I found, I think, [the page and link in question](https://www.terraform.io/docs/enterprise/install/installer.html#prepare-the-instance) (click "download the installer bootstrapper"), and the link works correctly and downloads the intended file, as far as I can tell.

A 301 return code is not what I'd typically imagine to be called out as "broken", since it's a redirect. There are many 301s within all of HashiCorp's docs and it doesn't really make sense to me why this one specific one would stand out as a broken. @nfagerlund, I think, originally wrote this link checking script and might have a better perspective than I do here?

nfagerlund:
Hello! Right, okay, I was pretty sure I remembered this having come up before, and it turns out it was in a PR that I closed without merging because most of it ended up in another PR: 

https://github.com/hashicorp/terraform-website/pull/1643

So it looks like I was briefly considering basically the exact same approach that @p0pr0ck5 did here, but on further thought was like "wait, maybe we should just ignore 301s entirely."

So, I endorse ignoring 301s and not treating them as failures. My original intent was that for 301s _within the Terraform docs and learn.hashicorp.com,_ we should really be maintaining the proper link destinations, but for external stuff, treating them as errors is actually quite bad.

nfagerlund:
Okay, I filed a PR with an alternate take -- just follow 301/308s and consider them fine, but log a warning so that if you're fixing a real broken link, you can update that at the same time. https://github.com/hashicorp/terraform-website/pull/1688

nfagerlund:
Closing in favor of a more brute force approach. 😝

PR #1688, opened 2021-04-22
Follow 301 and 308 redirects in PR link checker (instead of erroring)

nfagerlund:
This came up in https://github.com/hashicorp/terraform-website/pull/1687.

- Return a recursive call with the new URL, so if it's 200 we're 200.
- Use URI.join in case Location is a relative (to the request) URL.
- Cheeky little guard for infinite redirect loops.
- Log a warning so you can update it if you're fixing something else anyway.


judithpatudith (approved):
love it! Thank you!

nfagerlund:
Okay, that last force-push sorted out the remaining errors and this should be good to go.

PR #1694, opened 2021-04-29
Amend VMware RA diagram to promote happy path

ml4:
Amending just the VMware RA arch diagram.
This is:
- based on discussions between CSA/IS staff on customer feedback and sentiment.
- promotes focus for readers on a primary (happy path) deployment and greys out, and somewhat sidelines the backup DC
- not to suggest not having a DR solution (also partly why no verbiage change is suggested at this time), just that a symmetrical diagram IS and CSA agree confuses customers and makes them think that there is HA inherent in the solution which there is not.
- not including commensurate changes to the High Availability section which will be submitted in a separate PR due to staff bandwidth, stems the flow of confused customers but does not downplay any large customers with sufficient skills to actually enact a fully HA solution (as still referenced possibly in the modified diagram).  More careful consideration to the text is required, especially in the light of the forthcoming backup and recovery/restoration RPs which will impact the code here.


straubt1 (approved):
Great work

PR #1710, opened 2021-05-12
Update active-active.html.md

dwaiba:
Redis Family can only be C or P

## Labels

- [x] inaccuracy
- [ ] clarification
- [ ] new docs


sudomateo:
Hello! This table uses the terminology listed on Azure's Redis Cache pricing page: https://azure.microsoft.com/en-us/pricing/details/cache/

That is why the Cache Name is set to values like `C3` where `C` is the family and `3` is the capacity.

Happy to have a conversation if you think it would be helpful to communicate family and capacity in addition to Cache Name.

laurapacilio:
Hi both - I'm going to close this out, as it looks like Matthew has answered your question and it doesn't look like we want to change what's in the table. @dwaiba let us know if you'd like to add a note to this page about the family and capacity in addition to cache name. Thank you!

PR #1719, opened 2021-05-19
Fix sentinel docs link

EnMod:
Quick fix for a missing character in a link in the docs for Sentinel.


mwickett (approved):
LGTM. Just noting that it's odd that the broken links checker didn't pick this up.

PR #1732, opened 2021-06-04
Document Ingress Attributes

chrisarcand:
Ingress attributes are resources which include commit information associated with configuration versions sourced from VCS.

image

### DIFF
--- a/content/source/docs/cloud/api/configuration-versions.html.md
+++ b/content/source/docs/cloud/api/configuration-versions.html.md
@@ -143,7 +143,7 @@ curl \
 }
 ```
 
-## Show a Configuration Version's commit information
+## Show a Configuration Version's Commit Information
 
 An ingress attributes resource (`ingress-attributes`) is used to reference commit information for configuration versions created in a workspace with a VCS repository.
 
--- a/content/source/docs/cloud/api/configuration-versions.html.md
+++ b/content/source/docs/cloud/api/configuration-versions.html.md
@@ -143,6 +143,82 @@ curl \
 }
 ```
 
+## Show a Configuration Version's commit information
+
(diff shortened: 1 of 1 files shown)


chrisarcand:
Mmm. As I dwell on this more and am adding this to the TFC/E Go client, I think it may make more sense documented literally under Configuration Versions (as that is their purpose, really).

radditude:
+1, I think that makes sense since they're not a stand-alone resource!

chrisarcand:
@radditude Done!

nfagerlund (approved):
This approach seems good! 

I forget, do we associate an `ingress-attributes` with a `policy-set-versions` created from VCS? The docs for policy sets don't show that, but they also seem to be focused on handling manually created versions, rather than examining automatic ones.

radditude:
Ah, shoot, we totally do - good catch @nfagerlund! I guess ingress attributes will need their own page after all 😞

content/source/docs/cloud/api/configuration-versions.html.md
```diff
@@ -143,6 +143,82 @@ curl \
 }
 ```
 
+## Show a Configuration Version's commit information
```
radditude, line 146: Seems like we usually stick with title case for these headings!

```suggestion
## Show a Configuration Version's Commit Information
```

radditude (approved):

nfagerlund:
Ah, cool, that would fit with what that comment said about it being a vestige of old TFE — we probably thought better of this when it came time to do that policy sets API.

chrisarcand:
> Ah, cool, that would fit with what that comment said about it being a vestige of old TFE — we probably thought better of this when it came time to do that policy sets API.

Very much this ^_^ Yes, internally the modeling uses this but from the API's perspective it's just an object attribute, not a separate resource.

Thanks you two!

PR #1739, opened 2021-06-16
Adding new article. Frequently asked topic on the Terraform support.

antonakv:
## Labels

- [ ] inaccuracy
- [ ] clarification
- [x] new docs


nfagerlund (changes requested):
Hey thanks! This PR is pointed in the right direction, but it'll need some additional attention from both you and from someone on @hashicorp/terraform-education before it's ready to ship. 

- I agree with adding a new page about this! However, I think this is probably the wrong spot for it; instead, I suggest putting it in the Terraform CLI docs, under the "Managing Plugins" section of the sidebar! Once it's there, it'll need to be woven into the rest of the site via links:
    - It should probably be linked from some pages here in the Extend section. 
    - It should definitely be linked from the providers-focused pages of the Language docs. 
    - Currently the only place where we document this is in the CLI config file page at https://www.terraform.io/docs/cli/config/config-file.html#implied-local-mirror-directories — it might be possible to tighten that inline description up and mostly link to the new page. 
    - Similarly, this new page should link to the broader info about configuring provider mirrors and installation methods in the CLI config file page.
    - Possibly there are some pages on Learn that will want to exploit this, but the proper thing to do there is just give the team a heads-up. 
- This PR seems to add a redundant copy of the extend index to the top level of the repo, so that definitely needs fixed.
- The grammar and flow of the page needs some revision to meet the standards set by the rest of the documentation. Hopefully someone from @hashicorp/terraform-education will be able to help more with that!

PR #1742, opened 2021-06-21
Update agent docs with new Docker user.

RussellRollins:
In an upcoming tfc-agent version, we will change the configuration of the Docker image to execute as a non-root user. This will break existing tfc-agent pools that relied on the ability to apt-get install software during a local-exec script. We advise against installing software in this way, but we still want to warn users explicitly about the incoming change.

### DIFF
--- a/content/source/docs/cloud/agents/index.html.md
+++ b/content/source/docs/cloud/agents/index.html.md
@@ -155,6 +155,25 @@ docker run -e TFC_AGENT_TOKEN=your-token -e
 TFC_AGENT_NAME=your-agent-name hashicorp/tfc-agent
 ```
 
+This Docker image executes the tfc-agent process as the non-root tfc-agent user. For some workflows, such as those that require the ability to install software via apt-get during local-exec scripts, you may need to build a customized version of the agent Docker image for your internal use.
+
+```Dockerfile
+FROM hashicorp/tfc-agent:latest
+
+USER root
+
+# Install sudo. The container runs as a non-root user, but people may rely on
+# the ability to apt-get install things.
+RUN apt-get -y install sudo
+
+# Permit tfc-agent to use sudo apt-get commands.
+RUN echo 'tfc-agent ALL=NOPASSWD: /usr/bin/apt-get , /usr/bin/apt' >> /etc/sudoers.d/50-tfc-agent
+
+USER tfc-agent
+```
+
+An image customized in this way will permit installation of additional software via sudo apt-get.
+
 ### Stopping the Agent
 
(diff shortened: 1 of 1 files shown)


content/source/docs/cloud/agents/index.html.md
```diff
@@ -155,6 +155,25 @@ docker run -e TFC_AGENT_TOKEN=your-token -e
...
+USER root
+
+# Install sudo. The container runs as a non-root user, but people may rely on
+# the ability to apt-get install things.
+RUN apt-get -y install sudo
+
+# Permit tfc-agent to use sudo apt-get commands.
+RUN echo 'tfc-agent ALL=NOPASSWD: /usr/bin/apt-get , /usr/bin/apt' >> /etc/sudoers.d/50-tfc-agent
+
+USER tfc-agent
```
ryanuber, line 172: There's a `USER root` above. Do we actually need to specify the user at all, given that the upstream base container would already have set it to `tfc-agent`? Not sure if that gets inherited or not.

ryanuber (commented):
This looks really good but I left a question about the Dockerfile.

content/source/docs/cloud/agents/index.html.md
RussellRollins: I believe both these `USER` directives are required. If you try to just inherit from the base image, you end up with:

```
$ docker build -t tfc-agent-sudoer -f Dockerfile.addsudo .
[+] Building 0.4s (5/6)                                                                                                                                                                                                                   
 => [internal] load build definition from Dockerfile.addsudo                                                                                                                                                                         0.0s
 => => transferring dockerfile: 363B                                                                                                                                                                                                 0.0s
 => [internal] load .dockerignore                                                                                                                                                                                                    0.0s
 => => transferring context: 2B                                                                                                                                                                                                      0.0s
 => [internal] load metadata for docker.io/library/tfc-agent:latest                                                                                                                                                                  0.0s
 => [1/3] FROM docker.io/library/tfc-agent                                                                                                                                                                                           0.1s
 => ERROR [2/3] RUN apt-get -y install sudo                                                                                                                                                                                          0.3s
------                                                                                                                                                                                                                                    
 > [2/3] RUN apt-get -y install sudo:                                                                                                                                                                                                     
#5 0.252 E: Could not open lock file /var/lib/dpkg/lock-frontend - open (13: Permission denied)
#5 0.252 E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), are you root?
```

It can't install sudo because it's still executing as `tfc-agent`.

If you only switch over to root in order to install `sudo` and modify `/etc/sudoers`, then you end up staying as root with:

```
$ docker build -t tfc-agent-sudoer -f Dockerfile.addsudo .
<elided>
$ docker run -it --entrypoint="" tfc-agent-sudoer whoami
root
```

ryanuber (approved):

PR #1745, opened 2021-06-25
call out which registry module apis that are being deprecated

bethanyr:
Update the api docs calling out which Registry Module apis are being deprecated (these endpoints were removed from the doc as part of api updates for June 8).

## Labels

- [ ] inaccuracy
- [ ] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/cloud/api/changelog.html.md
+++ b/content/source/docs/cloud/api/changelog.html.md
@@ -14,13 +14,20 @@ page_id: "api-changelog"
 Keep track of changes to the API for Terraform Cloud and Terraform Enterprise.
 
 ### 2021-06-8
-
 * Updated [Registry Module APIs](./modules.html).
     * added `registry_name` scoped APIs.
     * added `organization_name` scoped APIs.
     * added [Module List API](./modules.html#list-registry-modules-for-an-organization).
     * updated [Module Delete APIs](./modules.html#delete-a-module).
     * ![cloud][] added public registry module related APIs.
+* ![deprecation] [] The following [Registry Module APIs](./modules.html) have been replaced with newer apis and will be removed in the future.
+    * The following endpoints to delete modules are replaced with [Module Delete APIs](./modules.html#delete-a-module)
+        * `POST /registry-modules/actions/delete/:organization_name/:name/:provider/:version`
+        * `POST /registry-modules/actions/delete/:organization_name/:name/:provider`
+        * `POST /registry-modules/actions/delete/:organization_name/:name`
+    * `POST /registry-modules` replaced with [Updated POST Endpoint](./modules.html#publish-a-private-module-from-a-vcs)
+    * `POST /registry-modules/:organization_name/:name/:provider/versions` replaced with new [endpoint](./modules.html#create-a-module-version)
+    * `GET /registry-modules/show/:organization_name/:name/:provider` replaced with new [GET Endpoint](./modules.html#get-a-module)
 
 ### 2021-05-27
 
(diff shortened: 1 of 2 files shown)


content/source/docs/cloud/api/modules.html.md
```diff
@@ -747,6 +753,12 @@ curl \
 * `DELETE /organizations/:organization_name/registry-modules/:registry_name/:namespace/:name/:provider`
 * `DELETE /organizations/:organization_name/registry-modules/:registry_name/:namespace/:name`
 
+~> **Deprecation warning**: the following endpoints:
+  * `POST /registry-modules/actions/delete/:organization_name/:name/:provider/:version`
+  * `POST /registry-modules/actions/delete/:organization_name/:name/:provider`
+  * `POST /registry-modules/actions/delete/:organization_name/:name`
+are replaced by the above endpoints and will be removed from future versions of the API!.
```
jbonhag, line 760: There's a funky issue with the list formatting here, and it looks like adding a newline fixes the list but breaks the alert box:

```
~> **Deprecation warning**: the following endpoints:

  * `POST /registry-modules/actions/delete/:organization_name/:name/:provider/:version`
  * `POST /registry-modules/actions/delete/:organization_name/:name/:provider`
  * `POST /registry-modules/actions/delete/:organization_name/:name`
```

I wonder if there's a way we can invoke the inline notification using a block syntax?

content/source/docs/cloud/api/modules.html.md
```diff
@@ -218,6 +218,8 @@ curl \
 
 `POST /organizations/:organization_name/registry-modules/vcs`
 
+~> **Deprecation warning**: the following endpoint `POST /registry-modules` is replaced by the above endpoint and will be removed from future versions of the API!.
```
jbonhag, line 221:
```suggestion
~> **Deprecation warning**: the following endpoint `POST /registry-modules` is replaced by the above endpoint and will be removed from future versions of the API!
```

content/source/docs/cloud/api/changelog.html.md
```diff
@@ -21,6 +21,14 @@ Keep track of changes to the API for Terraform Cloud and Terraform Enterprise.
     * added [Module List API](./modules.html#list-registry-modules-for-an-organization).
     * updated [Module Delete APIs](./modules.html#delete-a-module).
     * ![cloud][] added public registry module related APIs.
+* ![deprecation] [] The following [Registry Module APIs](./modules.html) have been replaced with newer apis and will be removed in the future.
```
jbonhag, line 24: Might be worth calling attention to the benefits of the new endpoints -- e.g. better JSON:API compliance, more idiomatic HTTP verbs, etc.

jbonhag (changes requested):
Thanks for following up on this!

content/source/docs/cloud/api/changelog.html.md
bethanyr: The changelog is not supposed to have a lot of detail - mostly just calling out what is changed.

content/source/docs/cloud/api/modules.html.md
jbonhag: We could this, which is a bit of a hack but it works:

```
<div class="alert alert-warning" role="alert">
  **Deprecation warning**: the following endpoints:

  * `POST /registry-modules/actions/delete/:organization_name/:name/:provider/:version`
  * `POST /registry-modules/actions/delete/:organization_name/:name/:provider`
  * `POST /registry-modules/actions/delete/:organization_name/:name`

  are replaced by the above endpoints and will be removed from future versions of the API!
</div>
```

Screen Shot 2021-06-29 at 11 52 11 AM

bethanyr:
I have moved the deprecation warnings to the beginning of each api section

jbonhag (approved):
Thanks, LGTM! 👍

PR #1749, opened 2021-06-29
Clarifying language re: spec plans vs applies

vaficionado:
Added a blurb per support request identifying that while a Plan may pass, the status check is not updated if the associated Apply fails.

### DIFF
--- a/content/source/docs/cloud/run/ui.html.md
+++ b/content/source/docs/cloud/run/ui.html.md
@@ -86,6 +86,8 @@ Whenever a pull request is _created or updated,_ Terraform Cloud checks whether 
 
     -> **Note:** If Terraform Cloud skips a plan because the changes weren't relevant, it will still post a passing commit status to the pull request.
 
+- Status checks on the pull request are not updated after an apply, so a commit with a successful plan but an errored apply will still show the passing commit status relevant to the plan.
+
 ### Contents of Pull Request Plans
 
 Speculative plans for pull requests use the contents of the head branch (the branch that the PR proposes to merge into the destination branch), and they compare against the workspace's current state at the time of the plan. This means that if the destination branch changes significantly after the head branch is created, the speculative plan might not accurately show the results of accepting the PR. To get a more accurate view, you can rebase the head branch onto a more recent commit, or merge the destination branch into the head branch.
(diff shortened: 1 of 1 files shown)


content/source/docs/cloud/run/ui.html.md
```diff
@@ -86,6 +86,8 @@ Whenever a pull request is _created or updated,_ Terraform Cloud checks whether
 
     -> **Note:** If Terraform Cloud skips a plan because the changes weren't relevant, it will still post a passing commit status to the pull request.
 
+- Status checks on the pull request are not updated after an apply, so a commit with a successful plan but an errored apply will still show the passing commit status relevant to the plan.
```
nfagerlund, line 89: Hmm, took me a few passes to understand that, but it does seem like an important thing to point out. How about this? 

```suggestion
- Commit statuses only show the results of speculative plans. If a pull request is merged and results in a normal (non-speculative) run, Terraform Cloud will not update the pull request's status checks with the results (for example, if an apply partially fails).
```

Or that might prompt you to think of a better way to describe that; either way, I'll go ahead and pre-approve.

nfagerlund (approved):

content/source/docs/cloud/run/ui.html.md
judithpatudith, line 89: Just a suggestion to remove some passive voice, and a clarification. Is the apply we're talking about here the apply that results from merging the PR itself?

```suggestion
- Terraform Cloud does not update the status checks on the pull request with the status of the associated apply. A commit with a successful plan but an errored apply will still show the passing commit status from the plan.
```

judithpatudith (approved):

content/source/docs/cloud/run/ui.html.md
vaficionado: > Just a suggestion to remove some passive voice, and a clarification. Is the apply we're talking about here the apply that results from merging the PR itself?

Yes! That's correct. I'm going to propose another version of this that merges both you and Nick's comments:

```suggestion
- Terraform Cloud does not update the status checks on a pull request with the status of an associated apply. This means that a commit with a successful plan but an errored apply will still show the passing commit status from the plan.
```

Sound good @nfagerlund @judithpatudith ?

beekus (approved):

PR #1754, opened 2021-07-01
Add note about cost estimation enabled by default

ritsok:


im2nguyen (approved):
First sentence is passive voice, but it sounds ok

tr0njavolta:
Do the docs have guidance on passive voice? I don't think they do but `The Team & Governance plan enables cost estimation by default` may work. I prefer the change in the PR as well though.

PR #1757, opened 2021-07-02
Add locked_by to workspace includes

nfagerlund:
Two small actual changes to the API, and an overdue update of some stuff in the state-versions API.

The stuff mentioned in the changelog is deployed to production.

### DIFF
--- a/content/source/docs/cloud/api/state-versions.html.md
+++ b/content/source/docs/cloud/api/state-versions.html.md
@@ -163,96 +163,180 @@ curl \
 {
     "data": [
         {
-            "id": "sv-SDboVZC8TCxXEneJ",
+            "id": "sv-g4rqST72reoHMM5a",
             "type": "state-versions",
             "attributes": {
-                "vcs-commit-sha": null,
-                "vcs-commit-url": null,
-                "created-at": "2018-08-27T14:49:47.902Z",
+                "created-at": "2021-06-08T01:22:03.794Z",
+                "size": 940,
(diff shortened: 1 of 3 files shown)


content/source/docs/cloud/api/state-versions.html.md
```diff
@@ -163,96 +163,180 @@ curl \
...
+                        "count": 1,
+                        "module": "root",
+                        "provider": "provider[\"registry.terraform.io/hashicorp/null\"]"
+                    }
+                ],
+                "resources-processed": false,
+                "serial": 9,
+                "state-version": 4,
+                "terraform-version": "0.15.4",
+                "vcs-commit-url": "https://gitlab.com/nfagerlund/terraform-minimum/-/commit/5cffb47f35a2ab2ab8faf8f07b030c9099775a6f",
```
radditude, line 206: I'm not sure if you meant to include this URL + the VCS URLs in the other responses - would it be better to replace those with anonymized/dummy data?

```suggestion
                "vcs-commit-url": "https://gitlab.com/my-organization/terraform-test/-/commit/abcdef12345",
```

radditude (approved):
Nice! One small note about dummy-data-ifying the VCS urls in the responses, but that's not really a blocker.

content/source/docs/cloud/api/state-versions.html.md
nfagerlund: I didn't, thank you for catching that! Caught a bunch of the other identifiers, but forgot that one.

PR #1573, opened 2020-12-19
Delete ALL remaining provider submodules and symlinks

nfagerlund:
**Draft status: Still waiting for eight providers to get published/redirected.** 

This change must not be merged to master until all of the remaining providers
have been published and redirected to the registry. 

See also: https://github.com/hashicorp/terraform-website/pull/1604

PR #1764, opened 2021-07-08
Redirect last few legacy providers

nfagerlund:
N.b.: This PR is targeted at the special branch that (via a terraform config) governs the redirects for providers that were migrated to the registry. 

Blocker on #1573. 

This is the end (almost) of a many-months-long effort to finally unpublish the last legacy provider documentation hanging around on terraform.io. It fixes one broken redirect, adds three redirects for providers that have officially migrated, and shunts the last four abandoned providers off to their archived github repos. 

After these last eight providers are redirected, we can:

- Remove their submodules on master
- Merge the registry/provider-sunset-transition branch to master
- Recenter the `hashicorp-terraform/static-sites-terraform-website-fastly` workspace onto master
- Remove the old provider documentation links from the page about provider documentation


judithpatudith (approved):
Thank you Nick!

bflad (approved):
Looks good to me -- verified the Fastly redirects and the external redirects are appropriate and valid.

PR #1766, opened 2021-07-09
Add a warning for TFE + external Vault

amy-hashi:
## Labels

- [ ] inaccuracy
- [x] clarification
- [ ] new docs

This changes adds an additional warning to using external Vault with TFE. Due to our current lack of support for custom paths, we cannot support two disparate TFE instances pointing to the same Vault instance/cluster.


aaron-lane (approved):
LGTM ⚠️

PR #1768, opened 2021-07-09
Redirect legacy provider docs index

nfagerlund:
The fact that you need three redirects if it's an index.html file is a weird old quirk of this relict website builder.

More context: https://github.com/hashicorp/terraform/pull/29134


laurapacilio (approved):
Thank you!

PR #1771, opened 2021-07-14
TFC Actions menu: Update references to locking and the queue plan form

nfagerlund:
**Draft status:** This is behind a feature flag right now, so the docs can't be merged yet. 

TFC workspaces have a new interface for manually starting runs and managing the
lock on a workspace; this updates (I think!) all of the places where we refer to
the old ways of doing those things.

It doesn't update the screenshots, because:

- We didn't have existing screenshots specifically dedicated to doing those actions.
- The number of screenshots that _incidentally_ show the old workspace header
  interface are too numerous. We need a better way to manage these! :coolcrying:

### DIFF
--- a/content/source/docs/cloud/run/modes-and-options.html.md
+++ b/content/source/docs/cloud/run/modes-and-options.html.md
@@ -19,12 +19,13 @@ Terraform Cloud runs support many of the same modes and options available in the
 
 > **Hands-on:** Try the [Use Refresh-Only Mode to Sync Terraform State](https://learn.hashicorp.com/tutorials/terraform/refresh) tutorial on HashiCorp Learn.
 
--> **Version note:** Refresh-only support is available in Terraform Enterprise v202106-1 or later and requires a workspace using at least Terraform CLI v0.15.2.
+-> **Version note:** Refresh-only support is available in Terraform Enterprise v202106-1 or later and requires a workspace using at least Terraform CLI v0.15.4.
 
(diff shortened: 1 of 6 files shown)


content/source/docs/cloud/run/modes-and-options.html.md
```diff
@@ -19,12 +19,13 @@ Terraform Cloud runs support many of the same modes and options available in the
 
 > **Hands-on:** Try the [Use Refresh-Only Mode to Sync Terraform State](https://learn.hashicorp.com/tutorials/terraform/refresh) tutorial on HashiCorp Learn.
 
--> **Version note:** Refresh-only support is available in Terraform Enterprise v202106-1 or later and requires a workspace using at least Terraform CLI v0.15.2.
+-> **Version note:** Refresh-only support is available in Terraform Enterprise v202106-1 or later and requires a workspace using at least Terraform CLI v0.15.4.
```
radditude, line 22: Oh whoops, good catch - it was [`replace` that landed in 0.15.2](https://github.com/hashicorp/terraform/blob/v0.15/CHANGELOG.md#0152-may-05-2021), not `refresh-only`!

radditude (approved):

content/source/docs/cloud/run/modes-and-options.html.md
nfagerlund: Hmm, maybe I should update that below, while I'm in here!

PR #1772, opened 2021-07-14
Add a warning to the docs that TFC Agent doesn't support TFE_PARALLELISM

RussellRollins:
TFC Agent doesn't support `TFE_PARALLELISM`, but that's okay, because Terraform supports setting flags via environment variables directly now.

## Labels

- [ ] inaccuracy
- [x] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/cloud/workspaces/variables.html.md
+++ b/content/source/docs/cloud/workspaces/variables.html.md
@@ -116,7 +116,7 @@ Terraform Cloud uses some special environment variables to control dangerous or 
 
 - `TFE_PARALLELISM` — If present, Terraform Cloud uses this to set `terraform plan` and `terraform apply`'s `-parallelism=<N>` flag ([more info](/docs/internals/graph.html#walking-the-graph)). Valid values are between 1 and 256, inclusive; the default is `10`. This is rarely necessary, but can fix problems with infrastructure providers that error on concurrent operations or use non-standard rate limiting. We recommend talking to HashiCorp support before using this.
 
-Please note that `TFE_PARALLELISM` is not supported by Terraform Cloud Agents. Modern versions of Terraform allow you to specify flags as environment variables directly via [TF_CLI_ARGS](/docs/cli/config/environment-variables.html#tf-cli-args). Use `TF_CLI_ARGS="parallelism=<N>"` instead.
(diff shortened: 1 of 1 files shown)


content/source/docs/cloud/workspaces/variables.html.md
```diff
@@ -116,6 +116,8 @@ Terraform Cloud uses some special environment variables to control dangerous or
 
 - `TFE_PARALLELISM` — If present, Terraform Cloud uses this to set `terraform plan` and `terraform apply`'s `-parallelism=<N>` flag ([more info](/docs/internals/graph.html#walking-the-graph)). Valid values are between 1 and 256, inclusive; the default is `10`. This is rarely necessary, but can fix problems with infrastructure providers that error on concurrent operations or use non-standard rate limiting. We recommend talking to HashiCorp support before using this.
 
+Please note that `TFE_PARALLELISM` is not supported by Terraform Cloud Agents. Modern versions of Terraform allow you to specify flags as environment variables directly via [TF_CLI_ARGS](/docs/cli/config/environment-variables.html#tf-cli-args). Use `TF_CLI_ARGS="parallelism=<N>"` instead.
```
laurapacilio, line 119: Adding the note callout notation which will make this appear as a blue note box! I recommend doing this so this stands out. Also - when we say "modern versions" should we specify a version number to be more specific?

```suggestion
-> **Note:** `TFE_PARALLELISM` is not supported by Terraform Cloud Agents, but modern versions of Terraform allow you to specify flags as environment variables directly via [TF_CLI_ARGS](/docs/cli/config/environment-variables.html#tf-cli-args). Use `TF_CLI_ARGS="parallelism=<N>"` instead.
```

laurapacilio (approved):
Hi! I'm assuming everything is correct, so I just reviewed for style and flow. Made two tiny nit changes and had one question for you about whether we need to specify what "modern versions" means. 

And you can merge even though CircleCI link check is failing because you added the missing page anchor in this PR: https://github.com/hashicorp/terraform/pull/29163 

Thank you for adding this!

content/source/docs/cloud/workspaces/variables.html.md
RussellRollins: This environment variable is quite old. I believe it was [introduced in 0.9.0](https://github.com/hashicorp/terraform/blob/b0711a0585c44f6ebb1a7c7ef8bedd5de8ac17ac/CHANGELOG.md#L1405). My intention with "modern" was just to explain why `TFE_PARALLELISM` ever existed. I'm fine with being specific, but TFC doesn't actually support anything older than 0.9.0.

PR #1786, opened 2021-07-26
Scope broken link checker to just docs page

im2nguyen:
## Labels

- [ ] inaccuracy
- [ ] clarification
- [ ] new docs

Broken link checker is throwing an error. Testing locally, this is the output, which is causing similar errors in the CI pipelines. Proposing we scope the link checker to just the docs which works as expected. In addition, this repo only touches the docs so this makes more sense imo.

Proposed:
```
$ linkcheck https://www.terraform.io/docs/
Perfect. Checked 170670 links, 832 destination URLs (1273 ignored).
```

Current:
```
$ linkcheck https://www.terraform.io/
Crawling: 0Unhandled exception:
Bad state: Element <html link> does not have any of the attributes [href, src]
#0      extractLink (package:linkcheck/src/parsers/html.dart:36)
#1      parseHtml.<anonymous closure> (package:linkcheck/src/parsers/html.dart:125)
#2      MappedListIterable.elementAt (dart:_internal/iterable.dart:415)
#3      ListIterable.toList (dart:_internal/iterable.dart:219)
#4      parseHtml (package:linkcheck/src/parsers/html.dart:127)
#5      checkPage (package:linkcheck/src/worker/worker.dart:169)
#6      _RootZone.runUnary (dart:async/zone.dart:1379)
## ...
Done crawling.
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,8 @@ jobs:
       - run:
           name: check links in changed pages
           working_directory: content
-          command: git diff --name-only -z --diff-filter=AMRCT $(git merge-base HEAD origin/master)..HEAD | bundle exec ./scripts/check-pr-links.rb
+          command:
+            git diff --name-only -z --diff-filter=AMRCT $(git merge-base HEAD origin/master)..HEAD | bundle exec ./scripts/check-pr-links.rb
             # --name-only: Return a list of affected files but don't show the changes.
             # -z: Make that a null-separated list (instead of newline-separated), and
             #     DON'T mangle non-ASCII characters.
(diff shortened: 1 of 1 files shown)


laurapacilio (approved):

.circleci/config.yml
```diff
@@ -52,7 +52,8 @@ jobs:
       - run:
           name: check links in changed pages
           working_directory: content
-          command: git diff --name-only -z --diff-filter=AMRCT $(git merge-base HEAD origin/master)..HEAD | bundle exec ./scripts/check-pr-links.rb
+          command:
+            git diff --name-only -z --diff-filter=AMRCT $(git merge-base HEAD origin/master)..HEAD | bundle exec ./scripts/check-pr-links.rb
```
nfagerlund, line 56: YAML is always so disorienting to me, so I'm counting on you to confirm that this syntax has the same effect. 😅

nfagerlund (approved):
Yeah, ok, this is a good and reasonably parsimonious way around the problem. (I wonder what the heck the problem actually is, since I can't find anything of the sort in web inspector, but you're right, it _totally doesn't matter_ w/r/t the contents of this repo.)

.circleci/config.yml
im2nguyen: i believe so, it was just an autoformatting thing 😄

PR #1787, opened 2021-07-26
Fix broken links on security model page

laurapacilio:
Fixing a few broken links that happened as a result of a recent PR. @eastebry can you take a look? Thank you! :-) 

- [ ] inaccuracy
- [ ] clarification
- [ ] new docs


eastebry (approved):
Looks good. Thanks!

PR #1788, opened 2021-07-26
Fixing issue with bulleted list not being formatted properly

eastebry:
Labels
- [ ] inaccuracy
- [ ] clarification
- [ ] new docs
- [X] cosmetic bug - fixing broken text or markup

Small issue - these bulleted lists aren't showing up properly. Seems like we forgot a newline before the start of the list


laurapacilio (approved):
Ah I should have caught this! thank you for fixing

PR #1792, opened 2021-07-27
Add note that TFE does not support redis cluster.

amy-hashi:
## Labels

- [ ] inaccuracy
- [x] clarification
- [ ] new docs

This PR adds a small note that TFE does not support Redis Cluster for Active/Active deployments.


laurapacilio (changes requested):
Thank you for adding! Made a formatting suggestion that I'm hoping will help make sure that users don't skip this information and that the amount of notes in this section doesn't get overwhelming. Please make sure I didn't materially change the meaning!

laurapacilio (approved):

PR #1793, opened 2021-07-27
content/source/docs: Replace golang.org/pkg links with pkg.go.dev

bflad:
Go has recently migrated standard library packages from golang.org/pkg to pkg.go.dev. Adjusting these links will prevent the link checker failures for the redirects.

## Labels

- [ ] inaccuracy
- [ ] clarification
- [ ] new docs


laurapacilio (approved):
Thank you for doing this! :)

PR #1813, opened 2021-08-12
psql - Added other managed services

roberthstrand:
For Managed PostgreSQL, there were only references to the AWS offering. Changed this to reflect services across the three major clouds, which should be the minimum for any vendor neutral documentation.

I see that there are more things related only to Amazon on this page. I will go through more documentation and see what I can do to help from an Azure perspective here, when time permits.

## Labels

- [ ] inaccuracy
- [X] clarification
- [ ] new docs


laurapacilio (approved):
Thank you! I think this is a great edit - I just want someone from eng. to approve as well before we merge. @sudomateo would you be willing to take a look?

sudomateo (approved):

PR #1815, opened 2021-08-13
add tags to tfrun docs

thrashr888:
adds tags to the tfrun docs https://www.terraform.io/docs/cloud/sentinel/import/tfrun.html

we missed 1 thing in #1765


cam-stitt (approved):
LGTM

Issue #1669, opened 2021-03-29
Update show workspace API sample response


assareh:
Now that we've added agents and agent pools, it would be good to update the Show Workspaces [sample response](https://www.terraform.io/docs/cloud/api/workspaces.html#sample-response-3). The current response includes agent-pool in relationships.

laurapacilio:
Hi! I put up a little PR to address this, if you'd like to review: https://github.com/hashicorp/terraform-website/pull/1818

laurapacilio:
And now that PR has been merged! So I'm going to close this out. :-)

PR #1818, opened 2021-08-16
Update API Show Workspace sample response

laurapacilio:
This is in response to [this ticket](https://github.com/hashicorp/terraform-website/issues/1669) which said that the show workspace sample response on [this page](https://www.terraform.io/docs/cloud/api/workspaces.html#sample-response-3) was out-of-date (the latest response includes agent-pools, among other things).

- [x] inaccuracy
- [x] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/cloud/api/workspaces.html.md
+++ b/content/source/docs/cloud/api/workspaces.html.md
@@ -1802,38 +1802,110 @@ $ curl \
 {
   "data": {
     "attributes": {
+      "actions": {
+        "is-destroyable": true
+      },
+      "allow-destroy-plan": true,
+      "apply-duration-average": null,
       "auto-apply": false,
-      "allow-destroy-plan": false,
-      "can-queue-destroy-plan": false,
-      "created-at": "2017-11-02T23:24:05.997Z",
-      "description": null,
+      "auto-destroy-at": null,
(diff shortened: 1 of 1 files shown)


assareh (approved):
LGTM

content/source/docs/cloud/api/workspaces.html.md
```diff
@@ -174,59 +174,101 @@ _Without a VCS repository_
 ```json
 {
   "data": {
-    "id": "ws-YnyXLq9fy38afEeb",
```
thrashr888, line 177: I think I would keep this order of this top level since it’s easier to follow what the data actually is. It makes it easier to understand it. I don’t think this is a blocker for merging, though, and if sorting makes this content easier to maintain then that’s fine too. 👍🏻

thrashr888 (approved):

content/source/docs/cloud/api/workspaces.html.md
laurapacilio: Yeah, that's a fair point.  Okay i'm going to merge this in now so that the responses are correct (even if they're sorted in an unexpected way) and then I'll make a note to maybe come back and work on this some more.

PR #1822, opened 2021-08-17
Fix policy set parameter API examples

tism:
The policy set id format in the examples for policy set parameters uses the individual policy resource format, `pol-` instead of the set format, `polset-`.

- [x] inaccuracy
- [ ] clarification
- [ ] new docs


dylanegan (approved):

dylanegan:
Many thanks! 🙇‍♂️

PR #1831, opened 2021-08-23
Ab tfe uninstall

amy-hashi:
## Labels

- [ ] inaccuracy
- [ ] clarification
- [x] new docs

This PR adds a new section to the Terraform Enterprise documentation with instructions on how make use of the new uninstaller script.


content/source/docs/enterprise/install/uninstall.html.md
```diff
@@ -0,0 +1,62 @@
+---
+layout: "enterprise"
+page_title: "Uninstall - Terraform Enterprise"
+---
+
+# Uninstall Terraform Enterprise
+
+If you installed Terraform Enterprise on VMWare instances, you may not be able to easily request new virtual machines for a broken or corrupted installation. Instead, you can use the `uninstall` script to remove Terraform Enterprise and all its services (excluding Docker) from a system.
```
laurapacilio, line 8:
```suggestion
If you installed Terraform Enterprise on VMWare instances, you may not be able to easily request new virtual machines for a broken or corrupted installation. Instead, you can use the `uninstall` script to remove Terraform Enterprise and all of its services (excluding Docker) from a system.
```

content/source/docs/enterprise/install/uninstall.html.md
```diff
@@ -0,0 +1,62 @@
...
+  - Select `Skip this step` to leave all Docker volumes intact.  
+![Uninstall - Prune all Docker volumes][uninstall-prune-all]
+
+3. The script removes dangling Docker volumes and the Docker networks that were created for the application. Terraform Enterprise has been removed from the system.
+
+![uninstall-cleanup][uninstall-cleanup]
+
+~> **Important**: This script does not touch the mounted disk path, so you will need to manually clean it up if necessary.
+
+## Run the Uninstaller
```
laurapacilio, line 35: To me it makes sense for this to come first on the page before we talk about what happens while the script runs, but open to pushback if we feel strongly that it should be second.

laurapacilio (changes requested):
I'm making a few minor comments (this is looking really great). But one thing I was thinking is that maybe we shouldn't do screenshots of the output. I say this because:
- I feel like they're not very accessible
- I feel like they might be harder to maintain if things change
- We don't do this for console output in our docs anywhere else I can see

I wonder if instead we could put the output in code blocks. For example - see how we do that here on HashiCorp Learn? https://learn.hashicorp.com/tutorials/terraform/aws-build?in=terraform/aws-get-started. 

Thoughts? Open to feedback from folks about the best way to do this.

content/source/docs/enterprise/install/uninstall.html.md
```diff
@@ -0,0 +1,62 @@
...
+
+## While the Script Runs
+
+After you initiate `uninstall`:
+
+1. Enter `yes` when asked if you want to continue with the installation.
+
+![Uninstall - Do you want to continue?][uninstall-prompt]
+
+   The script stops the Replicated services, removes the Docker containers, and removes replicated executables and configuration files from the system.
```
sudomateo, line 20:
```suggestion
   The script stops the Replicated services, removes the Docker containers, and removes Replicated executables and configuration files from the system.  
```

content/source/docs/enterprise/install/uninstall.html.md
sudomateo: I'm open to this as well.

sudomateo (approved):
Approving with a non-blocking suggestion. Great work Amy!

content/source/docs/enterprise/install/uninstall.html.md
```diff
@@ -0,0 +1,229 @@
...
+	retraced-postgres
+	Removing Replicated files and executables...
+	Run systemctl daemon-reload...
+	Terraform Enterprise and Replicated should now be uninstalled.
+	```
+
+   The script stops the Replicated services, removes the Docker containers, and removes Replicated executables and configuration files from the system.  
+
+
+2. Choose an option to prune Docker volumes:
```
laurapacilio, line 71: This is showing up as oddly indented when I test locally - is there any way to fix it?

content/source/docs/enterprise/install/uninstall.html.md
```diff
@@ -0,0 +1,229 @@
...
+	1bd96c43b48d2634b251ead68575a7bfba255f52a9909d2f0c00a5f5dd6cd14e
+	8388555131dd1ec8f9d9d218295d35ff287f9038430848ee2d721fd92c896b83
+	473ec860bc666b3265e4c50415bca607011f857024b8469f5aeffe0a251935e8
+	0a0cb265a2bfa773086d2a98a3a6412c1d54eaacad6706f319315485c9825996
+	dbf1b59e8b8451c0ece04fb57fc42aef4b6accb53fbcbe5850ed873a806f64a9
+	73bccf22ef84311ccf6dc2fa608191c8321fd77b4444c0406933c0eca7886bc9
+	```
+
+
+3. The script removes dangling Docker volumes and the Docker networks that were created for the application. Terraform Enterprise has been removed from the system.
```
laurapacilio, line 126: We don't need a number here because the user doesn't need to do anything.

content/source/docs/enterprise/install/uninstall.html.md
```diff
@@ -0,0 +1,229 @@
...
+	
+	Usage:  docker volume rm [OPTIONS] VOLUME [VOLUME...]
+	
+	Remove one or more volumes
+	Unable to remove dangling Docker volumes, or none to be removed.
+	
+	Uninstall Complete
+	```
+
+	You may see errors such as the ones above, these are normal and just mean the script is attempting to clean up something that's already been removed.
```
laurapacilio, line 203: I like this and I would move it up to be right before the "Terraform has been removed from the system." I worry that it's a bit hidden down here and I want folks to see this information about the errors.

laurapacilio (approved):
Nice work! I think the results in the code block play a lot better. Last few minor nits and I still think we should switch the order of these sections, but I'll approve so I don't block you!

content/source/docs/enterprise/install/uninstall.html.md
amy-hashi: I think this is fixed now.

PR #1844, opened 2021-08-30
Fix typo in the payload example making JSON invalid

achuchulev:
There is a typo within the documentation in the payload example of `Invite a User to an Organization` endpoint where there is an extra "," in the JSON which makes it invalid.


laurapacilio (approved):
Thank you!

PR #1853, opened 2021-09-06
Updated body for Module creation with no VCS

cerocool1203:
Namespace cannot be set for a private module, this value is no longer required as per my testing on TFE v202106-1. I was getting the below error when uploading the module via API
b'{"errors":[{"status":"422","title":"unprocessable entity","detail":"Validation failed: Namespace cannot be set for a private module"}]}'
Not 100% sure if the namespace attribute is still required for public

## Labels

- [ ] inaccuracy
- [X] clarification
- [ ] new docs


laurapacilio:
Hi @cerocool1203, thank you for submitting this! :-) 

I talked to the team, and we shouldn't remove this field from the table entirely because it's still required for public modules. So, we do think that it should be removed from the private modules example (thank you!) and then we think we should update the description for this field to specify that it can only be set for public modules. Maybe something like:

"The namespace for public modules. Cannot be set for private modules. May contain alphanumeric characters, with dashes and underscores allowed in non-leading or trailing positions. Maximum length is 64 characters."

Would you be open to updating your PR with this change?  Thank you!

cerocool1203:
Hi @laurapacilio, I had to create a new PR as I sort of messed up my branch as I did the change in the GUI 🙄 The corresponding PR is on 
https://github.com/hashicorp/terraform-website/pull/1867

cerocool1203:
Hi @laurapacilio, I had to create a new PR as I sort of messed up my branch as I did the change in the GUI 🙄 The corresponding PR is on https://github.com/hashicorp/terraform-website/pull/1868

PR #1856, opened 2021-09-08
Add a note clarification to write permissions

nikcbg:
The documentation is not clear enough about the write permissions specifically the lock/unlock part where only the holder of a locked workspace can unlock it.

## Labels

- [ ] inaccuracy
- [x] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/cloud/users-teams-organizations/permissions.html.md
+++ b/content/source/docs/cloud/users-teams-organizations/permissions.html.md
@@ -69,7 +69,7 @@ The following workspace permissions can be granted to teams on a per-workspace b
     - **Read runs:** — Allows users to view information about remote Terraform runs, including the run history, the status of runs, the log output of each stage of a run (plan, apply, cost estimation, policy check), and configuration versions associated with a run.
     - **Queue plans:** — _Implies permission to read runs._ Allows users to queue Terraform plans in a workspace, including both speculative plans and normal plans. Normal plans must be approved by a user with permission to apply runs.
     - **Apply runs:** — _Implies permission to queue plans._ Allows users to approve and apply Terraform plans, causing changes to real infrastructure.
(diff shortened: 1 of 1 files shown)


laurapacilio:
Thank you for spotting this! :-) Before I provide feedback, I have a clarifying question:

Do we grant lock/unlock permissions to teams, to individual users, or both? The [docs](https://www.terraform.io/docs/cloud/users-teams-organizations/permissions.html#general-workspace-permissions) imply that this is a permission that is granted by team. In that case, if someone else on my team locks my workspace, I will be able to unlock it. But if it's on a per user basis, it could be the case that if someone else on my team locks the workspace, I won't be able to unlock it (only they can). 

This affects the way that we will write this note, so I want to make sure I get the full picture of how permissions are granted!

nikcbg:
Hello Laura,

The lock/unlock permissions are granted to teams. The problem is that if
someone on my team locks a workspace, I am unable to unlock that same
workspace even though I have the same permissions as my teammate, only they
can unlock it. For example, let say I have a team called test with two
members John and Adam, if John locks a workspace only John can unlock it,
Adam cannot unlock it. So whoever is the holder of the lock, only that
holder can unlock a workspace.
I hope this helps and if you need anything else please let me know.

laurapacilio:
Okay this is very helpful, and now I understand why we want to clarify this in the documentation! :-) I have two suggestions. 

First, I think we want to add a line in "General workspace permissions" under lock/unlock that explains how even though this is a team-based permission, once a user sets a lock on a workspace, only that user can unlock it. Could you add that content in up there, and then I'll review?

Then I think that in this write section, let's move your note so that it's right underneath the lock/unlock bullet. And then let's say something like "Typically, only the user who created the lock can unlock a workspace. Administrators and members of the "Owners" team can use `Force unlock` to override this an unlock a workspace."

laurapacilio:
@nikcbg  Should we close this PR, since you opened that second one? Just wanted to check with you!

nikcbg:
Hi Laura, 

Yes, this one can be closed. I will comment on the other one.

Thanks.

laurapacilio:
Closing, per our conversation - thank you :-)

PR #1861, opened 2021-09-09
Some wording changes...

amy-hashi:
## Labels

- [ ] inaccuracy
- [x] clarification
- [ ] new docs


laurapacilio (approved):
Yes! and then can you just remove the "the script stops the replicated services" sentence from below the code block so it's not duplicated? Thank you :-)

PR #1863, opened 2021-09-10
docs/plugin/framework: Clarify new capabilities provided by v0.3.0

bflad:
Closes hashicorp/terraform-plugin-framework#139

## Labels

- [ ] inaccuracy
- [x] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/plugin/framework/types.html.md
+++ b/content/source/docs/plugin/framework/types.html.md
@@ -236,6 +236,9 @@ State, Config, and Plan](/docs/plugin/framework/accessing-values.html).
 
 You may want to build your own attribute value and type implementations to allow your provider to combine validation, description, and plan customization behaviors into a reusable bundle. This helps avoid duplication or reimplementation and ensures consistency.
 
+~> **Important:** Specifying plan customization for attribute types is not yet
+supported, limiting their utility. Support is expected in the near future.
+
 ### `attr.Type` Interface
 
 Use the [`attr.Type`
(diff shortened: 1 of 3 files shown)


content/source/docs/plugin/framework/types.html.md
```diff
@@ -236,10 +236,6 @@ State, Config, and Plan](/docs/plugin/framework/accessing-values.html).
 
 You may want to build your own attribute value and type implementations to allow your provider to combine validation, description, and plan customization behaviors into a reusable bundle. This helps avoid duplication or reimplementation and ensures consistency.
 
-~> **Important:** Specifying validation and plan customization for attribute
```
paddycarver, line 239: I think plan customization is still not supported.

paddycarver (changes requested):

content/source/docs/plugin/framework/types.html.md
bflad: Are you sure? There is now [`tfsdk.Attribute.PlanModifiers`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework@v0.3.0/tfsdk#Attribute.PlanModifiers), unless plan customization means something something else.

paddycarver: The full quote is "Specifying validation and plan customization for attribute **types** is not yet supported." We [have that](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/attr#TypeWithValidate) for Validation, but not for plan modification. Though we do, as you rightly point out, have it for listing modifiers on a schema attribute, we can't attach plan modification to an `attr.Type` just yet.

bflad: Ah sorry, was missing the _type_ part there. 🤦  I will add this back in with just a note about plan customization on types.

paddycarver (approved):
LGTM, let's do it. Awesome job. 👍 🚀

PR #1868, opened 2021-09-11
Docs: added changes specific to private modules

cerocool1203:
Namespace cannot be set for a private module, this value is no longer required as per my testing on TFE v202106-1. I was getting the below error when uploading the module via API
b'{"errors":[{"status":"422","title":"unprocessable entity","detail":"Validation failed: Namespace cannot be set for a private module"}]}'
Not 100% sure if the namespace attribute is still required for public
## Labels

- [X] inaccuracy
- [ ] clarification
- [ ] new docs
- [X] api - requires an update to the [changelog](https://www.terraform.io/docs/cloud/api/changelog.html)


cerocool1203:
Hi @laurapacilio , Applied changes as requested on https://github.com/hashicorp/terraform-website/pull/1853

laurapacilio (approved):
Thank you!

PR #1874, opened 2021-09-16
Remove providers index symlink

EnMod:
This PR removes the symlink at `./content/source/docs/providers/index.html.markdown`.

## Labels

- [ ] inaccuracy
- [ ] clarification
- [ ] new docs


nfagerlund (approved):
Confirmed, this is old detritus and doesn't need to be there.

PR #1881, opened 2021-09-24
document support for rhel7-based tfe custom worker images

p0pr0ck5:


sudomateo (approved):

content/source/docs/enterprise/install/installer.html.md
```diff
@@ -159,12 +159,15 @@ TFE runs `terraform plan` and `terraform apply` operations in a disposable Docke
 ![Terraform Enterprise docker image](./assets/tfe-docker-image.png)
 
 ### Requirements
- - The base image must be `ubuntu:bionic`.
+ - The base image must be `ubuntu:bionic` or an [offical RHEL7 image](https://catalog.redhat.com/software/containers/rhel7/57ea8cee9c624c035f96f3af) (e.g., `registry.access.redhat.com/rhel7`).
  - The image must exist on the Terraform Enterprise host. It can be added by running `docker pull` from a local registry or any other similar method.
+ - The software packages defined in the below exampes must be installed on the image.
```
laurapacilio, line 164: You know, so there's no confusion. Could we just list them here? Say something like "These are..." and list the packages they need?

```suggestion
 - The software packages defined in the examples below must be installed on the image.
```

content/source/docs/enterprise/install/installer.html.md
```diff
@@ -182,6 +185,33 @@ ADD example-intermediate-ca.crt /usr/local/share/ca-certificates/
...
+
+# Install software used by Terraform Enterprise.
+RUN yum -y install unzip sudo git openssh wget curl psmisc iproute nmap-ncat openssl rpm-build redhat-rpm-config make gcc && \
+  yum clean all && \
+  cd /tmp && git clone https://github.com/jacobm3/daemontools-rpm.git && \
+  cd /tmp/daemontools-rpm && wget https://cr.yp.to/daemontools/daemontools-0.76.tar.gz && ./buildrpm.sh && \
+  rpm -ivh /root/rpmbuild/RPMS/x86_64/daemontools-0.76-1.el7.x86_64.rpm
+```
+
+Note that the `daemontools` package is required, as it provides the `envdir` binary, which is used by TFE. If this package is not
```
laurapacilio, line 212: This implies to me that some of the things in here are not required? Should we tell folks which things they need and which ones they don't?  Regardless, I think I'd move this to be above the actual script, right under the heading. I worry that folks won't see it way down here :-)

laurapacilio (changes requested):
Looking good - left you a few comments to address!

content/source/docs/enterprise/install/installer.html.md
p0pr0ck5: I'll move this up and reword it a bit. There is another section under the requirements that says the installed packages listed in these examples are required.

content/source/docs/enterprise/install/installer.html.md
p0pr0ck5: Since the packages are different between RHEL and Ubuntu (and there are quite a few), listing them out here felt kind of awkward. Not sure what the best approach is here?

laurapacilio: okay I can accept that - I'll resolve this

content/source/docs/enterprise/install/installer.html.md
```diff
@@ -182,6 +185,33 @@ ADD example-intermediate-ca.crt /usr/local/share/ca-certificates/
...
+RUN subscription-manager register --username ${RH_username} --password "${RH_password}" --auto-attach
+
+# Include all necessary CA certificates.
+ADD example-root-ca.crt /usr/share/pki/ca-trust-source/anchors
+ADD example-intermediate-ca.crt /usr/share/pki/ca-trust-source/anchors
+
+# Update the CA certificates bundle to include newly added CA certificates.
+RUN update-ca-trust
+
+# Install software used by Terraform Enterprise.
```
laurapacilio, line 207: The passive voice bothered me, so I tried rewording :-) Can you change the one in Ubuntu as well?

```suggestion
# Install required software for Terraform Enterprise.
```

laurapacilio (approved):
One nit for passive voice - I'll approve so I don't block, but if you could fix that it would make me happy :-).  Thank you!

content/source/docs/enterprise/install/installer.html.md
p0pr0ck5: No problem! Done :+1:

PR #1886, opened 2021-10-01
Document version constraints in workspace TF version; remove "latest"

nfagerlund:
**Draft status: This isn't shipped yet.**

A longstanding secret feature of TFC is that you can specify a version
constraint as the Terraform version for a workspace, if you're using the API or
provider.

A longstanding documented-but-discouraged feature is that you can specify the
special keyword "latest" and have it always use the latest version.

We're getting rid of the special "latest" keyword, but in exchange, we're
documenting the version constraints feature and adding the most commonly useful
version constraints (`~>` for the last three minor versions) to the select menu
in the workspace settings.


nfagerlund:
Oh NO, that SSL verify error in the broken link checker 😱 HATE IT

Well, those are false alarms.

PR #1889, opened 2021-10-04
Being explicit about changes in changelog.

glenngillen:
Customer pointed out that linking to external docs to document what things had changed meant the value of the changelog was diminished in a historical context. For example, if you're 12mo behind on upgrading TFE and taking an incremental approach to catching up it's not clear which part of "updated Module Delete APIs" was updated on this specific date.

So I've gone back and made the relevant bit more explicit to avoid any confusion.


bethanyr (approved):
these changes make sense - and make it clearer which specific endpoint should be used instead of the deprecated endpoint.

dylanegan (approved):

PR #1890, opened 2021-10-04
Update automating-the-installer.html.md

susiesolis:
Adding more to the example in iact subnet list. Customer tripped over this when adding more than one subnet and the issue was a space between the subnets so updating with an example for multiple subnets.

## Labels

- [ ] inaccuracy
- [x] clarification
- [ ] new docs


laurapacilio (approved):
Thank you! :-)

PR #1938, opened 2021-11-02
Document `tfe-admin list-nodes` command

sudomateo:
Documenting the `tfe-admin list-nodes` command.

### DIFF
--- a/content/source/docs/enterprise/admin/active-active.html.md
+++ b/content/source/docs/enterprise/admin/active-active.html.md
@@ -75,9 +75,7 @@ tfe-app-config ()
 tfe-admin list-nodes
 ```
 
-This command will list the IP addresses of all active nodes in the active/active installation. Nodes send a heartbeat
-every 5 seconds to signal that they are active. If a heartbeat is not received within 30 seconds, the node is considered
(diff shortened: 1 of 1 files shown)


chrisfishwood:
:shipit:

content/source/docs/enterprise/admin/active-active.html.md
```diff
@@ -69,6 +69,16 @@ tfe-app-config ()
...
 
+#### list-nodes
+
+```bash
+tfe-admin list-nodes
+```
+
+This command will list the IP addresses of all active nodes in the active/active installation. Nodes send a heartbeat
+every 5 seconds to signal that they are active. If a heartbeat is not received within 30 seconds, the node is considered
+inactive and removed from this list.
```
laurapacilio, line 80:
```suggestion
This command lists the IP addresses of all active nodes in the active/active installation. Nodes send a heartbeat every 5 seconds to signal that they are active. If Terraform does not receive a heartbeat within 30 seconds, it considers the node inactive and removes the node from this list.
```

laurapacilio (approved):
Looks good :-) Minor suggestion from me.

PR #1945, opened 2021-11-05
Update docs sidebar and dropdown

laurapacilio:
Recently, we added the [Terraform Integration Program](https://www.terraform.io/guides/terraform-integration-program.html#terraform-cloud-integrations) page to the documentation site. It helps our integration partners understand what types of integrations they can create and how to work with HashiCorp to get their integrations verified.

Currently, the integration program is listed at the highest level of the documentation, meaning that it is specifically linked in the documentation dropdown menu and the main sidebar. The addition of this page made the dropdown very long and also harder to parse because there are so many items that start with "Terraform." We understand that these are official programs and products (with Terraform being a part of the official name), but we think that in these specific cases, users will be scanning for the keywords "integration" and "registry." Removing "Terraform" from in front of these two items makes it easier for users to see those keywords and also makes these menu items fit on one line, making the dropdown and sidebar shorter and less overwhelming. We will absolutely refer to the integration program and the Terraform Registry by their official names on their respective pages, but we think that this change is necessary to improver user experience. 

We also made one small change to the  plugin sidebar - previously the listing said "Partner  Integration Program" but the official program name is "Terraform Integration Program", so the sidebar listing has been updated to reflect this.

Here is a before and after shot of the dropdown menu, demonstrating the impact of this change:

**Before**
before

**After**
after

blocked by: https://github.com/hashicorp/terraform-website-next/pull/101

### DIFF
--- a/content/source/layouts/_otherdocs.erb
+++ b/content/source/layouts/_otherdocs.erb
@@ -10,8 +10,8 @@
     "Terraform Enterprise" => "/docs/enterprise/index.html",
     "Provider Use" => "/docs/language/providers/index.html",
     "Plugin Development" => "/docs/extend/index.html",
-    "Terraform Registry Publishing" => "/docs/registry/index.html",
-    "Terraform Integration Program" => "/guides/terraform-integration-program.html",
+    "Registry Publishing" => "/docs/registry/index.html",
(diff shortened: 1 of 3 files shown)


cgriggs01 (approved):

laurenrother (approved):
👍

Issue #1940, opened 2021-11-03
Claim that Sentinel mocks can only be downloaded for 7 days no longer seems true


rberlind:
The claim in https://www.terraform.io/docs/cloud/sentinel/mock.html#mock-data-availability that Sentinel mocks can only be downloaded for 7 days no longer seems true.  I have downloaded mocks that are 3 months old.  I do know that the 7 day limit used to be enforced.  But at some point in the past year, things changed.

What I don't know is whether the doc needs to be updated or whether the TFC engineering team introduced a bug that prevents the 7 day limit (if still desired by them) from being enforced properly.

In any case, this should be investigated.  If the limitation to 7 days was intentionally changed, we need to find out whether it was removed entirely or changed to some longer limit.  If the latter, we need to find out that limit and indicate it in the doc. Otherwise, we should remove mention of the limit.

dylanegan:
@rberlind I believe in the past year, somebody indeed made a change unrelated to mocks that did have an effect on their availability. In this case the documentation could be updated to remove that clause. Do you want to follow up on that or shall I?

rberlind:
Would you mind doing it @dylanegan ?

PR #1947, opened 2021-11-07
Sentinel mocks are available indefinitely

dylanegan:
Prior to July 23, 2021, the JSON plan output was deleted 7 days after it was generated. This limited Sentinel mock generation to that time period as it relies on the JSON plan output. They are now available longer than 7 days, which makes mock generation available for longer too.

## Labels

- [x] inaccuracy
- [ ] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/cloud/sentinel/mock.html.md
+++ b/content/source/docs/cloud/sentinel/mock.html.md
@@ -276,5 +276,3 @@ If a plan cannot have its mock data exported due to any of these reasons, the
 not be visible.
 
 -> **Note:** Only a successful plan is required for mock generation. Sentinel can still generate the data if apply or policy checks fail.
-mock generation - if the apply or the policy checks fail, the data can still be
-generated.
--- a/content/source/docs/cloud/sentinel/mock.html.md
+++ b/content/source/docs/cloud/sentinel/mock.html.md
@@ -267,7 +267,7 @@ The following factors can prevent you from generating mock data:
   produced via mock generation.
 * The run has not progressed past the planning stage, or did not create a plan
   successfully.
-* The run had progressed past the planning stage prior to July 23, 2021. Prior to this date, JSON plans were only kept for 7 days.
+* The run progressed past the planning stage prior to July 23, 2021. Prior to this date, Terraform Cloud only kept JSON plans for 7 days.
 
 [permissions-citation]: #intentionally-unused---keep-for-maintainers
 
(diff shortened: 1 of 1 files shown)


dylanegan:
This was reported in #1940.

hcrhall (approved):
LGTM

rberlind (commented):
I think it would be better to remove the third bullet altogether.  What we have ended up with will no longer apply to current runs and is awkwardly phrased: "The run had progressed past the planning stage prior to July 23, 2021. Prior to this date, JSON plans were only kept for 7 days." Why call attention to something that is no longer true?

dylanegan:
@rberlind it is true though and a potential support burden should folks be unsure why a mock isn't generating. Otherwise we do not clearly state the difference anywhere.

rberlind:
What you said does make sense, @dylanegan to cover the case of people trying to generate mocks from older runs.

content/source/docs/cloud/sentinel/mock.html.md
```diff
@@ -279,4 +277,4 @@ not be visible.
 
 -> **Note:** Only the plan needs to be successful for a run to be eligible for
```
laurapacilio, line 278: Can we make it more active? Something like this: 

```suggestion
-> **Note:** Only a successful plan is required for mock generation. Sentinel can still generate the data if apply or policy checks fail.
```

content/source/docs/cloud/sentinel/mock.html.md
```diff
@@ -267,9 +267,7 @@ The following factors can prevent you from generating mock data:
   produced via mock generation.
 * The run has not progressed past the planning stage, or did not create a plan
   successfully.
-* The run has been in a terminal state, such as applied or discarded, longer
-  than seven days. At this point, the data necessary to generate the mocks is no
-  longer available.
+* The run had progressed past the planning stage prior to July 23, 2021. Prior to this date, JSON plans were only kept for 7 days.
```
laurapacilio, line 270: Again, just want it to be active voice :-)

```suggestion
* The run progressed past the planning stage prior to July 23, 2021. Prior to this date, Terraform Cloud only kept JSON plans for 7 days.
```

laurapacilio (approved):
A couple suggestions to make it flow better - thank you for fixing this and to @rberlind for catching it!

content/source/docs/cloud/sentinel/mock.html.md
rberlind: I would rather see us use this:
```
The run progressed past the planning stage prior to July 23, 2021 and more than seven days had passed. (After that date, Terraform Cloud began permanently keeping the JSON plans that mock generation requires.)
```

laurapacilio: I like your edit except - I think let's omit the parenthesis and say "the JSON plans required for mock generation". If we do that, I'm sold!

PR #1953, opened 2021-11-11
Tf api/pagination 2

anamahalo:
Adding in _Workspace Resources_ due to pagination, but also, it was never added!  So not a new feature.

- [x] Note to self: Attach screenshot once confirmations are 👍 ✔️   <-- didn't get all the confirmations but we can open this up and just complete conversation here
- [x] Also apply this change if necessary for consistency: https://github.com/hashicorp/terraform-website/pull/1974

## Labels

API is checked but it isn't a new endpoint, just has never been documented!

- [ ] inaccuracy
- [ ] clarification
- [x] new docs

CleanShot 2021-11-30 at 13 33 04

### DIFF
--- a/content/source/docs/cloud/api/workspace-resources.html.md
+++ b/content/source/docs/cloud/api/workspace-resources.html.md
@@ -1,6 +1,6 @@
 ---
 layout: "cloud"
-page_title: "Workspaces - API Docs - Terraform Cloud and Terraform Enterprise"
+page_title: "Workspace Resources - API Docs - Terraform Cloud and Terraform Enterprise"
 ---
 
 [200]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200
--- a/content/source/docs/cloud/api/workspace-resources.html.md
+++ b/content/source/docs/cloud/api/workspace-resources.html.md
@@ -46,7 +46,7 @@ Parameter                   | Description
 
 ### Permissions
 
-In order to list resources, the user must have permission to read resources for the specified workspace. ([More about permissions.](/docs/cloud/users-teams-organizations/permissions.html))
+To list resources, the user must have permission to read resources for the specified workspace. ([More about permissions.](/docs/cloud/users-teams-organizations/permissions.html))
 
 [permissions-citation]: #intentionally-unused---keep-for-maintainers
 
--- a/content/source/docs/cloud/api/workspace-resources.html.md
+++ b/content/source/docs/cloud/api/workspace-resources.html.md
@@ -1,0 +1,117 @@
+---
+layout: "cloud"
+page_title: "Workspaces - API Docs - Terraform Cloud and Terraform Enterprise"
+---
+
+[200]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200
+[201]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201
+[202]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202
+[204]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204
+[400]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400
+[401]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401
+[403]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403
+[404]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404
+[409]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409
+[412]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/412
+[422]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422
+[429]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429
+[500]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500
+[504]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504
+[JSON API document]: /docs/cloud/api/index.html#json-api-documents
+[JSON API error object]: https://jsonapi.org/format/#error-objects
+
+# Workspace Resources API
+
+## List Workspace Resources
+
+`GET /workspaces/:workspace_id/resources`
+
+Parameter            | Description
+---------------------|------------
+`:workspace_id`      | The ID of the workspace to retrieve resources from. Obtain this from the [workspace settings](../workspaces/settings.html) or the [Show Workspace](./workspaces.html#show-workspace) endpoint.
+
+Status  | Response                                       | Reason
+--------|------------------------------------------------|----------
+[200][] | [JSON API document][] (`type: "resources"`)    | Request was successful
+[404][] | [JSON API error object][]                      | Workspace not found or user unauthorized to perform action
+
+### Query Parameters
+
+This endpoint supports pagination [with standard URL query parameters](./index.html#query-parameters); remember to percent-encode `[` as `%5B` and `]` as `%5D` if your tooling doesn't automatically encode URLs.
+
+Parameter                   | Description
+----------------------------|------------
+`page[number]`              | **Optional.** If omitted, the endpoint will return the first page.
+`page[size]`                | **Optional.** If omitted, the endpoint will return 20 workspace resources per page.
+
+### Permissions
+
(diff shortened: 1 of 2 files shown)


content/source/layouts/_cloud_content.erb
```diff
@@ -435,6 +435,11 @@
...
           <li>
             <a href="/docs/cloud/api/workspaces.html">Workspaces</a>
+            <ul class="nav">
+              <li>
+                <a href="/docs/cloud/api/workspace-resources.html">Workspace Resources</a>
+              </li>
+            </ul>
           </li>
           <li>
             <a href="/docs/cloud/api/workspace-variables.html">Workspace Variables</a>
```
anamahalo, line 445: Depending on how conversation goes am happy to move this so it's subheaded under `Workspaces` as well for consistency sake.  Wanted to leave the diversion for now for differences, but also upcoming POCs conversation to be had.

anamahalo: Just updating here so anyone coming thru is on the same page also — after speaking with @laurapacilio gonna hold off on reorganizing atm due to current other documentation work.  That said, for the purposes of _Workspace Resources_ it will also not be nested since it sounds like it’ll be easier to start the nesting after said work is completed, so will also persist _Variable Sets_ and _Workspace Specific Variables_ as-is also.  :thumbsup:

content/source/docs/cloud/api/workspace-resources.html.md
```diff
@@ -0,0 +1,108 @@
...
+    }
+  ],
+  "links": {
+    "self": "https://app.terraform.io/api/v2/workspaces/ws-DiTzUDRpjrArAfSS/resources?page%5Bnumber%5D=1&page%5Bsize%5D=20",
+    "first": "https://app.terraform.io/api/v2/workspaces/ws-DiTzUDRpjrArAfSS/resources?page%5Bnumber%5D=1&page%5Bsize%5D=20",
+    "prev": null,
+    "next": null,
+    "last": "https://app.terraform.io/api/v2/workspaces/ws-DiTzUDRpjrArAfSS/resources?page%5Bnumber%5D=1&page%5Bsize%5D=20"
+  },
+  ...
```
anamahalo, line 106: We OK with the ellipses here?  Otherwise it gets very long.  (This style is done elsewhere also iirc)

content/source/docs/cloud/api/workspace-resources.html.md
```diff
@@ -0,0 +1,108 @@
...
+
+```json
+{
+  "data": [
+    {
+      "id": "wsr-KNYb3Jj3JTBgoBFs",
+      "type": "resources",
+      "attributes": {
+        "address": "random_pet.animal",
+        "name": "animal",
```
anamahalo, line 73: Do we want to keep `random_pet`?  If not, any suggested output for the things without having to queue up actual resources?

content/source/docs/cloud/api/workspace-resources.html.md
```diff
@@ -0,0 +1,108 @@
+---
+layout: "cloud"
+page_title: "Workspaces - API Docs - Terraform Cloud and Terraform Enterprise"
```
dylanegan, line 3:
```suggestion
page_title: "Workspace Resources - API Docs - Terraform Cloud and Terraform Enterprise"
```

content/source/docs/cloud/api/workspace-resources.html.md
```diff
@@ -0,0 +1,108 @@
...
+
+# Workspace Resources API
+
+## List Workspace Resources
+
+`GET /workspaces/:workspace_id/resources`
+
+Parameter            | Description
+---------------------|------------
+`:workspace_id`      | The ID of the workspace to retrieve resources from. Obtain this from the [workspace settings](../workspaces/settings.html) or the [Show Workspace](./workspaces.html#show-workspace) endpoint.
```
dylanegan, line 31:
```suggestion
`:workspace_id`      | The ID of the workspace to retrieve resources from. Obtain this from the [workspace settings](../workspaces/settings.html) or the [show workspace](./workspaces.html#show-workspace) endpoint.
```

I'm unsure of preference here, so will defer to @laurapacilio, but with `workspace settings` lowercased, do we want the same with the other link?

content/source/docs/cloud/api/workspace-resources.html.md
dylanegan: I feel we should be good, otherwise it'll be lengthy, as you point out.

content/source/docs/cloud/api/workspace-resources.html.md
laurapacilio: I agree with lower case :)  I think we don't have a reason to capitalize this here.

content/source/docs/cloud/api/workspace-resources.html.md
```diff
@@ -0,0 +1,108 @@
...
+
+Status  | Response                                       | Reason
+--------|------------------------------------------------|----------
+[200][] | [JSON API document][] (`type: "resources"`)    | Request was successful
+[404][] | [JSON API error object][]                      | Workspace not found or user unauthorized to perform action
```
laurapacilio, line 36: Let's add periods at the end of the reasons for consistency. I don't love doing it after sentence fragments, BUT we do it in ever other table it seems. :-)

content/source/docs/cloud/api/workspace-resources.html.md
```diff
@@ -0,0 +1,108 @@
...
+This endpoint supports pagination [with standard URL query parameters](./index.html#query-parameters). Remember to percent-encode `[` as `%5B` and `]` as `%5D` if your tooling doesn't automatically encode URLs.
+
+Parameter                   | Description
+----------------------------|------------
+`page[number]`              | **Optional.** If omitted, the endpoint will return the first page.
+`page[size]`                | **Optional.** If omitted, the endpoint will return 20 workspace resources per page.
+
+### Permissions
+
+In order to list resources, the user must have permission to read resources for the specified workspace. ([More about permissions.](/docs/cloud/users-teams-organizations/permissions.html))
```
laurapacilio, line 49:
```suggestion
To list resources, the user must have permission to read resources for the specified workspace. ([More about permissions.](/docs/cloud/users-teams-organizations/permissions.html))
```

laurapacilio (approved):
A couple of nitpicks - but otherwise good. Thank you!!

content/source/docs/cloud/api/workspace-resources.html.md
anamahalo: Thanks @dylanegan && @laurapacilio !

content/source/docs/cloud/api/workspace-resources.html.md
anamahalo: Oh thanks!  Forgot to change that after reverting it back to no longer being nested.  Good 👁️

content/source/docs/cloud/api/workspace-resources.html.md
anamahalo: Oh I didn't realize I made this different here.  Thank you!  I would have actually argued the opposite (capitalize _Workspace Settings_ due to the proper naming with the _the_ but I think as long as we're consistent it's good 👍

content/source/docs/cloud/api/workspace-resources.html.md
anamahalo: Going to take this suggestion and then get rid of the comma splice 😄

content/source/docs/cloud/api/workspace-resources.html.md
anamahalo: I think the sentence fragment was the reason why I didn't do it haha.  SGTM though!

content/source/docs/cloud/api/workspace-resources.html.md
anamahalo: @laurapacilio - just double checking, does the ✅ mean we're OK w/this being here?  We use that publicly for tutorial purposes I think so in that way seems useful to customers but also not sure if we want to replace it with something else, in which case, my query above still stands.  😄  cc @dylanegan in case for suggestions, though we can also speak during our sync tomorrow too 👍

dylanegan: I'm happy with `random_pet`.

laurapacilio: Yes! This feels good to me - sorry I should have explicitly said :). We use random_pet throughout in tutorials, et al as you mention so I think this is fine. As long as our users are getting the full picture of what gets returned by this endpoint, I feel satisfied!

anamahalo: Thank you!

PR #1962, opened 2021-11-15
Remove heading

laurapacilio:
Removes a heading that shouldn't be there from the variable sets API docs.


rexredinger (approved):
:shipit:

PR #1967, opened 2021-11-17
docs/plugin/framework: Initial validation page

bflad:
Closes hashicorp/terraform-plugin-framework#144

## Labels

- [ ] inaccuracy
- [ ] clarification
- [x] new docs


content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,257 @@
...
+
+~> **NOTE:** When implementing validation logic, configuration values may be [unknown](./types.html#unknown) based on the source of the value. Implementations must account for this case, typically by returning early without returning new diagnostics.
+
+## Syntax and Basic Schema Validation
+
+The [Terraform configuration language](https://www.terraform.io/docs/language/) is declarative and an implementation of [HashiCorp Configuration Language](https://github.com/hashicorp/hcl) (HCL). Terraform CLI is responsible for reading and parsing configurations for validity, based on Terraform's concepts such as `resource` blocks and associated syntax. Basic validation of value type and behavior information, for example returning an error when a string value is given where a list value is expected or returning an error when a required attribute is missing from a configuration, is automatically handled by Terraform CLI based on the provider, resource, or data source schema.
```
paddycarver, line 16:
```suggestion
The [Terraform configuration language](https://www.terraform.io/docs/language/) is declarative and an implementation of [HashiCorp Configuration Language](https://github.com/hashicorp/hcl) (HCL). The Terraform CLI is responsible for reading and parsing configurations for validity, based on Terraform's concepts such as `resource` blocks and associated syntax. Basic validation of value type and behavior information, for example returning an error when a string value is given where a list value is expected or returning an error when a required attribute is missing from a configuration, is automatically handled by the Terraform CLI based on the provider, resource, or data source schema.
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,257 @@
...
+
+Any further validation provided by the framework occurs after these checks.
+
+## Attribute Validation
+
+It is common for provider implementations to introduce validation on attributes using the generic framework-defined types such as `types.String`. The [`tfsdk.Attribute` type `Validators` field](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#Attribute.Validators) can be supplied with a list of validations and diagnostics will be returned from all validators. For example:
```
paddycarver, line 22:
```suggestion
It is common for provider implementations to introduce validation on attributes using the generic framework-defined types such as [`types.String`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/types#String). The [`tfsdk.Attribute` type `Validators` field](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#Attribute.Validators) can be supplied with a list of validations and diagnostics will be returned from all validators. For example:
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,257 @@
...
+    // ... other Attribute configuration ...
+
+    Validators: []AttributeValidators{
+        // These are example validators
+        stringLengthBetween(10, 256),
+        stringRegularExpression(regexp.MustCompile(`^[a-z0-9]+$`)),
+    },
+}
+```
+
```
paddycarver, line 37:
```suggestion

All validators will always be run, regardless of whether previous validators returned an error or not.

```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,257 @@
...
+### Creating Attribute Validators
+
+To create an attribute validator, the [`tfsdk.AttributeValidator` interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#AttributeValidator) must be satisfied. For example:
+
+```go
+type stringLengthBetweenValidator struct {
+    Max int
+    Min int
+}
+
```
paddycarver, line 49:
```suggestion

// Description returns a plain text description of the validator's behavior, suitable for a practitioner to understand its impact.
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,257 @@
...
+func (v stringLengthBetweenValidator) Description(ctx context.Context) string {
+    return fmt.Sprintf("string length must be between %d and %d", v.Min, v.Max)
+}
+
```
paddycarver, line 53:
```suggestion

// MarkdownDescription returns a markdown formatted description of the validator's behavior, suitable for a practitioner to understand its impact.
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,257 @@
...
+func (v stringLengthBetweenValidator) MarkdownDescription(ctx context.Context) string {
+    return fmt.Sprintf("string length must be between `%d` and `%d`", v.Min, v.Max)
+}
+
```
paddycarver, line 57:
```suggestion

// Validate runs the main validation logic of the validator, reading configuration data out of `req` and updating `resp` with diagnostics.
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,257 @@
...
+    if !ok {
+        response.Diagnostics.AddAttributeError(
+            request.AttributePath,
+            "Invalid Attribute Value Type",
+            fmt.Sprintf("Expected types.String value, got: %T", request.AttributeConfig),
+        )
+
+        return
+    }
```
paddycarver, line 69:
```suggestion
    // types.String must be the attr.Value produced by the attr.Type in the schema for this attribute
    // for generic validators, use
    // https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#ConvertValue
    // to convert into a known type.
    var str types.String
    diags := tfsdk.ValueAs(ctx, val, &target)
    resp.Diagnostics.Append(diags...)
    if resp.Diagnostics.HasError() {
        return
    }
```

This explains that we can only use the `attr.Value` produced by the `attr.Type` in the schema, tells people how to convert between `attr.Type`s, and uses the `tfsdk.ValueAs` instead of type assertion to get the value.

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,257 @@
...
+
+    strLen := len(str)
+
+    if strLen < v.Min || strLen > v.Max {
+        response.Diagnostics.AddAttributeError(
+            request.AttributePath,
+            "Attribute Validation Error",
+            fmt.Sprintf("%s, got: %d", v.Description(ctx), strLen),
```
paddycarver, line 81: Does this confuse the matter of what `Description` is for? Do we want to de-duplicate it to make it clear the point of Description is for documentation and not to be part of the validator?

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,257 @@
...
+    if str.Unknown || str.Null {
+        return
+    }
+
+    strLen := len(str)
+
+    if strLen < v.Min || strLen > v.Max {
+        response.Diagnostics.AddAttributeError(
+            request.AttributePath,
+            "Attribute Validation Error",
```
paddycarver, line 80: Can we be more descriptive here, to prompt provider devs to write better errors?

```suggestion
            "Invalid String Length",
```

paddycarver (commented):
Looks like a great guide to me, @bflad. I left some suggestions, but super interested what @laurapacilio thinks of it, too :)

laurapacilio (commented):
Hooray - new docs! :-) 

Brian - once you address Paddy's feedback, I'll take a thorough pass through. I do have some thoughts on language - but I see Paddy made a few suggestions and so I want to let you resolve those first. 

A couple overall thoughts: 
-  I like how you've added call outs to this documentation on other pages, but I wonder if we can be more explicit about what we're pointing folks to. We have "Further Data Source Capabilities" and I wonder if instead we change that to "Add Validation" or something similar? Would match the way we've done other section titles and immediately let folks know what this section is giving them. Also - I think we want like 1-2 sentence of explanation and then a link. So something like: "You can include [validation logic](/docs/plugin/framework/validation.html) that helps practitioners understand the required syntax, types, and acceptable values for your provider."

- I think we should say when practitioners are going to encounter this validation feedback and see the errors. Is it when they run `terraform fmt` or when they run a plan or apply?  I know this is probably something we're assuming folks know if they've used terraform, but I think it's best to be explicit. I might do this in the very beginning of your doc, right when you're talking about the syntax and basic schema validation.

content/source/docs/plugin/framework/validation.html.md
bflad: Love it, just needs minor variable name fixes.

```suggestion
    // types.String must be the attr.Value produced by the attr.Type in the schema for this attribute
    // for generic validators, use
    // https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#ConvertValue
    // to convert into a known type.
    var str types.String
    diags := tfsdk.ValueAs(ctx, req.AttributeConfig, &str)
    resp.Diagnostics.Append(diags...)
    if diags.HasError() {
        return
    }
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+
+    if str.Unknown || str.Null {
+        return
+    }
+
+    strLen := len(str)
+
+    if strLen < v.Min || strLen > v.Max {
+        response.Diagnostics.AddAttributeError(
+            request.AttributePath,
```
bflad, line 83:
```suggestion
        resp.Diagnostics.AddAttributeError(
            req.AttributePath,
```

content/source/docs/plugin/framework/validation.html.md
bflad: Let's "duplicate" for now to reduce any confusion (and correctly do more of sentence structure).

```suggestion
            fmt.Sprintf("String length must be between %d and %d, got: %d.", v.Min, v.Max, strLen),
```

bflad:
Hey @laurapacilio 👋  I've addressed @paddycarver's feedback so you should be good to do a pass now.

I was hoping that the new bottom sections like `Further Resource Capabilities` would be expanded with links to other framework functionality that is specific to each concept, such as some ongoing new documentation on [plan modifiers](https://github.com/hashicorp/terraform-plugin-framework/issues/145) and future documentation on [resource import](https://github.com/hashicorp/terraform-plugin-framework/issues/155). I do really like the idea of expanding those to include some context, which I'll push up some quick adjustments here shortly.

I'll also try to be a little more specific in the lead-in information about validation, although these types of validation errors can occur across any of the following commands: `apply`, `plan`, and `validate`.

content/source/docs/plugin/framework/data-sources.html.md
```diff
@@ -95,3 +95,7 @@ func (p *provider) GetDataSources(_ context.Context) (map[string]tfsdk.DataSourc
 	}, nil
 }
 ```
+
+## Further Data Source Capabilities
+
+- [Validation](./validation.html)
```
bflad, line 101:
```suggestion
- [Validation](./validation.html) helps practitioners understand the required syntax, types, and acceptable values for your data source.
```

content/source/docs/plugin/framework/providers.html.md
```diff
@@ -110,3 +110,7 @@ in the configuration, including the provider prefix.
 
 The list of data sources is meant to be immutable. It should not change at
 runtime, and should consistently return the same values.
+
+## Further Provider Capabilities
+
+- [Validation](./validation.html)
```
bflad, line 116:
```suggestion
- [Validation](./validation.html) helps practitioners understand the required syntax, types, and acceptable values for your provider.
```

content/source/docs/plugin/framework/resources.html.md
```diff
@@ -159,3 +159,7 @@ func (p *provider) GetResources(_ context.Context) (map[string]tfsdk.ResourceTyp
 	}, nil
 }
 ```
+
+## Further Resource Capabilities
+
+- [Validation](./validation.html)
```
bflad, line 165:
```suggestion
- [Validation](./validation.html) helps practitioners understand the required syntax, types, and acceptable values for your resource.
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+
+~> **NOTE:** When implementing validation logic, configuration values may be [unknown](./types.html#unknown) based on the source of the value. Implementations must account for this case, typically by returning early without returning new diagnostics.
+
+## Syntax and Basic Schema Validation
+
+The [Terraform configuration language](https://www.terraform.io/docs/language/) is declarative and an implementation of [HashiCorp Configuration Language](https://github.com/hashicorp/hcl) (HCL). The Terraform CLI is responsible for reading and parsing configurations for validity, based on Terraform's concepts such as `resource` blocks and associated syntax. Basic validation of value type and behavior information, for example returning an error when a string value is given where a list value is expected or returning an error when a required attribute is missing from a configuration, is automatically handled by the Terraform CLI based on the provider, resource, or data source schema.
+
+Any further validation provided by the framework occurs after these checks.
```
bflad, line 18:
```suggestion
These syntax and basic schema checks are performed during the [`terraform apply`](https://www.terraform.io/docs/cli/commands/apply.html), [`terraform destroy`](https://www.terraform.io/docs/cli/commands/destroy.html), [`terraform plan`](https://www.terraform.io/docs/cli/commands/plan.html), and [`terraform validate`](https://www.terraform.io/docs/cli/commands/validate.html) commands. Any further validation provided by the framework occurs after these checks.
```

paddycarver (approved):
Seems reasonable to me!

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+page_title: "Plugin Development - Framework: Validation"
+description: |-
+  How to validate configuration values using the provider development framework.
+---
+
+# Validation
+
+Practitioners implementing Terraform configurations desire feedback surrounding the syntax, types, and acceptable values. This feedback, typically referred to as validation, is preferably given as early as possible before a configuration is applied. Terraform supports validation for values in provider, resource, and data source configurations. The framework supports returning [diagnostics](./diagnostics.html) feedback on all of these, offering the ability to specify validations on an attribute, provider-defined type, and the entire provider, resource, or data source schema.
+
+~> **NOTE:** When implementing validation logic, configuration values may be [unknown](./types.html#unknown) based on the source of the value. Implementations must account for this case, typically by returning early without returning new diagnostics.
```
laurapacilio, line 12: Can we move this note down to where folks will be reading about implementation? I just want them to see it in context - I worry they'll forget about it if it stays up here :)

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
+---
+layout: "extend"
+page_title: "Plugin Development - Framework: Validation"
+description: |-
+  How to validate configuration values using the provider development framework.
+---
+
+# Validation
+
+Practitioners implementing Terraform configurations desire feedback surrounding the syntax, types, and acceptable values. This feedback, typically referred to as validation, is preferably given as early as possible before a configuration is applied. Terraform supports validation for values in provider, resource, and data source configurations. The framework supports returning [diagnostics](./diagnostics.html) feedback on all of these, offering the ability to specify validations on an attribute, provider-defined type, and the entire provider, resource, or data source schema.
```
laurapacilio, line 10: Just thinking about how we can make this as concise as possible :) 

```suggestion
The framework can return [diagnostics](./diagnostics.html) feedback for values in provider, resource, and data source configurations. This allows you to write validations that give users feedback about required syntax, types, and acceptable values.
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+
+~> **NOTE:** When implementing validation logic, configuration values may be [unknown](./types.html#unknown) based on the source of the value. Implementations must account for this case, typically by returning early without returning new diagnostics.
+
+## Syntax and Basic Schema Validation
+
+The [Terraform configuration language](https://www.terraform.io/docs/language/) is declarative and an implementation of [HashiCorp Configuration Language](https://github.com/hashicorp/hcl) (HCL). The Terraform CLI is responsible for reading and parsing configurations for validity, based on Terraform's concepts such as `resource` blocks and associated syntax. Basic validation of value type and behavior information, for example returning an error when a string value is given where a list value is expected or returning an error when a required attribute is missing from a configuration, is automatically handled by the Terraform CLI based on the provider, resource, or data source schema.
```
laurapacilio, line 16:
```suggestion
The [Terraform configuration language](https://www.terraform.io/docs/language/) is declarative and an implementation of [HashiCorp Configuration Language](https://github.com/hashicorp/hcl) (HCL). The Terraform CLI is responsible for reading and parsing configurations for validity, based on Terraform's concepts such as `resource` blocks and associated syntax. The Terraform CLI automatically handles basic validation of value type and behavior information based on the provider, resource, or data source schema. For example, the Terraform CLI returns an error when a string value is given where a list value is expected and also when a required attribute is missing from a configuration.
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+
+These syntax and basic schema checks are performed during the [`terraform apply`](https://www.terraform.io/docs/cli/commands/apply.html), [`terraform destroy`](https://www.terraform.io/docs/cli/commands/destroy.html), [`terraform plan`](https://www.terraform.io/docs/cli/commands/plan.html), and [`terraform validate`](https://www.terraform.io/docs/cli/commands/validate.html) commands. Any further validation provided by the framework occurs after these checks.
```
laurapacilio, line 18:
```suggestion
Terraform CLI syntax and basic schema checks occur during the [`terraform apply`](https://www.terraform.io/docs/cli/commands/apply.html), [`terraform destroy`](https://www.terraform.io/docs/cli/commands/destroy.html), [`terraform plan`](https://www.terraform.io/docs/cli/commands/plan.html), and [`terraform validate`](https://www.terraform.io/docs/cli/commands/validate.html) commands. Any additional validation you define with the framework occurs directly after these checks are complete.
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+  How to validate configuration values using the provider development framework.
+---
+
+# Validation
+
+Practitioners implementing Terraform configurations desire feedback surrounding the syntax, types, and acceptable values. This feedback, typically referred to as validation, is preferably given as early as possible before a configuration is applied. Terraform supports validation for values in provider, resource, and data source configurations. The framework supports returning [diagnostics](./diagnostics.html) feedback on all of these, offering the ability to specify validations on an attribute, provider-defined type, and the entire provider, resource, or data source schema.
+
+~> **NOTE:** When implementing validation logic, configuration values may be [unknown](./types.html#unknown) based on the source of the value. Implementations must account for this case, typically by returning early without returning new diagnostics.
+
+## Syntax and Basic Schema Validation
```
laurapacilio, line 14: I want this heading to make the purpose clear - this is about when the checks you write with the framework occur, right? 

Also - just to be clear - if you write validation it will happen every single time we run one of the command you list below, right? So you write it one time and it gets applied everywhere.

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+        // These are example validators
+        stringLengthBetween(10, 256),
+        stringRegularExpression(regexp.MustCompile(`^[a-z0-9]+$`)),
+    },
+}
+```
+
+All validators will always be run, regardless of whether previous validators returned an error or not.
+
+-> The framework will implement or reference common use case attribute validations, such as string length, in the future.
```
laurapacilio, line 40: We probably don't want to say this only because it's making promises and we should avoid forward-looking statements. Instead, we probably just want to say that the framework does not currently reference common use case attribute validations. OR we omit this entirely :)

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+
+These syntax and basic schema checks are performed during the [`terraform apply`](https://www.terraform.io/docs/cli/commands/apply.html), [`terraform destroy`](https://www.terraform.io/docs/cli/commands/destroy.html), [`terraform plan`](https://www.terraform.io/docs/cli/commands/plan.html), and [`terraform validate`](https://www.terraform.io/docs/cli/commands/validate.html) commands. Any further validation provided by the framework occurs after these checks.
+
+## Attribute Validation
+
+It is common for provider implementations to introduce validation on attributes using the generic framework-defined types such as [`types.String`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/types#String). The [`tfsdk.Attribute` type `Validators` field](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#Attribute.Validators) can be supplied with a list of validations and diagnostics will be returned from all validators. For example:
```
laurapacilio, line 22:
```suggestion
You can introduce validation on attributes using the generic framework-defined types such as [`types.String`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/types#String). To do this, supply the [`tfsdk.Attribute` type `Validators` field](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#Attribute.Validators) with a list of validations, and the framework will return diagnostics from all validators. For example:
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+}
+```
+
+All validators will always be run, regardless of whether previous validators returned an error or not.
+
+-> The framework will implement or reference common use case attribute validations, such as string length, in the future.
+
+### Creating Attribute Validators
+
+To create an attribute validator, the [`tfsdk.AttributeValidator` interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#AttributeValidator) must be satisfied. For example:
```
laurapacilio, line 44:
```suggestion
To create an attribute validator, you must implement the [`tfsdk.AttributeValidator` interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#AttributeValidator). For example:
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+    return stringLengthBetweenValidator{
+        Max: maxLength,
+        Min: minLength,
+    }
+}
+```
+
+## Type Validation
+
+Providers that contain common attribute values with consistent validation rules may wish to create a custom type to simplify schemas. When validation is implemented on a type, it is not necessary to declare the same validation on the attribute, although additional validations can be supplied in that manner. For example:
```
laurapacilio, line 106:
```suggestion
You may want to create a custom type to simplify schemas if your provider that contains common attribute values with consistent validation rules. When you implement validation on a type, you do not need to declare the same validation on the attribute, but you can supply additional validations in that manner. For example:
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+    // to any validation performed by the type
+    Validators: []AttributeValidators{
+        stringLengthBetween(10, 256),
+    },
+}
+```
+
+### Defining Type Validation
+
+To support validation within a type, the [`attr.TypeWithValidate` interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/attr#TypeWithValidate) must be satisfied. For example:
```
laurapacilio, line 127:
```suggestion
To support validation within a type, you must implement the [`attr.TypeWithValidate` interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/attr#TypeWithValidate). For example:
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+            fmt.Sprintf("Missing `instance-` prefix, got: %s", value),
+        )
+        return diags
+    }
+}
+```
+
+## Schema Validation
+
+Provider, resource, and data source schemas also support validation across all attributes. This is helpful when checking values in multiple attributes, such as only accepting certain values in one attribute when another is a specific value.
```
laurapacilio, line 174: Could we make the example after "such as" more specific/a real-world scenario?

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+
+### Creating Provider Schema Validation
+
+There are two possible interface implementations that can be used for provider validation. Either or both can be implemented. This validation is performed in addition to any attribute and type validation within the provider schema.
```
laurapacilio, line 178:
```suggestion
The framework performs provider validation in addition to attribute and type validation. You can implement either or both of the following interfaces.
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+
+The [`tfsdk.ProviderWithConfigValidators` interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#ProviderWithConfigValidators) follows a similar pattern to attribute validation and allows for a more declarative approach, which is helpful for consistent validators across multiple providers. Each of the validators must satify the [`tfsdk.ProviderConfigValidator` interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#ProviderConfigValidator). For example:
```
laurapacilio, line 180:
```suggestion
The [`tfsdk.ProviderWithConfigValidators` interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#ProviderWithConfigValidators) follows a similar pattern to attribute validation and allows for a more declarative approach. This lets you write consistent validators across multiple providers. You must implement the [`tfsdk.ProviderConfigValidator` interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#ProviderConfigValidator) for each validator. For example:
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+type exampleProvider struct {}
+
+func (p exampleProvider) ConfigValidators(ctx context.Context) []tfsdk.ProviderConfigValidator {
+    return []tfsdk.ProviderConfigValidator{
+        /* ... */
+    }
+}
+```
+
+The [`tfsdk.ProviderWithValidateConfig` interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#ProviderWithValidateConfig) is more imperative in design and simplifies one-off functionality that typically applies to a single provider. For example:
```
laurapacilio, line 193:
```suggestion
The [`tfsdk.ProviderWithValidateConfig` interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#ProviderWithValidateConfig) is more imperative in design and is useful for validating unique functionality that typically applies to a single provider. For example:
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+
+func (p exampleProvider) ValidateConfig(ctx context.Context, req ValidateProviderConfigRequest, resp *ValidateProviderConfigResponse) {
+    // Retrieve values via req.Config.Get() or req.Config.GetAttribute(),
+    // then return any warnings or errors via resp.Diagnostics.
+}
+```
+
+### Creating Resource Schema Validation
+
+There are two possible interface implementations that can be used for resource validation. Either or both can be implemented. This validation is performed in addition to any attribute and type validation within the resource schema.
```
laurapacilio, line 207:
```suggestion
The framework performs resource schema validation in addition to any attribute and type validation. You can implement either or both of the following interfaces.
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+
+The [`tfsdk.ResourceWithConfigValidators` interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#ResourceWithConfigValidators) follows a similar pattern to attribute validation and allows for a more declarative approach, which is helpful for consistent validators across multiple resources. Each of the validators must satify the [`tfsdk.ResourceConfigValidator` interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#ResourceConfigValidator). For example:
```
laurapacilio, line 209:
```suggestion
The [`tfsdk.ResourceWithConfigValidators` interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#ResourceWithConfigValidators) follows a similar pattern to attribute validation and allows for a more declarative approach. This lets you create consistent validators across multiple resources. You must implement the [`tfsdk.ResourceConfigValidator` interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#ResourceConfigValidator) for each validator. For example:
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+type exampleResource struct {}
+
+func (r exampleResource) ConfigValidators(ctx context.Context) []tfsdk.ResourceConfigValidator {
+    return []tfsdk.ResourceConfigValidator{
+        /* ... */
+    }
+}
+```
+
+The [`tfsdk.ResourceWithValidateConfig` interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#ResourceWithValidateConfig) is more imperative in design and simplifies one-off functionality that typically applies to a single resource. For example:
```
laurapacilio, line 222:
```suggestion
The [`tfsdk.ResourceWithValidateConfig` interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#ResourceWithValidateConfig) is more imperative in design and is useful for validating unique functionality that typically applies to a single resource. For example:
```

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+
+func (r exampleResource) ValidateConfig(ctx context.Context, req ValidateResourceConfigRequest, resp *ValidateResourceConfigResponse) {
+    // Retrieve values via req.Config.Get() or req.Config.GetAttribute(),
+    // then return any warnings or errors via resp.Diagnostics.
+}
+```
+
+### Creating Data Source Schema Validation
+
+There are two possible interface implementations that can be used for data source validation. Either or both can be implemented. This validation is performed in addition to any attribute and type validation within the data source schema.
```
laurapacilio, line 236:
```suggestion
The framework performs data source schema validation in addition to any attribute and type validation. You can implement either or both of the following interfaces.
```

laurapacilio (approved):
I'm going to pre-approve so I don't block, but I have a few thoughts/language suggestions for you consider :) My goal is to make things concise and also to make sure we're using active voice, etc. throughout. Thank you for giving me the opportunity to do a full review, and thank you for your patience!

content/source/docs/plugin/framework/validation.html.md
bflad: Just removing an extra word from the suggestion 😄 

```suggestion
You may want to create a custom type to simplify schemas if your provider contains common attribute values with consistent validation rules. When you implement validation on a type, you do not need to declare the same validation on the attribute, but you can supply additional validations in that manner. For example:
```

content/source/docs/plugin/framework/validation.html.md
bflad: Let's omit it for now. 😄

content/source/docs/plugin/framework/validation.html.md
```diff
@@ -0,0 +1,261 @@
...
+        // These are example validators
+        stringLengthBetween(10, 256),
+        stringRegularExpression(regexp.MustCompile(`^[a-z0-9]+$`)),
+    },
+}
+```
+
+All validators will always be run, regardless of whether previous validators returned an error or not.
+
+-> The framework will implement or reference common use case attribute validations, such as string length, in the future.
```
bflad, line 40:
```suggestion
```

content/source/docs/plugin/framework/validation.html.md
bflad: Certainly makes sense! I'm worried about repeating this note up to 5 times (Creating Attribute Validators, Defining Type Validation, and the 3 schema validation sections). Do you have a suggestion for what do in this case?

content/source/docs/plugin/framework/validation.html.md
bflad: Newer practitioners and provider developers are generally curious about which type of validation is handled by which part of Terraform (CLI does some, SDK does some, provider does some). The general idea behind this section at a high level was to delineate between which of the validation responsibilities occurs where and when. I was thinking the general context might be helpful. It was maybe named poorly to highlight that this section is specifically what happens with Terraform CLI.

The Terraform CLI validation occurs before any framework/provider validation. Providers are not required to implement anything to receive the syntax and basic types validation. Maybe that's the key part missing from here that could help clarify this better.

content/source/docs/plugin/framework/validation.html.md
bflad: Maybe this will be slightly better (it is a little hard because these tend to be very-circumstance specific):

```suggestion
Provider, resource, and data source schemas also support validation across all attributes. This is helpful when checking values in multiple attributes, such as ensuring the values are compatible with each other.
```

bflad (commented):
Thanks so much, @laurapacilio! If you have a second, just some small followup questions/notes here.

content/source/docs/plugin/framework/validation.html.md
laurapacilio: fair - I will leave this note here then. :)

content/source/docs/plugin/framework/validation.html.md
laurapacilio: yeah I trust you on this but I wonder if we should say idk "Built-In Validation" "Default Validation"  or something similar? or even just "Default Terraform CLI Validation"?  yay. titles :p 

I just think this section title feels vague and so I want to be more specific about what it's about

bflad: Oh! I like Default Terraform CLI Validation -- let's give that a go. Thank you!

```suggestion
## Default Terraform CLI Validation
```

We can circle back about this after the ❄️ , if desired. 😄

PR #1969, opened 2021-11-17
Fix link to point to correct anchor.

anamahalo:
## Labels

API is checked b/c this has to do with fixing a broken link in the changelog!

- [x] inaccuracy
- [ ] clarification
- [ ] new docs


jondavidjohn (approved):

anamahalo:
@laurapacilio - tagged you but then I noticed you were off on Slack so asked a teammate since it was a quick fix :)

PR #1875, opened 2021-09-20
Tf api/fixes and cleanup

anamahalo:
## Labels

- [ ] inaccuracy
- [x] clarification
- [ ] new docs
- [ ] cosmetic bug - fixing broken text or markup
- [ ] enhancement - changing the website's behavior/layout
- [x] api - requires an update to the [changelog](https://www.terraform.io/docs/cloud/api/changelog.html)

API is marked but functionality is not new.
Please do not hesitate to reach out to me for questions!

PR #1974, opened 2021-11-19
Replace ; with . for all Query Parameters blocks

anamahalo:
- Follows up on a previous review suggestion
    https://github.com/hashicorp/terraform-website/pull/1875#discussion_r712522840
  - Since changes were preivously one-off prior
    new items going in remained with what was
    already across.  To respect style suggestions
    going back and making changes, but now across
    all areas so consistency remains intact.

## Labels

Just marking API since it affects API documentation but no changes.

- [ ] inaccuracy
- [ ] clarification
- [ ] new docs


laurapacilio (approved):
Thank you! I like this much better :-)

PR #1975, opened 2021-11-19
Update audit-trails.html.md

nicklhw:
Update doc to note that audit trail API is currently not available on Terraform Enterprise.

See my conversation with @amy-hashi for reference:

https://hashicorp.slack.com/archives/CKE00RWUT/p1637332702068600

## Labels

- [x] inaccuracy
- [x] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/cloud/api/audit-trails.html.md
+++ b/content/source/docs/cloud/api/audit-trails.html.md
@@ -22,11 +22,9 @@ page_title: "Audit Trails - API Docs - Terraform Cloud"
 
 # Audit Trails API
 
-Terraform Cloud retains 14 days of audit log information. The audit trails API exposes a stream of audit events, which describe changes to the application entities (workspaces, runs, etc.) that belong to a Terraform Cloud organization. Audit trails are a paid feature that is available as part of the **Terraform Cloud for Business** upgrade package. Refer to the [Terraform Cloud pricing page](https://www.hashicorp.com/products/terraform/pricing) for more details.
+Terraform Cloud retains 14 days of audit log information. The audit trails API exposes a stream of audit events, which describe changes to the application entities (workspaces, runs, etc.) that belong to a Terraform Cloud organization. Audit trails are a paid feature that is available as part of the **Terraform Cloud for Business** upgrade package. Refer to the [Terraform Cloud pricing page](https://www.hashicorp.com/products/terraform/pricing) for more details. Unlike other APIs, the Audit Trails API does not use the [JSON API specification](./index.html#json-api-formatting).
 
 -> **Note:** The Audit Trails API is not available for Terraform Enterprise.
-
--> **Note:** Unlike other APIs, the Audit Trails API does not use the [JSON API specification](./index.html#json-api-formatting).
 
 ## List an organization's audit events
 
--- a/content/source/docs/cloud/api/audit-trails.html.md
+++ b/content/source/docs/cloud/api/audit-trails.html.md
@@ -21,6 +21,7 @@ page_title: "Audit Trails - API Docs - Terraform Cloud"
 [JSON API error object]: https://jsonapi.org/format/#error-objects
 
 # Audit Trails API
+-> **Note:** Audit Trails API is currently not available for Terraform Enterprise.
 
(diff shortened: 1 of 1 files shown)


content/source/docs/cloud/api/audit-trails.html.md
```diff
@@ -21,6 +21,7 @@ page_title: "Audit Trails - API Docs - Terraform Cloud"
 [JSON API error object]: https://jsonapi.org/format/#error-objects
 
 # Audit Trails API
+-> **Note:** Audit Trails API is currently not available for Terraform Enterprise.
```
laurapacilio, line 24: Hi! Thanks for adding this :-) 

My first thought is that "currently" might imply that it will someday be supported in TFE. Even if we think it will be now, that could change so it's better to just say that it's not supported. 

I also noticed that we have four notes in a row up here at the top of this page. Is there any way that we could pull some of these notes out and put them in the text? I think we want one note and then the rest should be text - when we put everything in a separate note like this, it's hard for users to know what's important and what's not and it can be really overwhelming. What if we did something like this:

"Terraform Cloud retains 14 days of audit log information. The audit trails API exposes a stream of audit events, which describe changes to the application entities (workspaces, runs, etc.) that belong to a Terraform Cloud organization. Audit trails are a paid feature that is available in the **Terraform Cloud for Business** upgrade package. Refer to the [Terraform Cloud pricing](https://www.hashicorp.com/products/terraform/pricing) page for more details.

~> **Note:** The Audit Trails API is not available for Terraform Enterprise. 

Unlike other Terraform Cloud APIs, the audit trails API does not use the [JSON API specification](./index.html#json-api-formatting)."

^^and maybe we should also just say what kind of formatting it DOES use? or explain why it's different?

laurapacilio (changes requested):

content/source/docs/cloud/api/audit-trails.html.md
nicklhw: Thanks for the comments, I have updated with your suggestions. I am not exactly sure why the audit trails API does not use the JSON API spec or how it is different compare to the other API. Someone from engineering should probably chime in here.

anamahalo: @nicklhw I will reach out to you in DM about the above and hopefully the little I do know can help provide some insight.

content/source/docs/cloud/api/audit-trails.html.md
```diff
@@ -22,13 +22,11 @@ page_title: "Audit Trails - API Docs - Terraform Cloud"
...
+Terraform Cloud retains 14 days of audit log information. The audit trails API exposes a stream of audit events, which describe changes to the application entities (workspaces, runs, etc.) that belong to a Terraform Cloud organization. Audit trails are a paid feature that is available as part of the **Terraform Cloud for Business** upgrade package. Refer to the [Terraform Cloud pricing page](https://www.hashicorp.com/products/terraform/pricing) for more details.
 
--> **Note:** Unlike other APIs, the Audit Trails API does not use the [JSON API specification](./index.html#json-api-formatting).
-
--> **Note:** Terraform Cloud retains 14 days of audit log information.
+-> **Note:** The Audit Trails API is not available for Terraform Enterprise.
 
-The audit trails API exposes a stream of audit events, which describe changes to the application entities (workspaces, runs, etc.) that belong to a Terraform Cloud organization.
+-> **Note:** Unlike other APIs, the Audit Trails API does not use the [JSON API specification](./index.html#json-api-formatting).
```
laurapacilio, line 29: my last thought is - I think this could just be a one-liner underneath the nice paragraph you h ave above :-) Then the only note here would be the TFE one. And if you do that, I approve! Thank you so much for doing this and for your patience with the editing process!

laurapacilio (approved):
One final comment for you - once that's done, I say  we're good to merge! 🚀

content/source/docs/cloud/api/audit-trails.html.md
nicklhw: ok, done!

laurapacilio:
I like it! 👍  Feel free to merge whenever you feel ready!

PR #1986, opened 2021-11-29
Replace ; with . for all Query Parameters blocks

anamahalo:
- Admin + missed ones in search/find en masse
    - Follows up on a previous review suggestion
        https://github.com/hashicorp/terraform-website/pull/1875#discussion_r712522840
      - Since changes were preivously one-off prior
        new items going in remained with what was
        already across.  To respect style suggestions
        going back and making changes, but now across
        all areas so consistency remains intact.

Follow-up to https://github.com/hashicorp/terraform-website/pull/1974 happen to see some were missed in the mass search/find, but also admin was missed based on my original search so making sure all are included here.

## Labels

Just marking API since it affects API documentation but no changes.

- [ ] inaccuracy
- [ ] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/cloud/api/admin/organizations.html.md
+++ b/content/source/docs/cloud/api/admin/organizations.html.md
@@ -42,7 +42,7 @@ Status  | Response                                        | Reason
 
 ### Query Parameters
 
-This endpoint supports pagination [with standard URL query parameters](../index.html#query-parameters); remember to percent-encode `[` as `%5B` and `]` as `%5D` if your tooling doesn't automatically encode URLs.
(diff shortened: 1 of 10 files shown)


laurapacilio (approved):

anamahalo:
It looks like that failing GitLab link isn't related but guessing it's failing b/c the `/ce/` actually redirects to `/ee/`.  Will correct and see abouts CI.  :)

anamahalo:
Seems that didn't make a difference though fixing that redirect cleaned up CI a bit but the link that's there on the website works so seems it's a false negative.

anamahalo:
Just confirming here master is green so definitely a false negative.

laurapacilio:
Thank you so much for following up on that :-)

PR #1987, opened 2021-11-30
fix: moved list state versions outputs endpoint to state version outp…

sebasslash:
…uts docs page

Move the `List State Version Outputs` endpoint documentation to `State Version Outputs` API docs page and create a reference to that page in `State Version` API docs.

## Labels

- [ ] inaccuracy
- [ ] clarification
- [ ] new docs


content/source/docs/cloud/api/state-versions.html.md
```diff
@@ -593,98 +593,7 @@ curl \
...
-      "page-size": 20,
-      "prev-page": null,
-      "next-page": null,
-      "total-pages": 1,
-      "total-count": 2
-    }
-  }
-}
-```
+For more information please refer to the state version outputs documentation [here.](/docs/cloud/api/state-version-outputs.html#list-state-version-outputs)
```
nfagerlund, line 596: For various reasons, it's best to avoid having the text of a link be something like "here". (Screen readers present link text in isolation when you're jumping between links, so you want the text to actually say where the link goes.)

```suggestion
The output values from a state version are also available via the API. For details, see the [state version outputs documentation.](/docs/cloud/api/state-version-outputs.html#list-state-version-outputs)
```

nfagerlund (approved):
One fix to make link text more accessible; other than that, 👍🏼!

content/source/docs/cloud/api/state-version-outputs.html.md
```diff
@@ -14,6 +14,101 @@ State version outputs are the [output values](/docs/language/values/outputs.html
...
+
+Listing state version outputs requires permission to read state outputs for the workspace. ([More about permissions.](/docs/cloud/users-teams-organizations/permissions.html))
+
+Parameter | Description
+----------|---------
+`:state_version_id` | The ID of the desired state version.
+
+Status  | Response                                     | Reason
+--------|----------------------------------------------|----------
+[200][] | [JSON API document][]                        | Successfully returned a list of outputs for the given state version
```
laurapacilio, line 29: should we add periods after the reasons? It feels like in some tables we do end with a period and in some we don't. I would add them for consistency, but let me know if it's the convention that we don't have them here for some reason :)

laurapacilio (approved):
One minor comment but otherwise looks good with Nick's suggested edit! Thank you!

content/source/docs/cloud/api/state-version-outputs.html.md
sebasslash: Doing a quick visual scan of the docs, I see no period for the `Reason` column in the `Status Response` tables more frequently. Conversely, they are consistently used in `Request Body`, `Request Param` tables in the `Description` column. Would defer to @nfagerlund on this one.

nfagerlund: Ha, I totally missed that. 

If there's a convention here, I believe it's accidental! 

We should probably switch to using periods consistently in long table cells like these — either's fine in the case of a fragment, but when it might be either a fragment or a sentence, it's simpler to just put a dot. If you haven't merged it yet, we might as well start today, but it's clearly not urgent!

sebasslash: Copy that, I'll add the periods and keep that in mind for any future docs. Thanks @laurapacilio and @nfagerlund !

PR #1988, opened 2021-12-01
fix "target system" segment for module URLs in registry

jeremmfr:
Following the discussion in hashicorp/terraform#29532, the documentation for module URLs in public or private registry is ambiguous because it's not the provider in URL but a "target system" and the restriction for characters is not the same as for the name of a provider

## Labels

- [ ] inaccuracy
- [x] clarification
- [ ] new docs


laurenrother:
Hi @jeremmfr! I wanted to thank you for this PR and for the discussion on the issue you opened. While it is possible to create a custom provider with a `-` in the name right now, the private registry will soon support providers and the same naming restrictions will apply. Since these files explicitly deal with the registry, we want to keep the clear recommendation of our current standards. That said, this is an interesting point that perhaps we need a "what should you do when you want or need to break the naming rules/want to use an alternate registry/host" doc. Given that, I'm going to close this PR, but I do really appreciate your points here!

PR #1989, opened 2021-12-01
Redirect docs/plugin/framework/publishing.html to docs/registry/providers/publishing.html

bflad:
Closes hashicorp/terraform-plugin-framework#87

The manifest file is now recommended for any provider release so it is now included in the general provider publishing information. Each provider should include the appropriate protocol version (5.0 for Terraform Plugin SDK version 2 and 6.0 for Terraform Plugin Framework).

The Publishing to the Registry sidebar is promoted from under Framework to the top level Plugin Development navigation, as it is appliable to either SDK.

## Labels

- [ ] inaccuracy
- [x] clarification
- [ ] new docs


bflad:
The link check failures are associated with previously existing content on the Registry publishing page that points to GitHub pages that require being logged in. The links do work as-is, but I don't think they can work with the link checker due to the login redirect (unless there's some way to configure that to allow the 302).

paddycarver (approved):

bflad:
Going to give additional time (as much as they need 😄 ) for @hashicorp/terraform-education to review before merge. 👍

laurapacilio (approved):
Thanks!

PR #1954, opened 2021-11-11
TFE Azure AD document update

v-hagamp:

PR #1994, opened 2021-12-06
Use Microsoft docs as source of truth for Azure integration with TFE SSO

rexredinger:
## Labels

- [ ] inaccuracy
- [x] clarification
- [ ] new docs


laurapacilio:
guess what!? This PR encompasses a broader range...including this page! https://github.com/hashicorp/terraform-website/pull/1954

So I think this is a duplicate. However, we're going to suggest that the other PR link to this tutorial you found. Thank you for your work on this and I'm sorry for the churn!

rexredinger:
Thanks for the good catch @laurapacilio !

PR #2000, opened 2021-12-08
Update index.html.md

abuxton:
#ln1515 (approx) of the replicated install script now attempts to install container-selinux. 
it utilises a direct source method for the package so dependancies are not met. 

the change above calls out the dependancy `policycoreutils-python` the installer will exit if this package is not installed, manual installation of `container-selinux` infers this dependancy being met.

## Labels

- [] inaccuracy
- [x] clarification
- [ ] new docs

### DIFF
--- a/content/source/docs/enterprise/before-installing/index.html.md
+++ b/content/source/docs/enterprise/before-installing/index.html.md
@@ -314,3 +314,4 @@ For _Mounted Disk_ installations, you must:
     semanage fcontext -a -t container_file_t "/opt/tfe(/.*)?"
     restorecon -R /opt/tfe
     ```
+-> ** Note: ** The `install.sh` script will attempt to install `container-selinux` in online mode. Regradless of mode container-selinux package as a dependacy on the `policycoreutils-python` package being installed.


content/source/docs/enterprise/before-installing/index.html.md
```diff
@@ -314,3 +314,4 @@ For _Mounted Disk_ installations, you must:
     semanage fcontext -a -t container_file_t "/opt/tfe(/.*)?"
     restorecon -R /opt/tfe
     ```
+-> ** Note: ** The `install.sh` script will attempt to install `container-selinux` in online mode. Regradless of mode container-selinux package as a dependacy on the `policycoreutils-python` package being installed.
```
sudomateo, line 317:
```suggestion
-> **Note:** The online installer will attempt to install the `container-selinux` if it is not already installed.
```

sudomateo (changes requested):
Thank you for the pull request! Would you be able to provide more context on why this change is necessary? The existing documentation already mentions installing the `container-selinux` package and the online installer already attempts to install the `container-selinux` package if it does not already exist.

content/source/docs/enterprise/before-installing/index.html.md
abuxton: the install of `container-selinux ` will fail under RHEL as the installation method does not install `policycoreutils-python` which is a dependency on the OS of the container and not handled by the installer, `policycoreutils-python` is not part of the EPEL or RHEL extras standard packages , it is a package in Centos base  
the important part of the call out is the requirement for `policycoreutils-python`

laurapacilio:
Hi @abuxton we recently migrated our documentation site to next.js, which required that we migrate all of our files to .mdx. As a result, we need to recreate this PR in the new .mdx files for the site. I don't want this good work to get lost, so please let me know if you would like me to open a new PR that transfers these changes to the new files or if you feel comfortable doing so. I apologize for the inconvenience, but I hope you will enjoy all of the nice new documentation features that the migration brings (better documentation search, nicer code blocks - just to name a few!).

Thank you again for your work on this, and please let me know how I can be helpful!

sudomateo:
I'm waiting to hear back from Replicated on this since this happens in their installation script. Perhaps this will be followed up upon in the future.

PR #2003, opened 2021-12-08
Fix broken glossary link

laurapacilio:
Fix a broken link to the backends page from a recent PR.


apparentlymart (approved):
LGTM! I verified that the link here went to a specific anchor on the page when I loaded it in a browser.

PR #2010, opened 2021-12-14
Migrate docs to MDX

dstaley:
This PR overwrites the contents of this repo with the contents of `terraform-website-next`, which is the site that has been serving `terraform.io` for a hot minute.

Also along for the ride is the migration of our docs from markdown to MDX! :tada:


laurapacilio (approved):
🥳

laurapacilio (approved):
left a wording suggestion but otherwise - thank you!

PR #2014, opened 2021-12-15
Install the latest version of npm in CircleCI

dstaley:
CircleCI uses an older version of npm that incorrectly downloads incompatible binaries for swc when installing Next.js, so this PR ensures that we're always using the latest version of npm (which mirrors the behavior we use when building our Docker image).


BRKalow (approved):
👍

PR #2015, opened 2021-12-15
Create TFE v202112-2 Release Notes

bnferguson:
This PR adds the release notes for Terraform Enterprise version v202112-2.

### DIFF
--- a/content/enterprise/release/v202112-2.mdx
+++ b/content/enterprise/release/v202112-2.mdx
@@ -1,0 +1,41 @@
+---
+page_title: Releases - Terraform Enterprise
+---
+
+# TFE Release v202112-2 (590)
+
+APPLICATION LEVEL BUG FIXES SINCE v202112-1:
+
+1. Updated the version of the terraform-registry that shipped with the v202112-1 release.
+
+UPCOMING DEPRECATION NOTICE:
+
+1. Effective April, 2022 there will be an update to Terraform Enterprise container names. This change may break container monitoring or custom tooling that identifies containers by name. More specific information regarding name changes will be made available in future release notes.
+
+
+APPLICATION LEVEL BREAKING CHANGES:
+
+
+APPLICATION LEVEL FEATURES:
+1. SAML certificate signing and digest methods now are configurable
+1. Added conditional pagination ability on GET indices for the following: SSH Keys, Parameters (on Policy Sets), Policy Checks (on Runs), Organizations, Policy Checks (on Runs), Oauth Clients, Oauth Tokens, Authentication (User) Tokens, Notification Configurations, Feature Sets, Feature Sets (on Organizations). If pagination parameters are not provided all results will be returned.
+1. Added support for public provider and public module curation
+1. Added support for tfc-agent 1.x series
+
+APPLICATION LEVEL BUG FIXES:
+1. Fixed a bug where the `tfe-admin` command would set a configuration key to the value `''` instead of unsetting the configuration value.
+1. Fixed an issue where custom CA certificates were not injected into the `tfe-fluent-bit` container.
+1. Fixed an issue where Replicated snapshots were not executing for demo mode installations.
+1. Fixed provider/module APIs to allow prefix searching, also fixes bug where providers were returned for unrelated (but member of) organizations.
+1. Fixed structured run output to show a less verbose diff for json-encoded array fields
+1. Added a fix to prevent the removal of the last owner of an organization via the API. It also gives precedence to returning an error if you remove the last owner over removing yourself (if you are an owner of the organization). Meaning that even though if you try to remove yourself (and you happen to be the only owner), the error that you'll receive is the same as if one tried to remove the only owner. You'll only receive the error message: `You cannot remove yourself from an organization you own` if you try to remove yourself and are not the only owner of an organization.  Therefore the unit tests for "removing self as owner" had to be updated to include multiple owners in the organization. 
+
+
+APPLICATION LEVEL SECURITY FIXES:
+1. The Docker container running Nomad (`ptfe_nomad`) no longer runs with the `privileged` attribute.
+1. Updated the version of the internally-managed Vault server to 1.9.0
+1. Updated the version of the internally-managed Nomad server to 1.1.6
+1. Updated `tfe-fluent-bit` to use Fluent Bit 1.8.10.
+1. Updated `archivist` to address CVE in direct and indirect `jwt-go` dependency.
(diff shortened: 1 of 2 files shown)


dstaley:
@bnferguson Sorry for the force push! I confused TFE with the core CLI.

content/enterprise/release/v202112-2.mdx
```diff
@@ -0,0 +1,41 @@
...
+APPLICATION LEVEL BUG FIXES SINCE v202112-1:
+
+1. Updated the version of the terraform-registry that shipped with the v202112-1 release.
+
+UPCOMING DEPRECATION NOTICE:
+
+1. Effective April, 2022 there will be an update to Terraform Enterprise container names. This change may break container monitoring or custom tooling that identifies containers by name. More specific information regarding name changes will be made available in future release notes.
+
+
+APPLICATION LEVEL BREAKING CHANGES:
```
laurapacilio, line 16: do we think it might make sense to put "none" here? I just think having a heading with nothing underneath might look weird.

content/enterprise/release/v202112-2.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+
+APPLICATION LEVEL FEATURES:
```
laurapacilio, line 19: Add periods to the end of all of the sentences :)

content/enterprise/release/v202112-2.mdx
```diff
@@ -0,0 +1,41 @@
...
+1. Fixed an issue where custom CA certificates were not injected into the `tfe-fluent-bit` container.
+1. Fixed an issue where Replicated snapshots were not executing for demo mode installations.
+1. Fixed provider/module APIs to allow prefix searching, also fixes bug where providers were returned for unrelated (but member of) organizations.
+1. Fixed structured run output to show a less verbose diff for json-encoded array fields
+1. Added a fix to prevent the removal of the last owner of an organization via the API. It also gives precedence to returning an error if you remove the last owner over removing yourself (if you are an owner of the organization). Meaning that even though if you try to remove yourself (and you happen to be the only owner), the error that you'll receive is the same as if one tried to remove the only owner. You'll only receive the error message: `You cannot remove yourself from an organization you own` if you try to remove yourself and are not the only owner of an organization.  Therefore the unit tests for "removing self as owner" had to be updated to include multiple owners in the organization.
```
laurapacilio, line 31: I don't really understand this sentence: "It also gives precedence to returning an error if you are an organization owner and you remove the last owner over removing yourself."  Does it mean "The changes also cause Terraform Enterprise to return an error if you are the only organization owner and you try to remove yourself from the organization."

And the last sentence about unit tests - do users need to know that, or do they just need to know the result - that the functionality has changed?

```suggestion
1. Added a fix to prevent the removal of the last owner of an organization via the API. It also gives precedence to returning an error if you are an organization owner and you remove the last owner over removing yourself. If you try to remove yourself and you are the only owner, the error is the same as if you tried to remove the only owner:`You cannot remove yourself from an organization you own`. Therefore, we updated the unit tests for "removing self as owner" to include multiple owners in the organization. 
```

content/enterprise/release/v202112-2.mdx
```diff
@@ -0,0 +1,41 @@
...
+1. Fixed structured run output to show a less verbose diff for json-encoded array fields
+1. Added a fix to prevent the removal of the last owner of an organization via the API. It also gives precedence to returning an error if you remove the last owner over removing yourself (if you are an owner of the organization). Meaning that even though if you try to remove yourself (and you happen to be the only owner), the error that you'll receive is the same as if one tried to remove the only owner. You'll only receive the error message: `You cannot remove yourself from an organization you own` if you try to remove yourself and are not the only owner of an organization.  Therefore the unit tests for "removing self as owner" had to be updated to include multiple owners in the organization. 
+
+
+APPLICATION LEVEL SECURITY FIXES:
+1. The Docker container running Nomad (`ptfe_nomad`) no longer runs with the `privileged` attribute.
+1. Updated the version of the internally-managed Vault server to 1.9.0
+1. Updated the version of the internally-managed Nomad server to 1.1.6
```
laurapacilio, line 37: add periods :)

content/enterprise/release/v202112-2.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+1. Effective April, 2022 there will be an update to Terraform Enterprise container names. This change may break container monitoring or custom tooling that identifies containers by name. More specific information regarding name changes will be made available in future release notes.
+
+
+APPLICATION LEVEL BREAKING CHANGES:
+
+
+APPLICATION LEVEL FEATURES:
+1. SAML certificate signing and digest methods now are configurable
+1. Added conditional pagination ability on GET indices for the following: SSH Keys, Parameters (on Policy Sets), Policy Checks (on Runs), Organizations, Policy Checks (on Runs), Oauth Clients, Oauth Tokens, Authentication (User) Tokens, Notification Configurations, Feature Sets, Feature Sets (on Organizations). If pagination parameters are not provided all results will be returned.
```
laurapacilio, line 21:
```suggestion
1. Added conditional pagination ability on GET indices for the following: SSH Keys, Parameters (on Policy Sets), Policy Checks (on Runs), Organizations, Policy Checks (on Runs), Oauth Clients, Oauth Tokens, Authentication (User) Tokens, Notification Configurations, Feature Sets, Feature Sets (on Organizations). All results are returned if pagination parameters are not provided.
```

content/enterprise/release/v202112-2.mdx
```diff
@@ -0,0 +1,41 @@
...
+1. Added support for public provider and public module curation
```
laurapacilio, line 22:
```suggestion
1. Added support for public provider and public module curation
```
```suggestion
1. Added support for public provider and public module curation.
```

laurapacilio (approved):
Approving but I had a couple of wording suggestions for you :) Feel free to ignore any that materially change the meaning. Thank you!

content/enterprise/release/v202112-2.mdx
bnferguson: Yeah that's an excellent call. Also wonder if we should change these from numbered lists to bulleted because 

1. None

Feels really funny but having it without breaks the type hierarchy a bit.

laurapacilio: yes I totally understand! In this ONE instance, I might say you need neither. None could just be on its own without a bullet. Typically when there is only one item, I think it's better to not use bullets.

PR #2016, opened 2021-12-15
Pull submodules when building search index

dstaley:
CircleCI's `checkout` step doesn't initialize submodules, so we're only pushing search index entries from docs in this repo. This change makes sure that we initialize the submodules prior to generating the index.


BRKalow (approved):
Nice!

PR #2017, opened 2021-12-15
Bump react-search

dstaley:
This PR bumps `@hashicorp/react-search` to the latest version, which should allow us to index content from symlinked directories.


dstaley:
We'll get most of these updates when we migrate to Next.js v12.

PR #2029, opened 2021-12-16
Properly handle more redirects from old URLs

dstaley:
Currently, we have redirects defined using the canonical URL from the original sitemap, which includes `.html` and `index.html`. However, S3 would allow you to redirect from pages like `/docs/cdktf` to `/docs/cdktf/index.html` and `/docs/enterprise-beta/workspaces/naming` to `/docs/enterprise-beta/workspaces/naming.html`.

This PR replicates that functionality by appending `/index.html` and `.html` to incoming `/docs` requests and checks if any of them are defined in the `docs-redirects.json` file. If so, it performs the redirect.

Additionally, this PR returns a 404 error when the incoming request is for a page that's not part of our very small number of defined `/docs` pages (currently `/docs/glossary` and `/docs/partnerships`) and also isn't a defined redirect. Previously we returned a 500 error since we attempted to call the `generateStaticProps` method.

### DIFF
--- a/data/docs-redirects.js
+++ b/data/docs-redirects.js
@@ -845,7 +845,7 @@ export const docsRedirects = {
   '/docs/extend/hashicorp-provider-design-principles.html': '/plugin/hashicorp-provider-design-principles',
   '/docs/extend/how-terraform-works.html': '/plugin/how-terraform-works',
   '/docs/extend/index.html': '/plugin',
-  '/docs/extend/plugin-sdk.html\t\t\t\t': '/plugin/sdkv2/guides/v1-upgrade-guide',
+  '/docs/extend/plugin-sdk.html': '/plugin/sdkv2/guides/v1-upgrade-guide',
   '/docs/extend/plugin-types.html': '/plugin',
   '/docs/extend/resources.html': '/plugin/sdkv2/resources',
   '/docs/extend/resources/customizing-differences.html': '/plugin/sdkv2/resources/customizing-differences',
--- a/pages/docs/[[...page]].jsx
+++ b/pages/docs/[[...page]].jsx
@@ -34,17 +34,32 @@ export async function getStaticPaths() {
 }
 
 export async function getStaticProps({ params }) {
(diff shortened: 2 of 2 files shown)


pages/docs/[[...page]].jsx
```diff
@@ -34,17 +34,32 @@ export async function getStaticPaths() {
...
+        }
       }
     }
+
+    // We've hit a /docs page that's not in our redirects, and isn't one of the
+    // defined /docs routes, so return a 404 instead of trying to call the
+    // generateStaticProps method which returns a 500 error.
+    return {
+      notFound: true,
+    }
```
Comment (BRKalow) [inline pages/docs/[[...page]].jsx:60]: 👍

BRKalow:
Would something like this: https://github.com/hashicorp/vault/blob/master/website/redirects.next.js#L745-L747 in our redirects file still necessitate the need for this logic?

dstaley:
Those redirects are processed prior to the logic on our `/docs` page, so if we added those we'd need to update the 'docs-redirects.json` file to remove `.html` and `index.html` from all of the entries. This would also mean that anyone who updates the `docs-redirects.json` file in the future needs to remember that the URL they use can't have `.html` or `index.html`. I think we can alleviate this with a comment at the top of the file.

BRKalow (approved):

data/docs-redirects.js
```diff
@@ -845,7 +845,7 @@ export const docsRedirects = {
   '/docs/extend/hashicorp-provider-design-principles.html': '/plugin/hashicorp-provider-design-principles',
   '/docs/extend/how-terraform-works.html': '/plugin/how-terraform-works',
   '/docs/extend/index.html': '/plugin',
-  '/docs/extend/plugin-sdk.html\t\t\t\t': '/plugin/sdkv2/guides/v1-upgrade-guide',
```
jescalan, line 848: 🤦

jescalan (commented):
Thank you for tackling this one so quickly!

PR #2031, opened 2021-12-16
Update readme to remove custom components

laurapacilio:
We are evaluating which custom components we want to make available on the site and the guidance we will give about how to properly use these components. We are temporarily removing the custom components details until we have more detailed guidance. This approach will lead to higher-quality pull requests and documentation, saving contributors time and giving end users a more consistent experience.


README.md
```diff
@@ -189,90 +189,7 @@ There are several custom markdown plugins that are available by default that enh
...
-
-Continued markdown content...
-```
-
-It's also worth noting that this component will automatically adjust to the correct product colors depending on the context.
-
-#### Other Components
-
-Other custom components can be made available on a per-site basis, the above are the standards. If you have questions about custom components that are not documented here, or have a request for a new custom component, please reach out to @hashicorp/digital-marketing.
+A number of custom [mdx components](https://mdxjs.com/) are available for use within any `.mdx` file. Please refer to the [Swingset](https://react-components.vercel.app/?component=Tabs) documentation for the latest examples and API reference. If you have questions about custom components that are not documented here, or have a request for a new custom component, please reach out to @hashicorp/digital-marketing.
```
dstaley, line 192: Only a subset of those components are available by default (basically `Tabs`, `EnterpriseAlert` and code blocks); all the remaining ones need to be imported in the doc file manually. Since that's the case I'd prefer we not link out to the component API at all.

dstaley (changes requested):

README.md
judithpatudith: I'm ok with this if you are Laura :)

laurapacilio: yes! I will remove :)

laurapacilio, line 192:
```suggestion
A number of custom [mdx components](https://mdxjs.com/) are available for use within any `.mdx` file. If you have questions about custom components, or have a request for a new custom component, please reach out to @hashicorp/digital-marketing.
```

dstaley (approved):

PR #2033, opened 2021-12-17
plugin: Recommend Terraform v1.0 for framework instead of v1.0.3

bflad:
Practitioners should generally opt for the latest available patch version of a Terraform CLI minor version. Having this level of specificity in the documentation could potentially be confusing.

### DIFF
--- a/content/plugin/framework/index.mdx
+++ b/content/plugin/framework/index.mdx
@@ -9,7 +9,7 @@ description: |-
 
 The plugin framework is a new way to develop Terraform plugins, providing improvements and new features from [Teraform Plugin SDKv2](/plugin/sdkv2/sdkv2-intro).
 
-~> **Important**: HashiCorp is currently developing providers against the framework, but it is still early in its development cycle. Its interfaces are subject to change and it may not yet support all of your desired features. It can also only build providers that require **Terraform v1.0.3 or later**.
+~> **Important**: HashiCorp is currently developing providers against the framework, but it is still early in its development cycle. Its interfaces are subject to change and it may not yet support all of your desired features. It can also only build providers that require **Terraform v1.0 or later**.
 [Which SDK Should I Use?](/plugin/which-sdk) can help you decide whether the framework is right for your provider.
 
 ## Get Started
--- a/content/plugin/framework/index.mdx
+++ b/content/plugin/framework/index.mdx
@@ -9,7 +9,7 @@ description: |-
 
 The plugin framework is a new way to develop Terraform plugins, providing improvements and new features from [Teraform Plugin SDKv2](/plugin/sdkv2/sdkv2-intro).
 
-~> **Important**: HashiCorp is currently developing providers against the framework, but it is still early in its development cycle. Its interfaces are subject to change and it may not yet support all of your desired features. It can also only build providers that require **Terraform v1.0 or later**.
+~> **Important**: HashiCorp is currently developing providers against the framework, but it is still early in its development cycle. Its interfaces are subject to change and it may not yet support all of your desired features. It can also only build providers that require **Terraform CLI v1.0 or later**.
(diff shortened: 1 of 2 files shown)


bflad:
Maybe we should also specify Terraform _CLI_ when mentioning versions, but I'll let folks decide on whether or not that would also be a helpful change.

laurapacilio (approved):
🥳

laurapacilio:
> Maybe we should also specify Terraform _CLI_ when mentioning versions, but I'll let folks decide on whether or not that would also be a helpful change.

@bflad good point!  If it is something that folks should consider when picking a framework then we 100% should mention it :) My opinion is it can't hurt and it may be very helpful!

laurapacilio (changes requested):
We should also change the mention here: https://www.terraform.io/plugin/framework

content/plugin/framework/index.mdx
```diff
@@ -9,7 +9,7 @@ description: |-
 
 The plugin framework is a new way to develop Terraform plugins, providing improvements and new features from [Teraform Plugin SDKv2](/plugin/sdkv2/sdkv2-intro).
 
-~> **Important**: HashiCorp is currently developing providers against the framework, but it is still early in its development cycle. Its interfaces are subject to change and it may not yet support all of your desired features. It can also only build providers that require **Terraform v1.0.3 or later**.
+~> **Important**: HashiCorp is currently developing providers against the framework, but it is still early in its development cycle. Its interfaces are subject to change and it may not yet support all of your desired features. It can also only build providers that require **Terraform v1.0 or later**.
```
bflad, line 12:
```suggestion
~> **Important**: HashiCorp is currently developing providers against the framework, but it is still early in its development cycle. Its interfaces are subject to change and it may not yet support all of your desired features. It can also only build providers that require **Terraform CLI v1.0 or later**.
```

content/plugin/which-sdk.mdx
```diff
@@ -25,14 +25,14 @@ to build resources into an SDKv2 provider.
 If you are developing a greenfield provider, you should consider building it
 with the framework.
 
-## Do You Need to Support Users Not On Terraform v1.0.3 Yet?
+## Do You Need to Support Users Not On Terraform v1.0 Yet?
```
bflad, line 28:
```suggestion
## Do You Need to Support Users Not On Terraform CLI v1.0 Yet?
```

content/plugin/which-sdk.mdx
```diff
@@ -25,14 +25,14 @@ to build resources into an SDKv2 provider.
...
 
-If you support user bases that have not upgraded to Terraform v1.0.3 in
+If you support user bases that have not upgraded to Terraform v1.0 in
```
bflad, line 30:
```suggestion
If you support user bases that have not upgraded to Terraform CLI v1.0 in
```

content/plugin/which-sdk.mdx
```diff
@@ -25,14 +25,14 @@ to build resources into an SDKv2 provider.
...
 significant numbers, we recommend that you continue using SDKv2. The framework
 is built on Terraform protocol version 6, and Terraform could not download
-providers using protocol version 6 until Terraform v1.0.3.
+providers using protocol version 6 until Terraform v1.0.
```
bflad, line 33:
```suggestion
providers using protocol version 6 until Terraform CLI v1.0.
```

bflad:
Updated to include CLI when saying Terraform and I believe the /plugin/framework page change is already covered here: https://github.com/hashicorp/terraform-website/pull/2033/files#diff-6637e1186b1cd35477170f690384da00990f24ced068ec0c437996cdcc72afb1

laurapacilio (approved):
Thank you :)

paddycarver (approved):

PR #2035, opened 2021-12-17
plugin: Add terraform-provider-scaffolding and terraform-provider-scaffolding-framework CTAs

bflad:
Closes hashicorp/terraform-provider-scaffolding-framework#4

This should help increase visibility of this template repository, which can be very helpful for developers to bootstrap a new provider.

### DIFF
--- a/content/plugin/framework/index.mdx
+++ b/content/plugin/framework/index.mdx
@@ -14,7 +14,8 @@ The plugin framework is a new way to develop Terraform plugins, providing improv
 
 ## Get Started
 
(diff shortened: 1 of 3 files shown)


laurapacilio (commented):
I like it!! I wonder - are there other places we could add this if it's very helpful? 
I think maybe here: https://www.terraform.io/plugin/framework

bflad:
This change is already for /plugin/framework 😄  if there are other places you think it might be useful, happy to add the links!

laurapacilio:
Haha - forgive me! I was looking at the "Index" part of the file path and I confused /plugin/framework with this page that also has a "get started" section where maybe this would be helpful: https://www.terraform.io/plugin. Essentially, let's add it to both of them!

paddycarver (approved):

bflad:
Updated the 3 SDK index pages to each have template repository call to actions. 👍

laurapacilio (approved):
yay! thank you for adding it all the places!

PR #2036, opened 2021-12-17
Optimize Docker layers by building both images from same Dockerfile

dstaley:
This PR builds both our `:latest` and `:full` images from the same Dockerfile, using multi-stage builds. This gives a slight speed up (since we can reuse the previous base image if deps haven't changed), while ensuring that the `node_modules` folder between the two images is the same.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,9 @@
-FROM docker.mirror.hashicorp.services/node:14.17.0-alpine
+FROM docker.mirror.hashicorp.services/node:14.17.0-alpine AS base
 RUN apk add --update --no-cache git make g++ automake autoconf libtool nasm libpng-dev
 
-COPY ./package.json /website/package.json
-COPY ./package-lock.json /website/package-lock.json
+COPY ./package.json ./package-lock.json /website/
 WORKDIR /website
-RUN npm install -g npm@latest
-RUN npm install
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,8 +1,9 @@
...
-COPY ./package-lock.json /website/package-lock.json
+COPY ./package.json ./package-lock.json /website/
 WORKDIR /website
-RUN npm install -g npm@latest
-RUN npm install
\ No newline at end of file
+RUN npm install -g npm@latest && npm ci
+
+FROM base AS full
+COPY . /website
```
thiskevinwang, line 9: Loose understanding here:

I think you want something like 

```suggestion
FROM docker.mirror.hashicorp.services/node:14.17.0-alpine
COPY --from=base . /website
```

- Select a clean base image, like `node:14.17.0-alpine` again
- Copy stuff as needed from your _builder_ (named `base` here)
  - The exact `COPY` args here needs confirmation

dstaley: I think we're gonna need the same dependencies as the base image (since we're running Next.js in dev mode), so at that point I think it makes more sense to use `FROM base` since that saves us from doing another `apk add`.

thiskevinwang: Oh that makes sense! (I was missing the intent behind the Dockerfile)

PR #2043, opened 2021-12-20
Prevent the rendering of plain text Markdown references

dstaley:
The MDX parser has a bug where lines beginning with quote characters immediately following a Markdown reference definition will cause the definition to render in plain text. This PR adds the text "Acronym for" before these quoted strings to prevent this from happening.

Before:
image

After:
image


laurapacilio (approved):
Thank you! I like it :-)

PR #2048, opened 2021-12-21
CDKTF Website update

team-tf-cdk:
Updates CDKTF website to SHA 22dfee34c8e28716826a65424c7c912602f7f91d


laurapacilio:
Closing because there is a more recent one! Per guidance from web dev, we don't need this one anymore because there is a more recent automatic PR to update the CDKTF docs!

PR #2055, opened 2021-12-30
Refer to non-deprecated helper

will-in-wi:
`ValidateRFC3339TimeString` has been deprecated in favor of `IsRFC3339Time` and is completely removed in v2. I've updated the docs to reflect this, and updated the doc link to point to the v2 version of the api docs for this method.

Thanks for Terraform!


bflad (approved):
Looks good to me, thank you for this, @will-in-wi 🚀

PR #2058, opened 2022-01-03
feat: update downloads page to improve arch display

BRKalow:
Updating the arch display per internal discussions, this will just display the raw arch slug/identifier for more clarity.


thiskevinwang (approved):
Looks straightforward to me!

Is this the only diff?

https://user-images.githubusercontent.com/26389321/147970498-a1a8456e-4d6b-458d-8b0c-c0625e21b33e.mov

BRKalow:
@thiskevinwang Correct, that should be the only difference! 😄

PR #2062, opened 2022-01-04
Clarify some information about paginating the List Organizations endpoint

nfagerlund:
Once upon a time we went through a paginate-and-deprecate process for all the
list endpoints, but we never really finished the process for orgs, so an
unqualified request will still give you the full list.

1. That's bad and we want to change it soon, so therefore we should let people
know they shouldn't rely on it. Stay tuned.

2. The "if omitted" bit in the params table was a little unclear. The pagination
defaults only kick in if you provide at least ONE pagination query param; if you
omit both, you get the whole enchilada.

### DIFF
--- a/content/cloud-docs/api-docs/organizations.mdx
+++ b/content/cloud-docs/api-docs/organizations.mdx
@@ -51,7 +51,7 @@ The Organizations API is used to list, show, create, update, and destroy organiz
 
 This endpoint supports pagination [with standard URL query parameters](/cloud-docs/api-docs/#query-parameters). Remember to percent-encode `[` as `%5B` and `]` as `%5D` if your tooling doesn't automatically encode URLs.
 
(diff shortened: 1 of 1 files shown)


laurapacilio:
My only thought on this is - we typically shy away from future-facing statements like the one we are making here. Is it certain that we will make these changes? I think if it's absolutely certain, it's okay and it's good to warn folks. I just always get nervous when talking about what we're going to do, as plans can change :)

laurapacilio:
@nfagerlund are you still working on this? I didn't want to overstep if you weren't ready, but if you are I can review/approve so we can get this merged!

nfagerlund:
@laurapacilio whoop, sorry for missing the notif! You're right, so I adjusted this to avoid a reference to the ✨ the future ✨.

laurapacilio (approved):
Thank you!  Ship it! 🚀

PR #2063, opened 2022-01-04
Update api docs for updated global variable sets

rexredinger:
This is an update to the docs to match the update from the PR https://github.com/hashicorp/atlas/pull/11519

`is-global` becomes simply `global`


laurapacilio (approved):
Thank you! 🚀

PR #2064, opened 2022-01-04
website: filter out unwanted versions.

BRKalow:
Per internal discussions, filtering out old and unwanted versions to ensure our users aren't downloading versions with bugs. This should work without modification as new versions are released.

After:
Screen Shot 2022-01-04 at 2 56 50 PM


thiskevinwang (approved):
LGTM - I’ll merge this so we can hide the borked versions now, but I believe there’s an additional change needed for future patch releases.

PR #2066, opened 2022-01-05
Add redirect for provisioners overview page

laurapacilio:
Do not merge until https://github.com/hashicorp/terraform/pull/30296 is merged. 

Refer to https://github.com/hashicorp/terraform/issues/30207 for details about why the provisioners "Overview" page has been removed.

### DIFF
--- a/data/docs-redirects.js
+++ b/data/docs-redirects.js
@@ -1520,4 +1520,6 @@ export const docsRedirects = {
   '/docs/state/remote.html': '/language/state/remote',
   '/docs/state/sensitive-data.html': '/language/state/sensitive-data',
   '/docs/state/workspaces.html': '/language/state/workspaces',
+  '/language/resources/provisioners': '/language/resources/provisioners/syntax',
+  '/language/resources/provisioners/#provisioners-are-a-last-resort': '/language/resources/provisioners/syntax#provisioners-are-a-last-resort'
 }


dstaley (changes requested):
Alright so this one might need a few more changes just to keep things tidy:

1. If we no longer want to have the "Overview" page at all, I think the right thing would be to delete the `language/provisioners/index.mdx` file and remove it from the `language-nav-data.json` file. Alternatively, we could simply mark the "Overview" nav data object with `"hidden": true`.
2. If we're redirecting _new_ usages of `/language/resources/provisioners`, we should declare that redirect in `redirects.next.js`. The `docs-redirects.js` file is only for redirecting old URLs that start with `/docs`. I'm leaning towards not adding this redirect at all since we've had that URL for less than a month.
3. The second redirect (the one with the anchor link) would need a preceding `/docs` based on the URL that Martin shared.

laurapacilio:
@dstaley Thank you for helping me through my first redirect attempt with the new site haha. 

> 1. If we no longer want to have the "Overview" page at all, I think the right thing would be to delete the `language/provisioners/index.mdx` file and remove it from the `language-nav-data.json` file. Alternatively, we could simply mark the "Overview" nav data object with `"hidden": true`.
- I've got a PR in the `terraform` repo that does these two things! :-) https://github.com/hashicorp/terraform/pull/30296  Let me know if you see anything weird there. These files are in that repo, so that's why I made the changes there.

> 2. If we're redirecting _new_ usages of `/language/resources/provisioners`, we should declare that redirect in `redirects.next.js`. The `docs-redirects.js` file is only for redirecting old URLs that start with `/docs`. I'm leaning towards not adding this redirect at all since we've had that URL for less than a month.
- Okay - that feels okay to me! I will not redirect the new page then and focus on any OLD pages that deal with this. I just want to make sure if someone had it bookmarked, they won't be upset. If there are other places on the site that link to that page....how can I make sure they don't break?

> 3. The second redirect (the one with the anchor link) would need a preceding `/docs` based on the URL that Martin shared.
- okay - I'll make an update! :)

dstaley (approved):

PR #2068, opened 2022-01-06
Update index.mdx

nervda:
Added text to disable any lifecycle policies on the Object Storage containers as this may cause data inconsistencies.


content/enterprise/before-installing/index.mdx
```diff
@@ -81,6 +81,7 @@ Make sure your data storage services or device meet Terraform Enterprise's requi
 - _External Services_
   - [PostgreSQL Requirements](/enterprise/before-installing/data-storage-requirements/postgres-requirements)
   - Any S3-compatible object storage service, GCP Cloud Storage or Azure blob storage meets Terraform Enterprise's object storage requirements. You must create a bucket for Terraform Enterprise to use, and specify that bucket during installation. Depending on your infrastructure provider, you might need to ensure the bucket is in the same region as the Terraform Enterprise instance.
+    - Disable any lifecycle rules that would delete, archive or transition objects in this container. TFE expects to manage all data the object storage service and any lifecycle moves may result in unexpected data inconsistencies.
```
laurapacilio, line 84: Just a tweak for you :-) See what you think!  

```suggestion
    - Disable any lifecycle rules that would delete, archive, or transition objects in this container. Terraform Enterprise expects to manage all data the object storage service, so any lifecycle moves may result in unexpected data inconsistencies.
```

laurapacilio (approved):
Just some wording nits - thanks!

PR #2069, opened 2022-01-06
Cloud agent parallelism

aibadger:
Parallelism flag only able to be set with the TF_CLI_ARGS_plan and TF_CLI_ARGS_apply vars. If parallelism flag set with TF_CLI_ARGS, Terraform init fails. Updated doc to be more specific about setting the parallelism flag for Cloud agent's using these TF_CLI_ARGS_name variables.


laurapacilio (approved):
Thank you! :)

sudomateo:
Yep! For future readers, this is because `TF_CLI_ARGS` provides arguments to _all_ Terraform commands. Not all Terraform commands support the `-parallelism` argument.

PR #2070, opened 2022-01-06
[DO NOT MERGE] Upgrade to @hashicorp/platform-nextjs-plugin v2.0.0

dstaley:
This PR upgrades us to the latest version of `@hashicorp/platform-nextjs-plugin`, which breaks out the PostCSS config.


dstaley:
`@hashicorp/platform-nextjs-plugin` v2 is released. We'll be upgrading to it with the Next.js v12 upgrade in the future.

PR #2071, opened 2022-01-06
Fix Attribute Validator example

reinoudk:
This updates the Attribute Validator example so it takes the length form the `.Value` field, and not the whole struct, which would cause an error.


bflad (approved):
Thank you for this fix, @reinoudk 🚀

PR #2072, opened 2022-01-07
Updating recommended Arch to include Darwin/ARMv8

Oechiih:
Go 1.16 has been released for a while and most providers seem to already support Darwin / ARMv8 anyways. So to follow what's currently displayed on the live website, Darwin / ARMv8 should now be a first class citizen and recommended.


bflad (approved):
Yes it should, thank you for this, @Oechiih 🚀

PR #2073, opened 2022-01-07
VCS BitBucket Server: using custom and generated keys

omarismail:
Recent work for BitBucket server requires some documentation change.

For BitBucket Server, the UI has changed to allow the user to either 1) use pre-generated keys, or 2) use custom keys they provide. The documentation updates to reflect that.


content/cloud-docs/vcs/bitbucket-server.mdx
```diff
@@ -76,11 +76,21 @@ Leave the page open in a browser tab, and remain logged in as an admin user.
...
 
-   ![Terraform Cloud screenshot: Consumer key and public key](/img/docs/bitbucket-server-tfe-consumer-and-public-key.png)
+A) Leave this page open in a browser tab. In the next step, you will copy and paste the unique **Consumer Key** and **Public Key.**
+
+   ![Terraform Cloud screenshot: Consumer key and public key](/img/docs/bitbucket-server-tfe-generated-consumer-and-public-key.png)
+
+B) Insert the **Consumer Key**, **Public Key**, and **Private Key** into this
+form that matches the Application Link that already exists.
+
+   ![Terraform Cloud screenshot: Consumer key and public key](/img/docs/bitbucket-server-tfe-custom-consumer-and-public-key.png)
```
laurapacilio, line 93: Hi! Thank you for updating this :-) I was thinking about the best way to present these options to the user. And I'm wondering if we could do something like this:

Click "Continue" to go to the Configure settings step. Do one of the following to set the Consumer Key and Public Key.
    - To use generated default keys, leave this page open in your web browser so that you can copy and paste the Consumer Key and the Public Key into your Bitbucket Server Application.
    - To use custom keys, click the toggle for "Use Custom Keys" and then enter the Consumer Key, Public Key, and Private key into the associated fields.
    
    
  What do you think of that? Also, I'm wondering - when do we tell folks to click "Continue" for the final time?  And I'm also wondering if we need both screenshots? Or to be honest if we even need a screenshot at all here? I'll leave that up to you, but I think we likely don't need both. Maybe if we keep one we keep the one with the custom key fields.

laurapacilio (changes requested):
Hi,

A couple of thoughts for you about how to present this information - see what you think! Also, I think we might want to update the beginning of this page - In my version of the "settings" it just says "Providers" in the sidebar, not "VCS Providers" :-)

content/cloud-docs/vcs/bitbucket-server.mdx
omarismail: Yeah I think your write about both screen shots, and only having one. I also think its okay to go with the common approach (which is to use generated keys), and then mention the alternative approach.

omarismail:
> In my version of the "settings" it just says "Providers" in the sidebar, not "VCS Providers" :-)

Updated this

content/cloud-docs/vcs/bitbucket-server.mdx
omarismail: This is the new text

image

content/cloud-docs/vcs/bitbucket-server.mdx
```diff
@@ -76,11 +76,16 @@ Leave the page open in a browser tab, and remain logged in as an admin user.
 
    ![Terraform Cloud screenshot: text fields for adding a Bitbucket Server VCS provider](/img/docs/bitbucket-server-tfe-add-url-fields.png)
 
-1. Click "Create VCS Provider." This will take you back to the VCS Provider page, which now includes your new Bitbucket Server client.
+1. Click "Continue", which will allow you to use the **Consumer Key** and
+**Public/Private Keys** when creating a new Application Link in BitBucket
+Server. Leave this page open in a browser tab. In the next step, you will
+copy and paste the unique **Consumer Key** and **Public Key.**
```
laurapacilio, line 82:
```suggestion
1. Click "Continue" to view the generated **Consumer Key**, **Public Key**, and **Private Key** you can use to create a new Application Link in BitBucket Server. Leave this page open so you can copy and paste these values in Step 2.
```

content/cloud-docs/vcs/bitbucket-server.mdx
```diff
@@ -76,11 +76,16 @@ Leave the page open in a browser tab, and remain logged in as an admin user.
...
 
-1. Leave this page open in a browser tab. In the next step, you will copy and paste the unique **Consumer Key** and **Public Key.**
+**If you have your own _Consumer Key_ and _Public/Private Key_ from an existing
+Application Link, then toggle the _Use Custom Keys_ and insert them.**
```
laurapacilio, line 85: Can we just make sure this is indented under the step? 

```suggestion
    To use keys from an existing Application Link, toggle "Use Custom Keys" and enter them into the page.
```

laurapacilio (approved):
A couple final wording suggestions but I'll preapprove so I don't block.  Thanks!

PR #2079, opened 2022-01-11
Upgrade to Next.js v12 🎉

dstaley:
This PR upgrades `terraform.io` from Next.js v11 to v12! :tada:


BRKalow (commented):
Looks like it's already merged 😄, but all looks good

PR #2082, opened 2022-01-12
Use Next.js v12 middleware for /docs redirects

dstaley:
This PR uses the newly available Next.js v12 Edge Middleware to perform `/docs` redirects (instead of using a `'blocking'` fallback for page generation). It also adjusts the Redirects section of the README to more accurately reflect how redirects are defined.


README.md
```diff
@@ -405,62 +410,56 @@ You may customize the parameters in any way you'd like. To remove a prerelease f
...
 
-<!-- BEGIN: redirects -->
-<!-- Generated text, do not edit directly -->
-
 ## Redirects
 
 This website structures URLs based on the filesystem layout. This means that if a file is moved, removed, or a folder is re-organized, links will break. If a path change is necessary, it can be mitigated using redirects. It's important to note that redirects should only be used to cover for external links -- if you are moving a path which internal links point to, the internal links should also be adjusted to point to the correct page, rather than relying on a redirect.
 
-To add a redirect, head over to the `redirects.js` file - the format is fairly simple - there's a `source` and a `destination` - fill them both in, indicate that it's a permanent redirect or not using the `permanent` key, and that's it. Let's look at an example:
+To add a redirect, head over to the `redirects.next.js` file - the format is fairly simple - there's a `miscRedirectsMap` containing source keys and destination values - fill them both in, and that's it. Let's look at an example:
```
laurapacilio, line 417:
```suggestion
To add a redirect, head over to the `redirects.next.js` file. It has a `miscRedirectsMap` containing source keys and destination values. Enter both the source key and the destination value for every redirect you want to add and save the file. Let's look at an example:
```

README.md
```diff
@@ -405,62 +410,56 @@ You may customize the parameters in any way you'd like. To remove a prerelease f
...
 
-Let's look at an example. Say you have a page called `/docs/foo` which needs to be moved to `/docs/nested/foo`. Additionally, this is a page that has been around for a while and we know there are links into `/docs/foo.html` left over from our previous website structure. First, we move the page, then adjust the docs sidenav, in `data/docs-navigation.js`. Find the category the page is in, and move it into the appropriate subcategory. Next, we add to `_redirects` as such. The `.html` version is covered automatically.
+Let's look at an example. Say you have a page called `/language/foo` which needs to be moved to `/language/nested/foo`. Additionally, this is a page that has been around for a while and we know there are links into `/language/foo.html` left over from our previous website structure. First, we move the page, then adjust the docs sidenav, in `data/language-navigation.js`. Find the category the page is in, and move it into the appropriate subcategory. Next, we add to `miscRedirectsMap` as such.
```
laurapacilio, line 428:
```suggestion
Let's look at an example. Say you have a page called `/language/foo` which needs to be moved to `/language/nested/foo`. Additionally, this is a page that has been around for a while and we know there are links into `/language/foo.html` left over from our previous website structure. First, you would move the page to the correct directory and then adjust the docs sidenav in `data/language-navigation.js` to reflect the new structure.  Next, you would add to `miscRedirectsMap` (example below).
```

README.md
```diff
@@ -405,62 +410,56 @@ You may customize the parameters in any way you'd like. To remove a prerelease f
...
 
 ```js
-{ source: '/foo', destination: '/nested/foo', permanent: true }
+const miscRedirectsMap = {
+  '/language/foo': '/language/nested/foo',
+  '/language/foo.html': '/language/nested/foo'
 ```
 
-Next, we run a global search for internal links to `/foo`, and make sure to adjust them to be `/nested/foo` - this is to ensure that client-side navigation still works correctly. _Adding a redirect alone is not enough_.
+Next, we run a global search for internal links to `/language/foo`, and make sure to adjust them to be `/language/nested/foo` - this is to ensure that client-side navigation still works correctly. _Adding a redirect alone is not enough_.
```
laurapacilio, line 436:
```suggestion
Next, you would run a global search for internal links to `/language/foo`, and make sure to adjust them to be `/language/nested/foo`. This is to ensure that client-side navigation still works correctly. _Adding a redirect alone is not enough_.
```

README.md
```diff
@@ -405,62 +410,56 @@ You may customize the parameters in any way you'd like. To remove a prerelease f
...
 
-One more example - let's say that content is being moved to an external website. A common example is guides moving to `learn.hashicorp.com`. In this case, we take all the same steps, except that we need to make a different type of change to the `docs-navigation` file. If previously the structure looked like:
+One more example - let's say that content is being moved to an external website. A common example is guides moving to `learn.hashicorp.com`. In this case, we take all the same steps, except that we need to make a different type of change to the `language-navigation` file. If previously the structure looked like:
```
laurapacilio, line 438:
```suggestion
One more example - let's say that content is being moved to an external website. A common example is guides moving to `learn.hashicorp.com`. In this case, you would take all the same steps, but make a different type of change to the `language-navigation` file. Previously the structure looked like:
```

README.md
```diff
@@ -405,62 +410,56 @@ You may customize the parameters in any way you'd like. To remove a prerelease f
...
-  content: [
-    'foo'
+  "title": "Language",
+  "routes": [
+    { "title": "Foo", "path": "foo" }
   ]
 }
 ```
 
 If we no longer want the link to be in the side nav, we can simply remove it. If we do still want the link in the side nav, but pointing to an external destination, we need to slightly change the structure as such:
```
laurapacilio, line 449: change this from "we" to "you"

README.md
```diff
@@ -405,62 +410,56 @@ You may customize the parameters in any way you'd like. To remove a prerelease f
...
   ]
 }
 ```
 
 As the majority of items in the side nav are internal links, the structure makes it as easy as possible to represent these links. This alternate syntax is the most concise manner than an external link can be represented. External links can be used anywhere within the docs sidenav.
 
-It's also worth noting that it is possible to do glob-based redirects, for example matching `/docs/*`, and you may see this pattern in the redirects file. This type of redirect is much higher risk and the behavior is a bit more nuanced, so if you need to add a glob redirect, please reach out to the website maintainers and ask about it first.
-
-<!-- END: redirects -->
+It's also worth noting that it is possible to do glob-based redirects, for example matching `/nested/*`, and you may see this pattern in the redirects file. This type of redirect is much higher risk and the behavior is a bit more nuanced, so if you need to add a glob redirect, please reach out to the website maintainers and ask about it first.
```
laurapacilio, line 462:
```suggestion
It's also worth noting that it is possible to do glob-based redirects (e.g., matching `/nested/*`), and you may see this pattern in the redirects file. This type of redirect is much higher risk and the behavior is a bit more nuanced. If you need to add a glob redirect, please contact the website maintainers and ask about it first.
```

laurapacilio (approved):
A couple of language nits, but approving so I don't block. Thank you for your commitment to updating this for us :-)

PR #2083, opened 2022-01-12
Update installer.mdx

zisom-hc:
Update the name of the airgap installer bootstrapper file that's downloaded at https://install.terraform.io/airgap/latest.tar.gz from `latest.tar.gz` to `replicated.tar.gz`


laurapacilio (approved):
Thank you!

PR #2090, opened 2022-01-21
Corrects function receiver name

gdavison:
In the `stringDefaultModifier` `Description` and `MarkdownDescription` functions, the receiver name was not consistent with the usage in the function. The sample code does not compile as-is.


bflad (approved):
Thank you for this fix, @gdavison 🚀

PR #2091, opened 2022-01-21
Update index.mdx

jisorondo:
Updating Option 3: Docker Engine Using a Modified libseccomp Profile, step 4 to note that RHEL 7.x with docker 1.13.1 already has the `/etc/docker/seccomp.json` and only requires to modify the `defaultAction` line and restarting docker.

### DIFF
--- a/content/enterprise/before-installing/index.mdx
+++ b/content/enterprise/before-installing/index.mdx
@@ -217,7 +217,7 @@ Refer to the sections below for instructions on how to verify your Docker Engine
      https://raw.githubusercontent.com/moby/moby/master/profiles/seccomp/default.json
    ```
 
-1. Edit the `/etc/docker/seccomp.json` file and modify the line `"defaultAction": "SCMP_ACT_ERRNO",` to be `"defaultAction": "SCMP_ACT_TRACE",`. A `sed` command is included below for your convenience.
+1. Edit the `/etc/docker/seccomp.json` file and modify the line `"defaultAction": "SCMP_ACT_ERRNO",` to be `"defaultAction": "SCMP_ACT_TRACE",`. A `sed` command is included below for your convenience. For Docker Engine 1.13.1 (RHEL only), as the `/etc/docker/seccomp.json` file already exists you can proceed with restarting the docker engine after modifying the aforementioned line by running `sudo systemctl restart docker`.
 
    ```
    sudo sed -i 's/"defaultAction":\s*"SCMP_ACT_ERRNO"/"defaultAction": "SCMP_ACT_TRACE"/1' /etc/docker/seccomp.json
--- a/content/enterprise/before-installing/index.mdx
+++ b/content/enterprise/before-installing/index.mdx
@@ -141,34 +141,32 @@ Refer to the sections below for instructions on how to verify your Docker Engine
 #### Option 1: Docker Engine With a Compatible `runc` Version
 
 1. [Install](https://docs.docker.com/engine/install/) Docker Engine 17.06.2-ce, 18.09.x, 19.03.x, or 20.10.x for your operating system.
+
 1. Install the latest version of `containerd` for your operating system.
 
-- On Debian/Ubuntu:
+  On Debian/Ubuntu:
 
-  ````
-  ```
+  ```sh
   sudo apt install containerd
   ```
-  ````
-- On RHEL/CentOS:
 
-  ````
-  ```
+  On RHEL/CentOS:
+
+  ```sh
   sudo yum install containerd.io
   ```
-  ````
 
 1. Confirm that the installed `containerd` version is 1.4.9, 1.5.5, or greater.
 
-   ```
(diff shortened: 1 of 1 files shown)


content/enterprise/before-installing/index.mdx
```diff
@@ -217,7 +217,7 @@ Refer to the sections below for instructions on how to verify your Docker Engine
      https://raw.githubusercontent.com/moby/moby/master/profiles/seccomp/default.json
    ```
 
-1. Edit the `/etc/docker/seccomp.json` file and modify the line `"defaultAction": "SCMP_ACT_ERRNO",` to be `"defaultAction": "SCMP_ACT_TRACE",`. A `sed` command is included below for your convenience.
+1. Edit the `/etc/docker/seccomp.json` file and modify the line `"defaultAction": "SCMP_ACT_ERRNO",` to be `"defaultAction": "SCMP_ACT_TRACE",`. A `sed` command is included below for your convenience. For Docker Engine 1.13.1 (RHEL only), as the `/etc/docker/seccomp.json` file already exists you can proceed with restarting the docker engine after modifying the aforementioned line by running `sudo systemctl restart docker`.
```
sudomateo, line 220: Two steps above we have this:

```
1. Check if the file `/etc/docker/seccomp.json` exists. If it does, proceed to step 4.
```

Are you saying that on RHEL with Docker Engine 1.13.1 we don't need to create the drop-in systemd unit file because the `--seccomp-profile=/etc/docker/seccomp.json` option is already present in the default systemd unit file?

sudomateo (commented):
Left a comment! Thank you for working on this.

content/enterprise/before-installing/index.mdx
jisorondo: Correct, here is the content for reference:

[Unit]
Description=Docker Application Container Engine
Documentation=http://docs.docker.com
After=network.target rhel-push-plugin.service registries.service
Wants=docker-storage-setup.service
Requires=rhel-push-plugin.service registries.service
Requires=docker-cleanup.timer

[Service]
Type=notify
NotifyAccess=main
EnvironmentFile=-/run/containers/registries.conf
EnvironmentFile=-/etc/sysconfig/docker
EnvironmentFile=-/etc/sysconfig/docker-storage
EnvironmentFile=-/etc/sysconfig/docker-network
Environment=GOTRACEBACK=crash
Environment=DOCKER_HTTP_HOST_COMPAT=1
Environment=PATH=/usr/libexec/docker:/usr/bin:/usr/sbin
ExecStart=/usr/bin/dockerd-current \
          --add-runtime docker-runc=/usr/libexec/docker/docker-runc-current \
          --default-runtime=docker-runc \
          --authorization-plugin=rhel-push-plugin \
          --exec-opt native.cgroupdriver=systemd \
          --userland-proxy-path=/usr/libexec/docker/docker-proxy-current \
          --init-path=/usr/libexec/docker/docker-init-current \
          --seccomp-profile=/etc/docker/seccomp.json \
          $OPTIONS \
          $DOCKER_STORAGE_OPTIONS \
          $DOCKER_NETWORK_OPTIONS \
          $ADD_REGISTRY \
          $BLOCK_REGISTRY \
          $INSECURE_REGISTRY \
	  $REGISTRIES
ExecReload=/bin/kill -s HUP $MAINPID
LimitNOFILE=1048576
LimitNPROC=1048576
LimitCORE=infinity
TimeoutStartSec=0
Restart=on-abnormal
KillMode=process

[Install]
WantedBy=multi-user.target

laurapacilio, line 220: What if we tried something like this? 

I'm thinking we can reduce the text a bit if we do this because now it's more obvious that the command is going to do what we just told them to do. And then, I think it makes sense to move the Docker note after the `sed` command since a) it doesn't apply to everyone and b) it's something folks should only do after they run that command to update the `seccomp.json` file. Thoughts?  We'd need to test the build to make sure everything renders properly (I know sometimes markdown is finicky with formatting).

```suggestion
1. In the `/etc/docker/seccomp.json` file, change `"defaultAction": "SCMP_ACT_ERRNO",` to `"defaultAction": "SCMP_ACT_TRACE",`. 
     
      ```
       sudo sed -i 's/"defaultAction":\s*"SCMP_ACT_ERRNO"/"defaultAction": 
      "SCMP_ACT_TRACE"/1' /etc/docker/seccomp.json
       ```
    **Docker Engine 1.13.1 (RHEL only):**  After modifying the `/etc/docker/seccomp.json` file, run `sudo systemctl restart docker` to restart the Docker engine.
```

laurapacilio (approved):
Approving so I don't block you, but just leaving a quick formatting/flow note for you :) see what you think!

content/enterprise/before-installing/index.mdx
sudomateo: Love it! Thank you so much!

PR #2105, opened 2022-01-26
fix: you can't actually update the key text for an existing SSH key

brandonc:
From the very beginning it's not been possible to update the value of an SSH key text. The reason for this is the non-repudiation of security auditing. For example, if an attacker uses the ssh key to access a resource, the security audit of the ssh authentication can be trusted to mean key X was always key X and never updated.


laurapacilio (approved):
LGTM - thanks!

PR #2110, opened 2022-01-27
Remove Clustered documentation and add redirects

laurapacilio:
This PR:
- Removes documentation pages about clustered installation reference architectures as this mode is no longer supported. 
-  Adds notes on both the reference architectures overview page and the Active/Active architectures page to let users know that clustering is no longer supported and that we recommend migrating.
- Adds redirects for all of the clustering pages to the Active/Active architecture page.
- Updates the enterprise navigation file to remove the deleted pages so that our site doesn't break :-)


redirects.next.js
```diff
@@ -181,6 +181,11 @@ module.exports = (async () => {
     '/upgrade-guides/1-0.html': '/language/upgrade-guides/1-0',
     '/upgrade-guides/index.html': '/language/upgrade-guides',
     '/language/resources/provisioners': '/language/resources/provisioners/syntax',
+    '/enterprise/before-installing/cluster-architecture': '/enterprise/admin/infrastructure/active-active',
+    '/enterprise/install/cluster-aws': '/enterprise/admin/infrastructure/active-active',
+    '/enterprise/install/cluster-azure': '/enterprise/admin/infrastructure/active-active',
+    'content/enterprise/install/cluster-custom': '/enterprise/admin/infrastructure/active-active',
```
dstaley, line 187: @laurapacilio Looks like this line and the one below have a rogue `content` prefix :)

laurapacilio: heh thank you - fixed!!

content/enterprise/admin/infrastructure/active-active.mdx
```diff
@@ -6,6 +6,8 @@ page_title: Active/Active - Infrastructure Administration - Terraform Enterprise
 
 When your organization requires increased reliability or performance from Terraform Enterprise that your current single application instance cannot provide, it is time to scale to the Active/Active architecture. Please refer to  [Terraform Enterprise Active/Active](/enterprise/install/automated/active-active) for more information on moving to Active/Active.
 
+~> **Note:** Clustering is no longer supported and we strongly advise all customers to move to an Active/Active installation. Please contact your support representative if you need assistance.
```
bnferguson, line 9: It may be worth clarifying that this was the discontinued Kubernetes based Clustering product not just clustering as a concept (I mean that's also not supported but it feels a little ambiguous here if you didn't come from a redirect)

laurapacilio: Good point! I pushed up a change that is hopefully clearer about what it is that is discontinued and no longer supported. I found similar language on one of the deleted pages, so this seems like a good way to refer to it. @bnferguson Do you feel comfortable approving now?

bnferguson (approved):
Excellent! 🎉  Thanks so much for doing this.

PR #2112, opened 2022-01-28
Use SDK v2 imports in customizing differences section

timofurrer:


bflad (approved):
Looks good to me, thanks @timofurrer 🚀

PR #2114, opened 2022-01-28
Use SDK v2 imports in Resources - State Migration chapter examples

timofurrer:


bflad (approved):
Looks good to me, thank you @timofurrer 🚀

PR #2115, opened 2022-01-28
Update installer.mdx

jisorondo:
Adding a note to mention default ciphersuites used by nginx  HIGH:!aNULL:!MD5

### DIFF
--- a/content/enterprise/install/interactive/installer.mdx
+++ b/content/enterprise/install/interactive/installer.mdx
@@ -159,6 +159,8 @@ Terraform Enterprise uses a standard set of ciphersuites by default. If necessar
 
 (This value must be defined in the OpenSSL [cipher list format](https://www.openssl.org/docs/man1.1.1/man1/ciphers.html)
 
+~> Note: By default enabled ciphers for Nginx are `HIGH:!aNULL:!MD5`.
+
 #### Force TLS with HSTS
 
 Terraform Enterpise may be configured to force TLS via the enabling of HTTP Strict Transport Security:
--- a/content/enterprise/install/interactive/installer.mdx
+++ b/content/enterprise/install/interactive/installer.mdx
@@ -159,7 +159,7 @@ Terraform Enterprise uses a standard set of ciphersuites by default. If necessar
 
 (This value must be defined in the OpenSSL [cipher list format](https://www.openssl.org/docs/man1.1.1/man1/ciphers.html)
 
--> Note: The default enabled ciphers for Nginx are `HIGH:!aNULL:!MD5`.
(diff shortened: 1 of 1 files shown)


content/enterprise/install/interactive/installer.mdx
```diff
@@ -159,6 +159,8 @@ Terraform Enterprise uses a standard set of ciphersuites by default. If necessar
 
 (This value must be defined in the OpenSSL [cipher list format](https://www.openssl.org/docs/man1.1.1/man1/ciphers.html)
 
+~> Note: By default enabled ciphers for Nginx are `HIGH:!aNULL:!MD5`.
```
laurapacilio, line 162: I know we're a bit inconsistent about notes (I promise fixes for this are coming!) but I'm adjusting your callout formatting so that it matches the majority of other notes in the Terraform docs right now.  Also a tiny wording nit :-) 

```suggestion
-> Note: The default enabled ciphers for Nginx are `HIGH:!aNULL:!MD5`.
```

laurapacilio (approved):
thank you!

content/enterprise/install/interactive/installer.mdx
jisorondo: Thank you, looks great!

content/enterprise/install/interactive/installer.mdx
```diff
@@ -159,6 +159,8 @@ Terraform Enterprise uses a standard set of ciphersuites by default. If necessar
 
 (This value must be defined in the OpenSSL [cipher list format](https://www.openssl.org/docs/man1.1.1/man1/ciphers.html)
 
+-> Note: The default enabled ciphers for Nginx are `HIGH:!aNULL:!MD5`.
```
chair6, line 162:
```suggestion
-> Note: The default enabled ciphers for Terraform Enterprise are `HIGH:!aNULL:!MD5`.
```

We don't mention Nginx anywhere else in this documentation.. I think probably best just to refer to Terraform Enterprise here, rather than exposing that level of detail.

chair6 (commented):
Useful addition! Just one minor nit about referring to Terraform Enterprise instead of Nginx, which I'd consider to be an implementation detail that we don't need to expose.

chair6 (approved):
LGTM, 🚀 !

sudomateo (changes requested):
Instead of adding the note, I'd recommend putting this information in the beginning sentence of this section like so:

```
Terraform Enterprise uses the `HIGH:!aNULL:!MD5` ciphersuites by default. If necessary, a custom ciphersuite can be provided:
```

PR #2117, opened 2022-01-29
Reorganize Workspaces

anamahalo:
TBD

Please reach out to @laurapacilio / @judithpatudith / @andreadaniels1 for info.


content/cloud-docs/api-docs/_workspaces_bak.mdx
```diff
@@ -1,4 +1,4 @@
----
+<!-- ---
```
anamahalo, line 1: This file to be deleted when review is 👍

content/cloud-docs/api-docs/changelog.mdx
```diff
@@ -15,6 +15,10 @@ page_id: api-changelog
 
 Keep track of changes to the API for Terraform Cloud and Terraform Enterprise.
 
+### 2022-MM-DD
+
+* TBD
```
anamahalo, line 20: Reminder to self: Update pre-merge

data/docs-redirects.js
```diff
@@ -180,9 +180,15 @@ export const docsRedirects = {
...
   '/docs/cloud/api/vcs-events.html': '/cloud-docs/api-docs/vcs-events',
-  '/docs/cloud/api/workspace-resources.html': '/cloud-docs/api-docs/workspace-resources',
-  '/docs/cloud/api/workspace-variables.html': '/cloud-docs/api-docs/workspace-variables',
-  '/docs/cloud/api/workspaces.html': '/cloud-docs/api-docs/workspaces',
+  '/docs/cloud/api/workspaces/index.html': '/cloud-docs/api-docs/workspaces/index',
+  '/docs/cloud/api/worksapces/workspace-crud-actions.html': '/cloud-docs/api-docs/workspaces/workspace-crud-actions',
+  '/docs/cloud/api/worksapces/workspace-lock-mechanism.html': '/cloud-docs/api-docs/workspaces/workspace-lock-mechanism',
+  '/docs/cloud/api/worksapces/workspace-remote-state-consumers.html': '/cloud-docs/api-docs/workspaces/workspace-remote-state-consumers',
+  '/docs/cloud/api/worksapces/workspace-resources.html': '/cloud-docs/api-docs/workspaces/workspace-resources',
+  '/docs/cloud/api/worksapces/workspace-ssh-key-assignment.html': '/cloud-docs/api-docs/workspaces/workspace-ssh-key-assignment',
```
laurapacilio, line 188: you've got a couple lines here where there is a typo with "worksapces" :-)

data/cloud-docs-nav-data.json
```diff
@@ -378,20 +383,49 @@
...
+          {
+            "title": "Workspace-Specific Variables",
+            "path": "api-docs/workspaces/workspace-variables"
+          },
+          {
+            "title": "Resources",
+            "path": "api-docs/workspaces/workspace-resources"
+          },
+          {
+            "title": "Appendix",
```
laurapacilio, line 424: I know "Related Resources" has been handled somewhat inconsistently throughout the API docs, and this is also the first time we've had to really handle it with separated out pages on the same topic like this. But I think the way we're doing it could be clearer. "Appendix" as the name for this page doesn't make it super clear what users will get when they come to this page (related resources). 

I had two thoughts - lmk what you think. 
- Can we rename this page to "Endpoint Related Resources" or something similar? I want folks to understand what they're getting when they click.
- I wonder if at the bottom of each page, we should add a section called "Related Resources" and then point folks toward the common page. This way, we're notifying folks that there ARE related resources, but not repeating all of the content on the common page. (example below)

## Related Resources

Workspace GET endpoints can optionally return related resources when requested with the `include` query parameter. Refer to [Endpoint Related Resources](link) for a full list.

content/cloud-docs/api-docs/workspaces/workspace-remote-state-consumers.mdx
```diff
@@ -0,0 +1,400 @@
...
+[504]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504
+
+[JSON API document]: /cloud-docs/api-docs#json-api-documents
+
+[JSON API error object]: https://jsonapi.org/format/#error-objects
+
+[speculative plans]: /cloud-docs/run/#speculative-plans
+
+# Workspace Remote State Consumers API
+
```
laurapacilio, line 40:
```suggestion
# Workspace Remote State Consumers API

Workspaces within an organization can [share their state](/cloud-docs/workspaces/settings#remote-state-sharing) with other workspaces during Terraform Cloud runs.
```

content/cloud-docs/api-docs/workspaces/workspace-ssh-key-assignment.mdx
```diff
@@ -0,0 +1,405 @@
...
+[504]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504
+
+[JSON API document]: /cloud-docs/api-docs#json-api-documents
+
+[JSON API error object]: https://jsonapi.org/format/#error-objects
+
+[speculative plans]: /cloud-docs/run/#speculative-plans
+
+# Workspace SSH Assignment API
+
```
laurapacilio, line 40:
```suggestion
# Workspace SSH Assignment API

To access a private Git repository, Terraform either needs login credentials (for HTTPS access) or an [SSH key](/cloud-docs/workspaces/settings/ssh-keys). Terraform Cloud manages SSH keys used to clone Terraform modules at the organization level, and allows multiple keys to be added for the organization. 
```

content/cloud-docs/api-docs/workspaces/workspace-tags.mdx
```diff
@@ -0,0 +1,235 @@
...
+
+[JSON API document]: /cloud-docs/api-docs#json-api-documents
+
+[JSON API error object]: https://jsonapi.org/format/#error-objects
+
+[speculative plans]: /cloud-docs/run/#speculative-plans
+
+# Workspace Tags API
+
+## Get Tags
```
laurapacilio, line 41: Please check my permissions statement. But I think we should on each page give any needed permissions information for the endpoints on that page. 

```suggestion
# Workspace Tags API

Tags let you more easily [categorize, sort, and filter](/cloud-docs/workspaces#listing-and-filtering-workspaces) workspaces in Terraform Cloud. Managing tags requires permissions to read runs on the workspace. Refer to [Permissions](/cloud-docs/users-teams-organizations/permissions) for more details.

## Get Tags
```

laurapacilio (changes requested):
Hello! I am very excited to see this up 🥳   A couple of initial thoughts for you:
- Each page should have page description metadata. It's important for SEO!
- At the top of each page, let's include a tiny bit of context for users about the feature/concept that all of the endpoints are about, and do an in-text link to relevant docs pages where they could learn more about the concepts if needed. 
- In the intro section, let's include any relevant permissions information. We can't assume that users will look at an overview page, so imo each page should have a sentence or two about permissions (in the same place, for consistency across pages).
- I have some thoughts about how to handle the related resources. This is a new thing we need to tackle, so see what you think of the suggestions I left. We can work on it together to figure out what feels right.

Now that I see this, I do agree that CRUD is still a bit long BUT I think this is okay to keep for now because a) new designs are coming!  b) Our next project can be figuring out how to truncate responses so that they aren't as long and c) I think separating each CRUD action out onto it's own page is unnecessary and will lead to very short pages. I think the amount of scrolling has been really reduced, and with the other changes I listed above, leaving them on one page makes sense. 

Anyway, see what you think of this initial round of comments, and then I'd love to take a look at this again!

data/docs-redirects.js
anamahalo: 🤦‍♀️ Thanks!

data/cloud-docs-nav-data.json
anamahalo: I actually think this is a really good idea TBH.  I had the same thought but didn't want to change the standardization on this.  I think this is absolutely reasonable and would make it easier for users to navigate throughout the _Workspaces_ stuff.  👍

anamahalo: Oh, just of note -- I'll update this, but the only thing of note is that at least on the SideNav it might cause some confusion with plain ol' _Resources_ for workspaces so I think I'm going to make sure the title is "API Related Resources" for the SideNav (since it's nested), but on the page itself it will be "Workspaces API Related Resources" 👍

laurapacilio:
We're migrating all of the content in the cloud-docs directory to a new repository. To support the migration, there will be a docs code freeze on Wednesday May 25th where we'll ask folks not to merge any PRs for the TFC docs. I'm going to close this PR now and then w re-open it and recreate these changes in the new repository. That way, we don't have to rush the reviews on this large content change.

This in no way means we are not moving forward with this work. This is just to help prepare for the code freeze and content migration of this content.

Thank you!

PR #2121, opened 2022-01-31
Add log rotation information

sudomateo:
Operators have been experiencing increased disk utilization since Terraform Enterprise v202109-1. This is because the default `log-opts` for the `json-file` Docker logging driver were removed as a part of the log forwarding work. These changes detail how an operator can configure log rotation to limit disk utilization regardless whether or not they are using the log forwarding feature.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201747281670986

### DIFF
--- a/content/enterprise/admin/infrastructure/logging.mdx
+++ b/content/enterprise/admin/infrastructure/logging.mdx
@@ -9,7 +9,7 @@ destinations, a process called log forwarding. Log forwarding provides increased
 observability, assistance complying with log retention requirements, and
 information during troubleshooting.
 
--> **Note:** If you are using a version of Terraform Enterprise prior to v202109-1, please refer to the [previous version of this page](/enterprise/admin/logging-old) with logspout information.
+-> **Note:** For Terraform Enterprise versions prior to v202109-1, please refer to the [previous version of this page](/enterprise/admin/logging-old) with logspout information.
 
 ## Requirements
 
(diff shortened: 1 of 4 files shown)


amy-hashi (approved):

laurapacilio (changes requested):
Hi! Leaving some initial thoughts and then per our conversation, we think in cases where the two possibilities are short and very straightforward like this,  it makes the most sense not to use tabs. If you could try revising it without tabs, I will take a final pass and approve :)

content/enterprise/admin/infrastructure/logging.mdx
```diff
@@ -334,4 +334,28 @@ recommend forwarding all Terraform Enterprise logs to a log aggregation system,
...
 the audit logs to the desired destination.
 
--> **Note:** If you are using a version of Terraform Enterprise prior to v202109-1, please refer to the [previous version of this page](/enterprise/admin/logging-old) with logspout information.
+## Log Rotation
+
+Log forwarding uses the `journald` Docker logging driver to send Terraform
+Enterprise logs to `systemd-journald`. This can cause increased disk utilization
+for the `/var/log/journal` path.
+
+To limit this disk utilization, configure the `SystemMaxFileSize` and
```
laurapacilio, line 343: Beautifully done!  Two thoughts:
- What do we mean by "reasonable values"? I wonder if we could say something more specific OR simplify to say "To limit disk use, set the `SystemMaxFileSize` and `SystemMaxFiles` settings within `etc/systemd/journald.conf`.
- I would move the explanation of what the example does to before the example. So I'd say "The following example tells..."   I think it's nice to tell folks what they're about to see before they see it.

content/enterprise/admin/infrastructure/logging.mdx
```diff
@@ -70,28 +70,28 @@ the `fluent-bit.conf` file.
...
-    ...
+```json
     "log_forwarding_config": {
         "value": "# Match all logs and do not forward them anywhere.\n[OUTPUT]\n    Name null\n    Match *\n"
     },
-    ...
 ```
 
-```
+Here's what that escaped JSON string will render to:
```
laurapacilio, line 92:
```suggestion
That escaped JSON string renders to the following:
```

laurapacilio (approved):
Pre-approving so I don't block! Just a couple tiny nits/suggestions :) Otherwise, LGTM!

content/enterprise/admin/infrastructure/logging.mdx
sudomateo: I like it! Thank you so much!

PR #2132, opened 2022-02-03
Update index.mdx

jisorondo:
Adding requirement when using IMDSv2 to configure the metadata options and set the PUT http response hop limit to 2

### DIFF
--- a/content/enterprise/before-installing/index.mdx
+++ b/content/enterprise/before-installing/index.mdx
@@ -335,7 +335,7 @@ At a minimum, Terraform Enterprise will require the following permissions if the
 
 #### Instance Profile as Default Credentials
 
-You can use Terraform Enterprise's instance profile to provide default credentials to workspaces. Terraform will attempt to use the instance profile to provision resources when you do not set credentials as environment variables. However, this approach presents a few security risks:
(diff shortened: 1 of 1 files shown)


content/enterprise/before-installing/index.mdx
```diff
@@ -335,7 +335,7 @@ At a minimum, Terraform Enterprise will require the following permissions if the
 
 #### Instance Profile as Default Credentials
 
-You can use Terraform Enterprise's instance profile to provide default credentials to workspaces. Terraform will attempt to use the instance profile to provision resources when you do not set credentials as environment variables. However, this approach presents a few security risks:
+You can use Terraform Enterprise's instance profile to provide default credentials to workspaces. Configure the [instance metadata options](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html) and specify the PUT response hop limit value as 2 for instances using IMDSv2. Terraform will attempt to use the instance profile to provision resources when you do not set credentials as environment variables. However, this approach presents a few security risks:
```
sudomateo, line 338:
```suggestion
You can use Terraform Enterprise's instance profile to provide default credentials to workspaces. When using IMDSv2, configure the PUT response hop limit with a value of 2 within the [instance metadata options](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html). Terraform will attempt to use the instance profile to provision resources when you do not set credentials as environment variables. However, this approach presents a few security risks:
```

sudomateo (changes requested):

sudomateo (approved):

PR #2133, opened 2022-02-04
Remove unclear line around caching of public providers/modules in TFC

bethanyr:
We received feedback that this line makes it appear that public providers and modules are actually cached in Terraform Cloud Private Registry. Actually, the Private Registry will store a pointer to the module or provider in the public registry and behind the scenes the modules and providers are pulled from the public registry. I have removed the line from the documentation, but I'm open to suggestions on if there should be something to indicate that we are only storing a pointer to the public registry.

### DIFF
--- a/content/cloud-docs/registry/add.mdx
+++ b/content/cloud-docs/registry/add.mdx
@@ -11,7 +11,7 @@ description: >-
 
 # Adding Public Providers and Modules to the Private Registry
 
-You can add providers and modules from the public [Terraform Registry](/registry) to your Terraform Cloud private registry. This lets you clearly designate which public providers and modules are recommended for the organization and makes their supporting documentation and examples centrally accessible.
+You can add providers and modules from the public [Terraform Registry](/registry) to your Terraform Cloud private registry. The private registry stores a pointer to these public providers and modules so that you can view their data from within Terraform Cloud. This lets you clearly designate which public providers and modules are recommended for the organization and makes their supporting documentation and examples centrally accessible.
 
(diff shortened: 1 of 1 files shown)


content/cloud-docs/registry/add.mdx
```diff
@@ -17,8 +17,6 @@ You can add providers and modules from the public [Terraform Registry](/registry
 
 [permissions-citation]: #intentionally-unused---keep-for-maintainers
```
laurapacilio, line 19: My suggestion would be to update the first paragraph as follows:

"You can add providers and modules from the public [Terraform Registry](/registry) to your Terraform Cloud private registry. The private registry stores a pointer to these public providers and modules so that you can view their data from within Terraform Cloud. This lets you clearly designate which public providers and modules are recommended for the organization and makes their supporting documentation and examples centrally accessible."

Rather than removing the offending paragraph entirely, I think we should do something like what I have above because it helps explain how it actually works. I think if we remove it entirely we could get folks wondering if we are caching, or mirroring, or whatever.

laurapacilio (approved):
I'm going to approve so I don't block you here, as I know you might be OOO next week. But just thinking through how we can help users understand how it _really_ works.

laurapacilio:
P.S. If you need to merge this and I'm not around, ignore the Percy error. As long as the deploy builds successfully and you can see that the rendered page looks okay, you can merge.

content/cloud-docs/registry/add.mdx
bethanyr: Awesome - that makes sense to me. I'll make the change you are suggesting.

bethanyr: I just updated the wording based on your feedback.

PR #2134, opened 2022-02-07
Change wording for better understanding

pabloehc:


content/plugin/sdkv2/schemas/schema-behaviors.mdx
```diff
@@ -98,7 +98,7 @@ defined by `Elem` for complex types.
 - Cannot be used if `Required` is `true`
 - Cannot be used with `DefaultFunc`
 
-If `Default` is specified, that value that is used when this item is not set in
+If `Default` is specified, that is the value that is used when this item is not set in
 the configuration.
```
laurapacilio, line 102:
```suggestion
If `Default` is specified, Terraform will use that value when this item is not set in
the configuration.
```

laurapacilio (changes requested):
Thank you! :-) Suggesting an edit to make it even clearer with active voice. @bflad Could you take a look at this to make sure my suggested change is technically correct?

bflad (approved):
Looks good to me 🚀

PR #2143, opened 2022-02-17
Add clarity to release page around the Terraform CLI Version

jaredfholgate:
The existing release page does not make it clear whether the 'Lastest Terraform CLI' refers to the bundled version of the CLI or whether this is the maximum version that version of TFE supports. Customer feedback prompted adding clarity to the release page so it is clear what it means.


laurapacilio (approved):
Nice! Glad you made this fix!

PR #2146, opened 2022-02-18
correcting which keys are shown

kristinahunter:
Documentation advised that a private key would be shown. This is not true and is reflected in the screenshots. Removed mention of private key in this step to avoid confusion from customers.


laurapacilio (approved):
Thank you! :-)

PR #2151, opened 2022-02-21
Replace duplicate entry in not needed config values

rescDev:
Changes:
- `registry_session_encryption_key` was present twice in the note
- replaced with `registry_session_secret_key`  which in my opinion was meant here ;)


bnferguson (approved):
Looks good here! Nice catch on that typo! 🙌

laurapacilio:
@rescDev I think we're almost ready to merge this in, except that our automated deploy is failing. I suspect that we pushed some changes to `master` in the time since you've  made this PR that may be causing the failure. Would you be able to sync your branch with `master`? I think that will fix the deploy errors. Thank you! :-)

rescDev:
@laurapacilio sync with `master` done :)

PR #2152, opened 2022-02-21
Clarify where to find VCS Providers

Jonher937:
The docs did not match up with what I was seeing in terraform cloud.
I was looking for "VCS Providers" that didn't exist.


laurapacilio (approved):
Thank you! We appreciate it :-)

PR #2160, opened 2022-02-25
Add clarity to automated install ip address flags

jaredfholgate:
The existing documentation did not make it clear that if the `private-address` and `public-address` are not populated either automatically or manually, then Terraform Enterprise will fail to install correctly. Trouble shooting this was time consuming and difficult, so improving the notes will hopefully aid our customers if they hit the same issue and are not using one of our curated modules.


content/enterprise/install/automated/automating-the-installer.mdx
```diff
@@ -295,7 +295,7 @@ bash ./install.sh \
...
+Note the `private-address` and `public-address` flags provided to the installer.  If these are left out the installer will attempt to determine them itself and will prompt for them if it is unable to. If using cloud-init or another non-interactive method to run the installer you will see this prompt in the log output, but the installer will continue to run. If the installer is unable to determine the IP addresses and you do not provide them via the flags, it will appear that the install has been successful, but functionality of Terraform Enterprise will be degraded. For example you may see a `Internal error: SIW-001` error when importing modules to the private registry or when you run the `tfe-admin health-check` command you may see that the Archivist and Vault containers are unhealthy. If the virtual machine instance will not have its own separate public IP address (i.e. you are using a load balancer and a private subnet), you should provide the private IP address for both flags.
```
laurapacilio, line 298: I wonder if there's a more clear way to write this. Do we recommend that people include these? If so I wonder if we should say that directly. Like "We recommend including the flags, x, y, z. If you leave them out...."  Thoughts?

laurapacilio (commented):
Question first for you before I copy edit :)

content/enterprise/install/automated/automating-the-installer.mdx
laurapacilio: oh! we have a recommendation t the end! I wonder if we could start out with that and then explain why.

jaredfholgate: Hi @laurapacilio. The recommendation was specifically about the values in the flags as opposed to always supplying them. I have tried to make it clearer by adding the recommendation at the start and then a new paragraph for the explanation.

content/enterprise/install/automated/automating-the-installer.mdx
```diff
@@ -295,7 +295,9 @@ bash ./install.sh \
...
+Note the `private-address` and `public-address` flags provided to the installer. We recommend you always supply these flags. If the virtual machine instance will not have its own separate public IP address (i.e. you are using a load balancer and a private subnet), you should provide the private IP address for both flags.
+
+If these flags are left out the installer will attempt to determine them itself and will prompt for them if it is unable to. If using cloud-init or another non-interactive method to run the installer you will see this prompt in the log output, but the installer will continue to run. If the installer is unable to determine the IP addresses and you do not provide them via the flags, it will appear that the install has been successful, but functionality of Terraform Enterprise will be degraded. For example you may see a `Internal error: SIW-001` error when importing modules to the private registry or when you run the `tfe-admin health-check` command you may see that the Archivist and Vault containers are unhealthy.
```
laurapacilio, line 300:
```suggestion
We recommend you always supply the `private-address` and `public-address` flags to the installer. If the virtual machine instance will not have its own separate public IP address (i.e. you are using a load balancer and a private subnet), you should provide the private IP address for both flags.

The installer will prompt you for these flags if it cannot determine them automatically. If you use `cloud-init` or another non-interactive method to run the installer, you will see this prompt in the log output, but the installer will continue to run. If the installer is unable to determine the IP addresses and you do not provide them, the installation will appear to be successful, but Terraform Enterprise will have degraded functionality. For example, you may see `Internal error: SIW-001` when importing modules to the private registry or notice that the Archivist and Vault containers are unhealthy when you run the `tfe-admin health-check` command. 
```

laurapacilio (approved):
I like it! A couple of wording suggestions for flow, active voice, and clarity - see what you think! Thank you :)

PR #2168, opened 2022-03-03
Updated Gigabit (Gb/gb) to Gigabyte (GB) acronyms

cstella84:
A user pointed out that our documentation shows lowercase "b" instead of capital "B". This is misleading, as Gb is short for Gigabit and GB is short for Gigabyte. The doc needs to accurately reflect "Gigabyte" (GB).


sudomateo (approved):
Thank you!

PR #2170, opened 2022-03-04
updated guidance for TFE disk requirements

hrvojep:
updated guidance for TFE disk requirements

### DIFF
--- a/content/enterprise/system-overview/capacity.mdx
+++ b/content/enterprise/system-overview/capacity.mdx
@@ -60,6 +60,8 @@ monopolizing the CPU resources of the host.
 
 The amount of disk storage available to a system plays a small role in the capacity of an instance.
 A root volume with 200 GB of storage can sustain a capacity well over 100 concurrent runs.
+
+
 It is recommended to use a disk with a minimum of 500 IOPs or 100+MB/s throughput, whichever is the smaller and these are provisioned values.
 
 ## Disk I/O
--- a/content/enterprise/system-overview/capacity.mdx
+++ b/content/enterprise/system-overview/capacity.mdx
@@ -61,8 +61,6 @@ monopolizing the CPU resources of the host.
 The amount of disk storage available to a system plays a small role in the capacity of an instance.
 A root volume with 200 GB of storage can sustain a capacity well over 100 concurrent runs.
 
-
-
 ## Disk I/O
 
 Because of the amount of churn caused by container creation as well as Terraform state management,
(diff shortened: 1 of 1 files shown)


content/enterprise/system-overview/capacity.mdx
```diff
@@ -72,3 +74,6 @@ This resource is harder to predict than memory or CPU usage because it varies pe
 but in general we suggest 35 IOPS per concurrent Terraform run. So if an instance is configured for
 20 concurrent runs, the disk should have 700 IOPS allocated. For reference, on AWS, an EBS volume
 with an allocated size of 250 GB comes with a steady state of 750 IOPS.
+
+It is recommended to use a disk with a minimum of 500 IOPS or 100+MB/s throughput, whichever is the smaller and these are provisioned values.
```
sudomateo, line 78: Where does this recommendation come from?

content/enterprise/system-overview/capacity.mdx
```diff
@@ -61,6 +61,8 @@ monopolizing the CPU resources of the host.
 The amount of disk storage available to a system plays a small role in the capacity of an instance.
 A root volume with 200 GB of storage can sustain a capacity well over 100 concurrent runs.
 
+
+
```
sudomateo, line 65: The extra space between the text and the heading isn't needed. It'll render the same as long as there's at least 1 empty line in between.

```suggestion
```

sudomateo (changes requested):

content/enterprise/system-overview/capacity.mdx
hrvojep: https://support.hashicorp.com/hc/en-us/articles/4409044739859-Container-ptfe-base-startup-failed

hrvojep: Solution 2

sudomateo: I'm an engineer on the Terraform Enterprise team and I haven't seen any official recommendation for IOPS and/or throughput aside from the paragraph right above this.

> This resource is harder to predict than memory or CPU usage because it varies per Terraform module, but in general we suggest 35 IOPS per concurrent Terraform run. So if an instance is configured for 20 concurrent runs, the disk should have 700 IOPS allocated. For reference, on AWS, an EBS volume with an allocated size of 250 GB comes with a steady state of 750 IOPS.

In our load testing, we've generally used a disk type that favors I/O, such as `io2` (AWS) with 700 IOPS. 

However, to stay consistent with Support and since this is a minimum recommendation, I would recommend this wording.

```suggestion
It is recommended to use a disk with at least 500 IOPS or 100+ MB/s throughput, whichever is lower.
```

PR #2171, opened 2022-03-07
Add missing include relations in docs

sebasslash:
## Description
This PR does two things:
* Add missing `include` relations for several resources. These relations are defined in a `Available Related Resources` section at the bottom of the page.
* Some relations were listed in bullet point format (see [Organizations](https://www.terraform.io/cloud-docs/api-docs/organizations#relationships)) and have now been moved to table format. 

## Notes for Reviewer
There is some slight discrepancy between `Available Related Resources` sections, where for some the description mentions the relations are available for all GET endpoints and others are more ambiguous, mentioning they may be available in several responses. Which convention should we follow?

### DIFF
--- a/content/cloud-docs/api-docs/agents.mdx
+++ b/content/cloud-docs/api-docs/agents.mdx
@@ -545,3 +545,11 @@ $ curl \
   --request DELETE \
   https://app.terraform.io/api/v2/agent-pools/apool-MCf6kkxu5FyHbqhd
 ```
+
+### Available Related Resources
+
+The GET endpoints above can optionally return related resources, if requested with [the `include` query parameter](/cloud-docs/api-docs/#inclusion-of-related-resources). The following resource types are available:
+
+| Resource Name  | Description                                 |
+| -------------- | ------------------------------------------- |
+| `workspaces`   | The workspaces attached to this agent pool. |
--- a/content/cloud-docs/api-docs/oauth-clients.mdx
+++ b/content/cloud-docs/api-docs/oauth-clients.mdx
@@ -409,3 +409,11 @@ curl \
   --request DELETE \
   https://app.terraform.io/api/v2/oauth-clients/oc-XKFwG6ggfA9n7t1K
 ```
+
+### Available Related Resources
+
+The GET endpoints above can optionally return related resources, if requested with [the `include` query parameter](/cloud-docs/api-docs/#inclusion-of-related-resources). The following resource types are available:
+
+| Resource Name  | Description                              |
+| -------------- | ---------------------------------------- |
+| `oauth_tokens` | The OAuth tokens managed by this client  |
--- a/content/cloud-docs/api-docs/organizations.mdx
+++ b/content/cloud-docs/api-docs/organizations.mdx
@@ -800,12 +800,22 @@ curl \
 }
 ```
 
+## Available Related Resources
+
+The GET endpoints above can optionally return related resources, if requested with [the `include` query parameter](/cloud-docs/api-docs/#inclusion-of-related-resources). The following resource types are available:
+
+| Resource Name         | Description                                                                                  |
+| --------------------- | -------------------------------------------------------------------------------------------- |
+| `entitlement_set`     | The entitlement set that determines which Terraform Cloud features the organization can use. |
+
 ## Relationships
 
-The following relationships may be present in various responses:
+The following relationships may be present in various responses.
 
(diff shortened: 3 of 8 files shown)


content/cloud-docs/api-docs/agents.mdx
```diff
@@ -545,3 +545,11 @@ $ curl \
...
   https://app.terraform.io/api/v2/agent-pools/apool-MCf6kkxu5FyHbqhd
 ```
+
+### Available Related Resources
+
+The GET endpoints above can optionally return related resources, if requested with [the `include` query parameter](/cloud-docs/api-docs/#inclusion-of-related-resources). The following resource types are available:
+
+| Resource Name  | Description                                 |
+| -------------- | ------------------------------------------- |
+| `workspaces`   | The workspaces attached to this agent pool. |
```
brandonc, line 555: I thinnk this was added in 1.x, which is unreleased?

brandonc, line 555: I thinnk this was added in 1.x, which is unreleased?

content/cloud-docs/api-docs/oauth-clients.mdx
```diff
@@ -409,3 +409,11 @@ curl \
...
   https://app.terraform.io/api/v2/oauth-clients/oc-XKFwG6ggfA9n7t1K
 ```
+
+### Available Related Resources
+
+The GET endpoints above can optionally return related resources, if requested with [the `include` query parameter](/cloud-docs/api-docs/#inclusion-of-related-resources). The following resource types are available:
+
+| Resource Name  | Description                              |
+| -------------- | ---------------------------------------- |
+| `oauth-tokens` | The OAuth tokens managed by this client  |
```
brandonc, line 419: I think the field is actually "oauth_tokens" with an underscore. I don't think we do any kind of dasherizing on these values

content/cloud-docs/api-docs/organizations.mdx
```diff
@@ -800,12 +800,14 @@ curl \
 }
 ```
 
-## Relationships
+## Available Related Resources
```
brandonc, line 803: I don't think this is true. The relationships exist, but [only the "entitlement_set" relationship can be included](https://github.com/hashicorp/atlas/blob/main/app%2Fcontrollers%2Fapi%2Fv2%2Forganizations_controller.rb#L155-L157).

brandonc (changes requested):
I found some issues, hopefully this clears up the question you asked in the PR description.

content/cloud-docs/api-docs/oauth-clients.mdx
nfagerlund: Yeah, Brandon is (unfortunately!) correct! Although we use `dasherized-case` (kebab case? I can never remember) within the API path names, we use `snake_case` in the `include` parameter. :coolcrying:

content/cloud-docs/api-docs/agents.mdx
brandonc: That doesn't matter, disregard

content/cloud-docs/api-docs/organizations.mdx
```diff
@@ -800,12 +800,22 @@ curl \
 }
 ```
 
+## Available Related Resources
+
+The GET endpoints above can optionally return related resources, if requested with [the `include` query parameter](/cloud-docs/api-docs/#inclusion-of-related-resources). The following resource types are available:
+
+| Resource Name         | Description                                                                                  |
+| --------------------- | -------------------------------------------------------------------------------------------- |
+| `entitlement-set`     | The entitlement set that determines which Terraform Cloud features the organization can use. |
```
brandonc, line 809:
```suggestion
| `entitlement_set`     | The entitlement set that determines which Terraform Cloud features the organization can use. |
```

content/cloud-docs/api-docs/policies.mdx
```diff
@@ -438,4 +438,6 @@ curl \
 
 The GET endpoints above can optionally return related resources, if requested with [the `include` query parameter](/cloud-docs/api-docs/#inclusion-of-related-resources). The following resource types are available:
 
-* `policy-sets` - Policy sets that any returned policies are members of.
+| Resource Name | Description                                            |
+| ------------- | ------------------------------------------------------ |
+| `policy-sets` | Policy sets that any returned policies are members of. |
```
brandonc, line 443:
```suggestion
| `policy_sets` | Policy sets that any returned policies are members of. |
```

brandonc (changes requested):
These darn snake_case include params

PR #2173, opened 2022-03-08
Added example of setting TFC_ADDRESS as environment variable

aibadger:
TFC_ADDRESS called out in this doc but also good to include a basic example similar to agent setup exports in TFE.


laurapacilio:
@sudomateo - This looks good to me, but wanted to get input from someone on TFE :-)

sudomateo (changes requested):
I like that we're providing an example now. However, I think we should reword the section a bit. It's a bit too verbose as it stands and honestly, a bit misleading since it's not a hostname being configured, but rather an address.

Here's what I think we should change it to:

```
* **Address Registration**: `tfc-agent` attempts to register with Terraform
  Cloud by default. To register `tfc-agent` with TFE, you must explicitly
  define the TFE address via the `TFC_ADDRESS` environment variable or the
  `-address` CLI flag.

    Here's an example of setting the TFE address using the `TFC_ADDRESS`
    environment variable.

    ```
    export TFC_ADDRESS='https://tfe.example.com'
    tfc-agent
    ```

    Here's an example setting the TFE address using the `-address` CLI flag.

    ```
    tfc-agent -address 'https://tfe.example.com'
    ```
```

PR #2182, opened 2022-03-11
fix: POST requires singular form

ansgarm:
Hi 👋 We noticed that the `type` seems to be `notification-configuration` instead of `notification-configurations`. The docs also mention the singular form right above this sample.
Else we'd get an HTTP 400 error when trying to post a new notification configuration.

It seems a bit inconsistent across other calls (we've encountered both `notification-configuration` and `notification-configurations` in the `type` field). But I haven't checked whether that matches the docs.


laurapacilio (approved):
Thank you so much!! Please merge when you're ready

ansgarm:
I don't have write permissions, but I'm sure somebody else will merge this :)

PR #2189, opened 2022-03-16
Added the Comments API docs

sebasslash:
Adding documentations for the comments API. Support for the comments API will be added to [go-tfe](https://github.com/hashicorp/go-tfe/pull/355) and thus go-tfe's docs need to be able to reference a API documentation page that exists.

### DIFF
--- a/content/cloud-docs/api-docs/comments.mdx
+++ b/content/cloud-docs/api-docs/comments.mdx
@@ -38,8 +38,7 @@ page_title: Comments - API Docs - Terraform Cloud and Terraform Enterprise
 
 # Comments API
 
-A Comment represents a Terraform Enterprise comment. 
-
+Comments allow users to leave feedback or record decisions about a run. 
 
 ## List Comments for a Run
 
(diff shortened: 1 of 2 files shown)


content/cloud-docs/api-docs/comments.mdx
```diff
@@ -0,0 +1,224 @@
...
+
+[504]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504
+
+[JSON API document]: /cloud-docs/api-docs#json-api-documents
+
+[JSON API error object]: https://jsonapi.org/format/#error-objects
+
+# Comments API
+
+A Comment represents a Terraform Enterprise comment.
```
annawinkler, line 41: Where would we find a Terraform Enterprise comment? I'm wondering if more description of what a comment is might be helpful here. What do you think?

annawinkler (approved):
🥇 Thank you for adding the docs for this API! I have one question/comment that isn't a blocker.

content/cloud-docs/api-docs/comments.mdx
laurapacilio: I 100% agree with this! Can we provide a more detailed descriptions of what comments are? Also - is this NOT relevant to terraform cloud? If so, I think we should explicitly say that here. If it IS relevant, then what does it look like in TFC? I just noted that we explicitly say "Terraform Enterprise" which makes me think it's not relevant to TFC

annawinkler (approved):
💯 Looks good to me! Thank you.

PR #2199, opened 2022-03-22
Private providers update for TFE release

laurenrother:
This PR removes all the notes/callouts that private providers do not work in TFE (because now they do 🎉 ).


laurapacilio (commented):
I found two more notes we should take out:

- https://www.terraform.io/cloud-docs/registry/using
- https://www.terraform.io/registry/private

I think once we do those two additional, we should be good!

laurenrother:
@laurapacilio now that GH is back up. I updated the two pages I missed :D

laurapacilio (approved):
yay!!

PR #2204, opened 2022-03-23
update private registry provider publishing docs

slavrd:
When a user is trying to publish a new provider to the TFC private registry they need to create the provider before creating a version for it (step 2).

This step is missing from the [Publishing Private Providers docs](https://www.terraform.io/cloud-docs/registry/publish-providers#publishing-a-provider-and-creating-a-version).

### DIFF
--- a/content/cloud-docs/registry/publish-providers.mdx
+++ b/content/cloud-docs/registry/publish-providers.mdx
@@ -41,7 +41,7 @@ To publish a new provider through the Terraform Cloud API:
     ```json
     "key-id": "32966F3FB5AC1129"
     ```
-1. If the provider does not yet exist in the private registry use the [Create a Provider endpoint](/https://www.terraform.io/cloud-docs/api-docs/private-registry/providers#create-a-provider) to create it. If you are publishing a new version for an already existing provider skip this step.
+1. Use the [Create a Provider endpoint](/cloud-docs/api-docs/private-registry/providers#create-a-provider) to create the provider in Terraform Cloud. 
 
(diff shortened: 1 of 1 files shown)


content/cloud-docs/registry/publish-providers.mdx
```diff
@@ -41,6 +41,8 @@ To publish a new provider through the Terraform Cloud API:
     ```json
     "key-id": "32966F3FB5AC1129"
     ```
+1. If the provider does not yet exist in the private registry use the [Create a Provider endpoint](/https://www.terraform.io/cloud-docs/api-docs/private-registry/providers#create-a-provider) to create it. If you are publishing a new version for an already existing provider skip this step.
```
laurapacilio, line 44: Thank you!  I'm wondering - could we put this as the first step in the process before the GPG key endpoint step? I think that would help everything flow better and would make logical sense :-) 

```suggestion
1. Use the [Create a Provider endpoint](/https://www.terraform.io/cloud-docs/api-docs/private-registry/providers#create-a-provider) to create the provider in Terraform Cloud. 
```

laurapacilio (changes requested):
thank you!! I appreciate you catching this and adding it.

content/cloud-docs/registry/publish-providers.mdx
slavrd: Telling users that they can skip the provider creation does indeed seem a bit redundant :) 

To me, personally having the step to create the GPG key first seems more logical as the GPG key is something that can and likely will be used across multiple providers in this namespace. Then we have a sequence of tightly related resources - Provider > Provider versions > Provider version platforms.

Still, I would ultimately defer to you on this so if you feel that it is better we can change it.

laurapacilio: okay! I trust your judgment - let's have this be the second step, just after the GPG key :-).

laurapacilio (approved):
Thank you! please merge when you're ready :-)

PR #2205, opened 2022-03-23
Single Sign-on: Microsoft Azure AD

munnep:
Procedure currently described doesn't work and has some naming mistakes

- The certificate in Azure AD Single sign-on won't be generated until after providing the Terraform Cloud information entered
- You manually need to download the certificate from Azure
- You then need to edit the SSO information in Terraform cloud to include this certificate

- Sometimes in the documentation Assertion Consumer Service URL is mentioned but this refers to the Reply URL

Someone need to check if the steps are correct and functioning more smoothly when you go over them

### DIFF
--- a/content/cloud-docs/users-teams-organizations/single-sign-on/azure-ad.mdx
+++ b/content/cloud-docs/users-teams-organizations/single-sign-on/azure-ad.mdx
@@ -63,10 +63,11 @@ For more information on the listed features, visit the [Microsoft Azure AD SAML 
 1. In the **Add Assignment** dialog, click the **Assign** button.
 
 ## Configuration (Terraform Cloud)
-1. Edit settings of your Azure SSO configuration
-   1. Go to **Public Certificate** 
-   1. Paste the contents of the SAML Signing Certificate downloaded from Microsoft Azure AD
-   1. Save Settings  
+To edit your Azure SSO configuration settings:
+
+   1. Go to **Public Certificate**. 
+   1. Paste the contents of the SAML Signing Certificate you downloaded from Microsoft Azure AD.
+   1. Save Settings.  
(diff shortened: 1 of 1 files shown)


content/cloud-docs/users-teams-organizations/single-sign-on/azure-ad.mdx
```diff
@@ -47,25 +44,29 @@ For more information on the listed features, visit the [Microsoft Azure AD SAML
...
 1. In the **Users and groups** dialog, select your user from the Users list, then click the **Select** button at the bottom of the screen.
 1. If you are expecting a role to be assigned to the users, you can select it from the **Select a role** dropdown. If no role has been set up for this app, you see "Default Access" role selected.
 1. In the **Add Assignment** dialog, click the **Assign** button.
 
 ## Configuration (Terraform Cloud)
-
+1. Edit settings of your Azure SSO configuration
+   1. Go to **Public Certificate** 
+   1. Paste the contents of the SAML Signing Certificate downloaded from Microsoft Azure AD
+   1. Save Settings
```
laurapacilio, line 69: Okay a couple quick notes here - I'm thinking this first one "go to" - is that a button? a dropdown? I also think - the last thing should be "Click **Save**". We want to phrase it in terms of an action and reference the exact UI thing.

```suggestion
To edit your Azure SSO configuration settings:

   1. Go to **Public Certificate**. 
   1. Paste the contents of the SAML Signing Certificate you downloaded from Microsoft Azure AD.
   1. Save Settings.  
```

laurapacilio (changes requested):
A couple of nits - thank you so much for fixing this!

laurapacilio (approved):

PR #2206, opened 2022-03-23
adds release notes for TFE v202203-1

theoskolnik:
These are the release notes and supplemental website changes for TFE v202203-1 release.

Hey @laurapacilio - mind taking a look over? Thanks!


content/enterprise/releases/2022/v202203-1.mdx
```diff
@@ -0,0 +1,39 @@
...
+1. Update the names of containers, which may break container monitoring or custom tooling that identifies containers by name. The April 2022 release notes will explain these name changes in more detail and provide a complete list of old and new container names.
+1. Change the default value of [`restrict_worker_metadata_access`](https://www.terraform.io/enterprise/install/automated/automating-the-installer#restrict_worker_metadata_access) to 1 (true) instead of 0 (false). If you rely on the instance metadata endpoint (and make use of its instance profile), you must explicitly set the `restrict_worker_metadata_access` configuration flag in `replicated.conf` to 0.
```
laurapacilio, line 13:
```suggestion
1. Change the default value of [`restrict_worker_metadata_access`](https://www.terraform.io/enterprise/install/automated/automating-the-installer#restrict_worker_metadata_access) to 1 (true) instead of 0 (false). If you rely on the instance metadata endpoint and make use of its instance profile, you must explicitly set the `restrict_worker_metadata_access` configuration flag in `replicated.conf` to 0.
```

content/enterprise/releases/2022/v202203-1.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+## APPLICATION LEVEL BREAKING CHANGES:
+
+1. The minimum required Replicated version is now 2.53.4. Airgap customers can download the latest version of Replicated from this [url](https://install.terraform.io/airgap/latest.tar.gz).
```
laurapacilio, line 17: I think if we're going to talk about the URL, we should list it out in the text. We don't recommend making "url" or "link" the hyperlink text, if that makes sense.

```suggestion
1. The minimum required Replicated version is now 2.53.4. Airgap customers can visit this URL to download the latest Replicated version: [https://install.terraform.io/airgap/latest.tar.gz](https://install.terraform.io/airgap/latest.tar.gz).
```

content/enterprise/releases/2022/v202203-1.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+## APPLICATION LEVEL FEATURES:
+
+1. Added the ability to [archive configuration versions](https://www.terraform.io/cloud-docs/workspaces/configurations#archiving-configuration-versions) either automatically or manually in order to free up storage space and limit the amount of sensitive data stored in Terraform Enterprise.
```
laurapacilio, line 21:
```suggestion
1. [Archive configuration versions](https://www.terraform.io/cloud-docs/workspaces/configurations#archiving-configuration-versions) either automatically or manually to free up storage space and limit the amount of sensitive data stored in Terraform Enterprise.
```

content/enterprise/releases/2022/v202203-1.mdx
```diff
@@ -0,0 +1,39 @@
...
+2. Improved the performance and reliability when destroying workspaces.
+3. Changed documentation link on preparing a module repository.
```
laurapacilio, line 23: do we really need this? This feels minor enough that it's not necessary. If it is necessary to mention, can we give a link?

content/enterprise/releases/2022/v202203-1.mdx
```diff
@@ -0,0 +1,39 @@
...
+4. Private providers are now supported in the private registry.
```
laurapacilio, line 24:
```suggestion
4. [Publish private providers](/cloud-docs/registry/publish-providers) to an organization's Terraform Cloud private registry. Once you have published a private provider through the API, members of your organization can search for it in the private registry UI and use it in configurations.
```

content/enterprise/releases/2022/v202203-1.mdx
```diff
@@ -0,0 +1,39 @@
...
+5. Added API endpoints for [downloading configuration versions](https://www.terraform.io/cloud-docs/api-docs/configuration-versions#download-configuration-files).
```
laurapacilio, line 25:
```suggestion
5. [Download configuration versions](https://www.terraform.io/cloud-docs/api-docs/configuration-versions#download-configuration-files) through the Terraform Cloud API. A configuration version is a resource used to reference the uploaded configuration files that are associated with the Terraform runs.
```

content/enterprise/releases/2022/v202203-1.mdx
```diff
@@ -0,0 +1,39 @@
...
+1. Added [Variable Sets](https://www.terraform.io/cloud-docs/api-docs/variable-sets).
```
laurapacilio, line 26:
```suggestion
1. [Variable Sets](https://www.terraform.io/cloud-docs/api-docs/variable-sets) let you reuse the same variables across multiple workspaces. For example, you could define a variable set of provider credentials and automatically apply it to all of the workspaces using that provider. 
```

content/enterprise/releases/2022/v202203-1.mdx
```diff
@@ -0,0 +1,39 @@
...
+1. Change the default value of [`restrict_worker_metadata_access`](https://www.terraform.io/enterprise/install/automated/automating-the-installer#restrict_worker_metadata_access) to 1 (true) instead of 0 (false). If you rely on the instance metadata endpoint (and make use of its instance profile), you must explicitly set the `restrict_worker_metadata_access` configuration flag in `replicated.conf` to 0.
+
+## APPLICATION LEVEL BREAKING CHANGES:
+
+1. The minimum required Replicated version is now 2.53.4. Airgap customers can download the latest version of Replicated from this [url](https://install.terraform.io/airgap/latest.tar.gz).
+
+## APPLICATION LEVEL FEATURES:
+
+1. Added the ability to [archive configuration versions](https://www.terraform.io/cloud-docs/workspaces/configurations#archiving-configuration-versions) either automatically or manually in order to free up storage space and limit the amount of sensitive data stored in Terraform Enterprise.
+2. Improved the performance and reliability when destroying workspaces.
```
laurapacilio, line 22: What does this mean/why is this awesome and included in our release notes? Was there a problem before? I just want a statement of value to the user

content/enterprise/releases/2022/v202203-1.mdx
```diff
@@ -0,0 +1,39 @@
...
+1. Added [Variable Sets](https://www.terraform.io/cloud-docs/api-docs/variable-sets).
+1. Added html and/or plain text versions to all account-related emails.
+
+## APPLICATION LEVEL BUG FIXES:
+
+1. Fixed response code for JSON parse errors.
+2. Fixed API dasherizing state output keys when the top level output type is an array containing maps.
+3. Fixed 500 error when fetching admin organization that does not exist.
+1. Fixed syntax error when setting a Terraform variable to `null`.
+1. Fixed Tag creation dates that are now displayed correctly.
```
laurapacilio, line 35: In general, I'd want to see a bit more context on these. For example - where was JSON parsing failing? What was wrong with tag creation dates?

content/enterprise/releases/2022/v202203-1.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+## APPLICATION LEVEL SECURITY FIXES:
+
+1. Ongoing container updates to address reported vulnerabilities in underlying packages / dependencies, including OpenSSL-related [CVE-2022-0778](https://nvd.nist.gov/vuln/detail/CVE-2022-0778).
```
laurapacilio, line 39: We don't need a numbered list if there is only one item! :-) 

```suggestion
Ongoing container updates to address reported vulnerabilities in underlying packages / dependencies, including OpenSSL-related [CVE-2022-0778](https://nvd.nist.gov/vuln/detail/CVE-2022-0778).
```

laurapacilio (approved):
I'm approving so I don't block this from going out, but I have a few suggestions and comments for you. In general, I have the following thoughts that inform my recommendations:

We shouldn't start all the updates with words like “Added,” “Fixed,” or “Changed” because readers have to do extra mental work to ignore them and get to the substance of the update. Instead, we should write a complete sentence that opens with what part of the system has been changed or the new feature that is available.

**Needs Work**: “Added Prometheus format for usage metrics.”
**Better:** "Usage metrics are available in Prometheus format."

Less experienced users may not understand the potential impact of a new feature. We should always explain why fixes were necessary, why additions are valuable, and what new terminology means. Where possible, we should also link to related documentation or Learn guides with more information.

Again, I know there's a deadline, but wanted to provide this feedback and explain the rationale for my edits :-)  Thank you for the opportunity to provide feedback!

sudomateo (changes requested):
Aside from the content changes that Laura already mentioned, you'll need to update `content/enterprise/releases/index.mdx` and add the release information to the table and update the date for the next release.

theoskolnik:
Hey @laurapacilio - thanks for leaving your changes, it's really helpful to get a writer's eyes on this. About the past tense commit messages ("added", "fixed" etc) -- it's a byproduct of internal but also general git conventions that are [documented here](https://hashicorp.atlassian.net/wiki/spaces/TFENG/pages/2300084414/Monthly+TFE+Release+Sign+Off). Generally it's recommended to start commit messages with a verb (some use imperative, some use present tense, in this case we're using past tense) so that a commit message can complete the sentence ("if applied this commit will..."). However, maybe it's time to revisit these conventions so that our release notes are easier to understand. So for now, I'm going to maintain the past tense, though I think we need to revisit this.

content/enterprise/releases/2022/v202203-1.mdx
theoskolnik: Going to leave the past tense for now.

content/enterprise/releases/2022/v202203-1.mdx
theoskolnik: Going to leave the past tense for now.

content/enterprise/releases/2022/v202203-1.mdx
theoskolnik: Going to leave the past tense for now.

content/enterprise/releases/2022/v202203-1.mdx
theoskolnik: Removed.

sudomateo (approved):

content/enterprise/releases/2022/v202203-1.mdx
laurapacilio: okay! but could we please add in the extra context? Regardless of how this starts out, all of my suggested changes contain additional context. Is it possible to add that additional context in? I think it will be really helpful for users.

laurapacilio:
okay! But I feel very strongly that we need to include additional context. My suggested edits have additional context and links in them. Would it be possible to add that in before merging, even if you don't change the beginning?

theoskolnik:
Hey @laurapacilio -- I went ahead and incorporated the descriptions from above. I think we all agree that the entire release notes process can be better, so here are a few things I'd like to do: 

- [include you on this PR instead](https://github.com/hashicorp/ptfe-releases/pull/153) - this is when all commits from all services are gathered in a single PR and [engineers are instructed](https://hashicorp.atlassian.net/wiki/spaces/TFENG/pages/2300084414/Monthly+TFE+Release+Sign+Off) to move their commit to the appropriate section.
- This is a good place to set expectations around messaging as well (and will require a bit of training in the beginning) -- I heard you have a guide? Maybe we use this to give engineers better guidance. 

Thanks for helping us out here :)

PR #2211, opened 2022-03-25
update 'protocols' attribute in the examples for creating private provider version

slavrd:
I'd like to propose updating the value for the  'protocols' attribute in the examples for creating a private provider version. The current value - `["4.0"]` is valid but it will not work if used in practice. I tested this and also found that it is stated in the provider registry protocol [docs](https://www.terraform.io/internals/provider-registry-protocol#protocols).

> Only Terraform 0.13 and later support third-party provider registries and that Terraform version requires API version 5.0 or later, so in practice it isn't useful to list major versions 4 or earlier in a third-party provider registry.

I hope this will help avoid situations like:

* A user has a custom provider that they use locally and decide to publish it
* They are unclear about the 'protocols' attribute and so copy/paste the example from the docs
* They finish the provider publishing process successfully

When they attempt to use the published provider the Terraform CLI will return an error saying that "no suitable version of the provider are available" and it is not quite apparent why that is.

### DIFF
--- a/content/cloud-docs/api-docs/private-registry/provider-versions-platforms.mdx
+++ b/content/cloud-docs/api-docs/private-registry/provider-versions-platforms.mdx
@@ -79,7 +79,7 @@ Properties without a default value are required.
 | `data.attributes.key-id`    | string |         | A valid gpg-key string.                                           |
 | `data.attributes.protocols` | array  |         | An array of Terraform provider API versions that this version supports. Must be one or all of the following values `["4.0","5.0","6.0"]`. |
 
-**Note:** Only Terraform 0.13 and later support third-party provider registries and that Terraform version requires provider API version 5.0 or later, so in practice it isn't useful to list major versions 4 or earlier in the `protocols` attribute.
+-> **Note:** Only Terraform 0.13 and later support third-party provider registries, and that Terraform version requires provider API version 5.0 or later. So you do not need to list major versions 4.0 or earlier in the `protocols` attribute.
 
 ### Sample Payload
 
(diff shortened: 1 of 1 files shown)


laurapacilio:
Thank you! this is very helpful!  Question for you - would it be worthwhile to include something in the actual text about this as well - perhaps in the beginning of the endpoint, just under the title? just to warn people that they should be intentional about setting the protocols to 5.0 and above? 

I just want to make sure that people don't miss this, as it seems like it could be a pretty frustrating "gotcha"

slavrd:
I'd say it does make sense to clarify this a bit. I added a note below the table with the attributes of the request payload and also updated the `protocols` attr description - basically copied the text from the provider registry protocol [docs](https://www.terraform.io/internals/provider-registry-protocol#protocols). Hope this works :)

content/cloud-docs/api-docs/private-registry/provider-versions-platforms.mdx
```diff
@@ -77,7 +77,9 @@ Properties without a default value are required.
 | `data.type`                 | string |         | Must be `"registry-provider-versions"`.                           |
 | `data.attributes.version`   | string |         | A valid semver version string.                                    |
 | `data.attributes.key-id`    | string |         | A valid gpg-key string.                                           |
-| `data.attributes.protocols` | array  |         | Must be one or all of the following values `["4.0","5.0","6.0"]`. |
+| `data.attributes.protocols` | array  |         | An array of Terraform provider API versions that this version supports. Must be one or all of the following values `["4.0","5.0","6.0"]`. |
+
+**Note:** Only Terraform 0.13 and later support third-party provider registries and that Terraform version requires provider API version 5.0 or later, so in practice it isn't useful to list major versions 4 or earlier in the `protocols` attribute.
```
laurapacilio, line 82:
```suggestion
-> **Note:** Only Terraform 0.13 and later support third-party provider registries, and that Terraform version requires provider API version 5.0 or later. So you do not need to list major versions 4.0 or earlier in the `protocols` attribute.
```

laurapacilio (approved):
The tiniest formatting corrections :-) If you add these in, you can go ahead and get this merged!

PR #2212, opened 2022-03-25
Add Current State Version Outputs API

sebasslash:
Pulling the description of this API given by @brandonc :

This endpoint behaves similarly to the workspace current-state-version API, but solves a tricky access control problem: Unless users have permission to fully read the entire state, they are unable to fetch outputs without having a the specific output ids already. Within Terraform, the command terraform output will not allow you to read outputs unless you had the full state read permission. This solves the issue by returning the latest state version outputs without having to dig into the state first to find the output ids.

Sensitive values are not revealed and it's expected the caller will follow links to individual outputs to see sensitive values unredacted.


content/cloud-docs/api-docs/state-version-outputs.mdx
```diff
@@ -153,3 +153,65 @@ curl \
   }
 }
 ```
+
+## Show Current State Version Outputs for a Workspace
+
+This endpoint allows users without permissions to read state versions to fetch outputs for a workspace.
```
annawinkler, line 159: Clarifying questions - does this mean that anyone with the URL can send a request to the API? No token no auth, etc, is needed?

content/cloud-docs/api-docs/state-version-outputs.mdx
```diff
@@ -153,3 +153,65 @@ curl \
...
+
+-> **Note:** Sensitive values are not revealed and will be returned as `null`.
```
annawinkler, line 161: 💯 I like that this is called out.

content/cloud-docs/api-docs/state-version-outputs.mdx
```diff
@@ -153,3 +153,65 @@ curl \
   }
 }
 ```
+
+## Show Current State Version Outputs for a Workspace
```
annawinkler, line 157: What are version outputs? 🤔

annawinkler: Another, slightly related question, why would someone use this API?

annawinkler (dismissed):
Thank you for adding this document! I have a few general questions about this API, including why one might use it. I don't have the API implementation context, so perhaps this is because I'm new ✨

content/cloud-docs/api-docs/state-version-outputs.mdx
sebasslash: Nope, you will need to be a member of the organization the workspace belongs to and supply your user token. However it's for users who are a member of a team (within said organization) that doesn't have permissions to read state versions.

content/cloud-docs/api-docs/state-version-outputs.mdx
sebasslash: Excellent question! This is how I understand it (anyone feel free to correct me on this): when you initiate a successful run in TFC/E, the resulting Terraform state file is versioned... what we call a state version. Like a regular Terraform statefile, it can have outputs, e.g:
```hcl
# somewhere in my configuration
output "my-ec2-private-ip-addrress" {
  value = aws_instance.server.private_ip
  sensitive = false
}
```
When we talk about state version outputs, we refer to these values. Here are the docs if my explanation isn't clear: https://www.terraform.io/language/values/outputs

Hope this helps! 😃

content/cloud-docs/api-docs/state-version-outputs.mdx
annawinkler: Ah OK - that makes sense. What do you think about updating the sentence to reflect what you've said in your comment?

Maybe something like:

```suggestion
This endpoint allows workspace organization users who do not have permissions to read state versions to fetch outputs for a workspace.
```

content/cloud-docs/api-docs/state-version-outputs.mdx
```diff
@@ -153,3 +153,65 @@ curl \
...
+
+This endpoint allows organization users, who do not have permissions to read state versions, to fetch the latest [output values](/language/values/outputs) for a workspace. ([More about permissions.](/cloud-docs/users-teams-organizations/permissions))
```
annawinkler, line 159: 🏆 This is great!

annawinkler (approved):
Thanks for updating the docs and adding helpful links! ✨

content/cloud-docs/api-docs/state-version-outputs.mdx
brandonc: Please add a note w/ link about how to fetch sensitive output values (you have to use the outputs/:id GET endpoint for a specific output, which is the the related -> self link

brandonc (commented):
Thanks for taking this on! It looks good. I had one small improvement in mind.

PR #2213, opened 2022-03-28
fix: use filterVersions from hashicorp/dev-portal#231

zchsh:
👀 [Preview](https://terraform-website-git-zsfix-version-filter-hashicorp.vercel.app/downloads)

This PR fixes an issue with the Terraform `/downloads` page. Pre-release versions were previously not filtered as expected, causing build errors. See hashicorp/dev-portal#231 for details.


BRKalow (approved):
🙇

PR #2214, opened 2022-03-29
correct case of X-Tfc-Task-Signature

assareh:
This can be case sensitive
Screen Shot 2022-03-29 at 11 11 00 AM


laurapacilio (approved):
Thank you! Please merge when you're ready 🚀

PR #2215, opened 2022-03-29
Tfe updates

amy-hashi:
Updated TFE documentation to provide better guidance around Docker versions on RHEL7 vs 8, as well as some adjustments to the performance capacity docs to better guide customers around IOPs.

### DIFF
--- a/content/enterprise/before-installing/index.mdx
+++ b/content/enterprise/before-installing/index.mdx
@@ -132,7 +132,9 @@ Terraform Enterprise requires **at least one** of the following Docker Engine co
 
 1. Docker Engine 17.06.2-ce, 18.09.x, 19.03.x, or 20.10.x with `runc` v1.0.0-rc93 or greater.
 1. Docker Engine 20.10.x with `libseccomp` 2.4.4 or greater.
-1. Docker Engine 1.13.1 (RHEL only), 17.06.2-ce, 18.09.x, 19.03.x, or 20.10.x using a modified `libseccomp` profile.
+1. 17.06.2-ce, 18.09.x, 19.03.x, or 20.10.x using a [modified `libseccomp` profile](https://www.terraform.io/enterprise/before-installing#option-2-docker-engine-with-a-compatible-libseccomp-version).
+
+If you are installing on RHEL7, you also have the option of Docker Engine 1.13.1 from the Extras Repo, with a [modified `libseccomp` profile](https://www.terraform.io/enterprise/before-installing#option-2-docker-engine-with-a-compatible-libseccomp-version).
 
 Refer to the sections below for instructions on how to verify your Docker Engine configuration.
 
--- a/content/enterprise/before-installing/index.mdx
+++ b/content/enterprise/before-installing/index.mdx
@@ -134,7 +134,7 @@ Terraform Enterprise requires **at least one** of the following Docker Engine co
 1. 20.10.x with `libseccomp` 2.4.4 or greater.
 1. 17.06.2-ce, 18.09.x, 19.03.x, or 20.10.x using a [modified `libseccomp` profile](https://www.terraform.io/enterprise/before-installing#option-2-docker-engine-with-a-compatible-libseccomp-version).
 
-If you are installing on RHEL7, you can use Docker Engine 1.13.1 from the Extras Repo, with a [modified `libseccomp` profile](https://www.terraform.io/enterprise/before-installing#option-2-docker-engine-with-a-compatible-libseccomp-version).
+If you are installing on RHEL7, you can use Docker Engine 1.13.1 from the Extra Packages for Enterprise Linux (EPEL) repository, with a [modified `libseccomp` profile](https://www.terraform.io/enterprise/before-installing#option-2-docker-engine-with-a-compatible-libseccomp-version).
 
 Refer to the sections below for instructions on how to verify your Docker Engine configuration.
 
--- a/content/enterprise/before-installing/index.mdx
+++ b/content/enterprise/before-installing/index.mdx
@@ -130,11 +130,11 @@ See [Network Requirements](/enterprise/before-installing/network-requirements) f
 
 Terraform Enterprise requires **at least one** of the following Docker Engine configurations, in order of preference:
 
-1. Docker Engine 17.06.2-ce, 18.09.x, 19.03.x, or 20.10.x with `runc` v1.0.0-rc93 or greater.
-1. Docker Engine 20.10.x with `libseccomp` 2.4.4 or greater.
+1. 17.06.2-ce, 18.09.x, 19.03.x, or 20.10.x with `runc` v1.0.0-rc93 or greater.
+1. 20.10.x with `libseccomp` 2.4.4 or greater.
 1. 17.06.2-ce, 18.09.x, 19.03.x, or 20.10.x using a [modified `libseccomp` profile](https://www.terraform.io/enterprise/before-installing#option-2-docker-engine-with-a-compatible-libseccomp-version).
 
-If you are installing on RHEL7, you also have the option of Docker Engine 1.13.1 from the Extras Repo, with a [modified `libseccomp` profile](https://www.terraform.io/enterprise/before-installing#option-2-docker-engine-with-a-compatible-libseccomp-version).
+If you are installing on RHEL7, you can use Docker Engine 1.13.1 from the Extras Repo, with a [modified `libseccomp` profile](https://www.terraform.io/enterprise/before-installing#option-2-docker-engine-with-a-compatible-libseccomp-version).
 
 Refer to the sections below for instructions on how to verify your Docker Engine configuration.
 
--- a/content/enterprise/before-installing/os-specific/rhel-requirements.mdx
+++ b/content/enterprise/before-installing/os-specific/rhel-requirements.mdx
@@ -31,7 +31,7 @@ Afterwards, restart the Docker service and verify the newly installed version us
 
 ### Can I use the Docker version in the RHEL Extras repository?
 
-Yes! Just be sure to [modify the default `libseccomp` profile](/enterprise/before-installing#option-3-docker-engine-using-a-modified-libseccomp-profile).
+If you are running Red Hat Enterprise Linux v7, yes. Red Hat Enterprise Linux v8 does not offer docker in the extras repo. If you are running v7 and choose to run docker from extras, just be sure to [modify the default `libseccomp` profile](/enterprise/before-installing#option-3-docker-engine-using-a-modified-libseccomp-profile).
 
 ### When I run the installer, it allows me to download and install Docker CE on RedHat. Can I use that?
 
--- a/content/enterprise/before-installing/os-specific/rhel-requirements.mdx
+++ b/content/enterprise/before-installing/os-specific/rhel-requirements.mdx
@@ -31,7 +31,9 @@ Afterwards, restart the Docker service and verify the newly installed version us
 
 ### Can I use the Docker version in the RHEL Extras repository?
 
-If you are running Red Hat Enterprise Linux v7, yes. Red Hat Enterprise Linux v8 does not offer docker in the extras repo. If you are running v7 and choose to run docker from extras, just be sure to [modify the default `libseccomp` profile](/enterprise/before-installing#option-3-docker-engine-using-a-modified-libseccomp-profile).
+Yes, if you are running Red Hat Enterprise Linux v7. If you are running v7 and choose to run Docker from the Extra Packages for Enterprise Linux (EPEL) repository, you must [modify the default `libseccomp` profile](/enterprise/before-installing#option-3-docker-engine-using-a-modified-libseccomp-profile).
+
+Red Hat Enterprise Linux v8 does not offer docker in the extras repo.
 
 ### When I run the installer, it allows me to download and install Docker CE on RedHat. Can I use that?
 
--- a/content/enterprise/before-installing/os-specific/rhel-requirements.mdx
+++ b/content/enterprise/before-installing/os-specific/rhel-requirements.mdx
@@ -29,7 +29,7 @@ Afterwards, restart the Docker service and verify the newly installed version us
 
 ## FAQ
 
-### Can I use the Docker version in the RHEL Extras repository?
+### Can I use the Docker version in the RHEL Extras Repository?
 
 Yes, if you are running Red Hat Enterprise Linux v7. If you are running v7 and choose to run Docker from the Extra Packages for Enterprise Linux (EPEL) repository, you must [modify the default `libseccomp` profile](/enterprise/before-installing#option-3-docker-engine-using-a-modified-libseccomp-profile).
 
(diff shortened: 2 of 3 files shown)


content/enterprise/before-installing/index.mdx
```diff
@@ -132,7 +132,9 @@ Terraform Enterprise requires **at least one** of the following Docker Engine co
 
 1. Docker Engine 17.06.2-ce, 18.09.x, 19.03.x, or 20.10.x with `runc` v1.0.0-rc93 or greater.
 1. Docker Engine 20.10.x with `libseccomp` 2.4.4 or greater.
-1. Docker Engine 1.13.1 (RHEL only), 17.06.2-ce, 18.09.x, 19.03.x, or 20.10.x using a modified `libseccomp` profile.
+1. 17.06.2-ce, 18.09.x, 19.03.x, or 20.10.x using a [modified `libseccomp` profile](https://www.terraform.io/enterprise/before-installing#option-2-docker-engine-with-a-compatible-libseccomp-version).
```
laurapacilio, line 135: I see you removed "Docker Engine" which I agree with since we specify that they are all docker engines above. Can we remove that from the first two bullets, too? also! Is this a repeat of the first option?

content/enterprise/before-installing/index.mdx
```diff
@@ -132,7 +132,9 @@ Terraform Enterprise requires **at least one** of the following Docker Engine co
...
+
+If you are installing on RHEL7, you also have the option of Docker Engine 1.13.1 from the Extras Repo, with a [modified `libseccomp` profile](https://www.terraform.io/enterprise/before-installing#option-2-docker-engine-with-a-compatible-libseccomp-version).
```
laurapacilio, line 137: I'm also wondering where this falls in order of preference. The ones above are listed in order of most to least recommended. Should folks on RHEL7 use this one, or would we recommend one of the above?

```suggestion
If you are installing on RHEL7, you can also use Docker Engine 1.13.1 from the Extras Repo, with a [modified `libseccomp` profile](https://www.terraform.io/enterprise/before-installing#option-2-docker-engine-with-a-compatible-libseccomp-version).
```

content/enterprise/before-installing/os-specific/rhel-requirements.mdx
```diff
@@ -31,7 +31,7 @@ Afterwards, restart the Docker service and verify the newly installed version us
 
 ### Can I use the Docker version in the RHEL Extras repository?
 
-Yes! Just be sure to [modify the default `libseccomp` profile](/enterprise/before-installing#option-3-docker-engine-using-a-modified-libseccomp-profile).
+If you are running Red Hat Enterprise Linux v7, yes. Red Hat Enterprise Linux v8 does not offer docker in the extras repo. If you are running v7 and choose to run docker from extras, just be sure to [modify the default `libseccomp` profile](/enterprise/before-installing#option-3-docker-engine-using-a-modified-libseccomp-profile).
```
laurapacilio, line 34: I'm struggling with the repository name here. Is it actually called RHEL extras? I wonder if we should use the official name which I think is this: Extra Packages for Enterprise Linux (EPEL). But lmk if I'm very wrong :-)  

```suggestion
Yes, if you are running Red Hat Enterprise Linux v7. If you choose to run Docker from `extras`, you must [modify the default `libseccomp` profile](/enterprise/before-installing#option-3-docker-engine-using-a-modified-libseccomp-profile).

Red Hat Enterprise Linux v8 does not offer Docker in the `extras` repository. 
```

content/enterprise/system-overview/capacity.mdx
```diff
@@ -25,12 +25,12 @@ After factoring in the memory needed to run the base services that make up the a
 The settings for per-run memory and concurrency are available in the dashboard on port 8800, on the Settings page, under the Capacity section. They can also be set via
 the [application settings JSON file when using the automated install procedure](/enterprise/install/automated/automating-the-installer#available-settings).
 
-## Increasing Capacity
+## Increasing Concurrency
 
-To increase the number of concurrent runs, adjust the **Capacity** setting. Note that this setting is not limited by
+To increase the number of concurrent runs, adjust the `capacity_concurrency` setting. Note that this setting is not limited by
```
laurapacilio, line 30: This is a nitpick, but I think we shouldn't start with "Note that" unless we're going to put something in a Note block. Otherwise, we can just make it a statement. 

```suggestion
To increase the number of concurrent runs, adjust the `capacity_concurrency` setting. This setting is not limited by
```

content/enterprise/system-overview/capacity.mdx
```diff
@@ -25,12 +25,12 @@ After factoring in the memory needed to run the base services that make up the a
...
 system checks; it depends on the operator to provide enough memory to the system to accommodate the requested
-concurrent capacity. For instance, if **Capacity** is set to `100`, the instance would require, at a minimum,
-52 GB of memory reserved for Terraform runs.
+concurrent capacity. For instance, if `capacity_concurrency` is set to `100`, the instance would require, at a minimum,
+52 GB of memory just for Terraform runs, if the worker memory was set to 512. The rest of the TFE application requires 4 GB of memory at a minimum, on top of the Operating System requirements.
```
laurapacilio, line 33:
```suggestion
concurrent capacity. For example, if `capacity_concurrency` is set to `100` and the worker memory is set to 512, the instance would require a minimum of 52 GB of memory just for Terraform runs. The rest of the TFE application requires a minimum of 4 GB of memory in addition to the Operating System requirements.
```

content/enterprise/system-overview/capacity.mdx
```diff
@@ -39,6 +39,10 @@ system checks; it depends on the operator to provide enough memory to the system
 If the memory limit is adjusted to 1024 MB with the default capacity of 10, the instance would require, at a minimum,
 10 GB of memory reserved for Terraform runs.
 
+## Limiting CPU cores
+
+The default setting of `0` allows a Terraform run to utilize all available CPU cores on the TFE instance. You can adjust this setting by inputting a specific limit to the number of CPU cores the Terraform runs will have access to.
```
laurapacilio, line 44: Thinking of this, could we tell folks where to adjust this? What's the name of the attribute or whatever that they need to set. I'm thinking we say "You can also set a value for `attribute_name` to specify a limit for the number of CPU cores Terraform runs can access." Or something similar. Thoughts?

```suggestion
The default setting of `0` allows a Terraform run to use all available CPU cores on the TFE instance. You can also input a limit for the number of CPU cores Terraform runs can access. 
```

content/enterprise/system-overview/capacity.mdx
```diff
@@ -66,12 +70,12 @@ A root volume with 200 GB of storage can sustain a capacity well over 100 concur
...
 highly concurrent setups will begin pushing hard on disk I/O. In cloud environments like AWS that limit disk
 I/O to IOPS that are credited per disk, it's important to provision a minimum number to prevent I/O related
-stalls.
+stalls. Low disk I/O can create significant performance issues. 
 
 This resource is harder to predict than memory or CPU usage because it varies per Terraform module,
-but in general we suggest 35 IOPS per concurrent Terraform run. So if an instance is configured for
-20 concurrent runs, the disk should have 700 IOPS allocated. For reference, on AWS, an EBS volume
+but in general we suggest a minimum of 50 IOPS per concurrent Terraform run. So if an instance is configured for
+10 concurrent runs, the disk should have 500 IOPS allocated. For reference, on AWS, an EBS volume
```
laurapacilio, line 77:
```suggestion
but in general we recommend a minimum of 50 IOPS per concurrent Terraform run. So if an instance is configured for 10 concurrent runs, the disk should have 500 IOPS allocated. For reference, on AWS, an EBS volume
```

content/enterprise/system-overview/capacity.mdx
```diff
@@ -66,12 +70,12 @@ A root volume with 200 GB of storage can sustain a capacity well over 100 concur
...
 with an allocated size of 250 GB comes with a steady state of 750 IOPS.
 
-It is recommended to use a disk with a minimum of 500 IOPS or 100+MB/s throughput, whichever is the smaller and these are provisioned values.
+It is recommended to use a disk with a minimum of 500 IOPS, but high load systems should consider increasing this significantly. A production instance, with a consistantly high level of utilization, and a concurrency of 10, should have a disk with around 3,000 IOPS, ideally.
```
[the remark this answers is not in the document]
laurapacilio: I like that. Also I think we say "we recommend" rather than "it is recommended"  One nit is just I think I'd say "Scaling the disk beyond 8,000 IOPs does not significantly improve performance."

I think add it and I'll do a copyedit pass :-)

laurapacilio (changes requested):
Thank you :-) Some thoughts/suggestions for you! Let me know if you want to talk about any of them further!

content/enterprise/before-installing/index.mdx
amy-hashi: I'll make that update. And no, it's different from 1. 1 is any of those version combined with the specified version of runc. The other one is if you don't have that updated runc installed, you instead need the modified libseccomp profile. My brain hasn't come up with a better way to display this data, but I am open to suggestions.

content/enterprise/before-installing/index.mdx
amy-hashi: Yeah, that's the least preferred option as it's old and not kept as up to date as everything else.

content/enterprise/before-installing/index.mdx
laurapacilio: okay I'll leave as is for now :)

content/enterprise/system-overview/capacity.mdx
```diff
@@ -66,12 +65,11 @@ A root volume with 200 GB of storage can sustain a capacity well over 100 concur
...
 
 This resource is harder to predict than memory or CPU usage because it varies per Terraform module,
-but in general we suggest 35 IOPS per concurrent Terraform run. So if an instance is configured for
-20 concurrent runs, the disk should have 700 IOPS allocated. For reference, on AWS, an EBS volume
+but in general we recommend a minimum of 50 IOPS per concurrent Terraform run. So if an instance is configured for 10 concurrent runs, the disk should have 500 IOPS allocated. For reference, on AWS, an EBS volume
 with an allocated size of 250 GB comes with a steady state of 750 IOPS.
 
-It is recommended to use a disk with a minimum of 500 IOPS or 100+MB/s throughput, whichever is the smaller and these are provisioned values.
+It is recommended to use a disk with a minimum of 500 IOPS, but high load systems should consider increasing this significantly. A production instance, with a consistantly high level of utilization, and a concurrency of 10, should have a disk with around 3,000 IOPS, ideally. Internal testing has shown performance increases with additional IOPs up to 8,000. Scaling the disk beyond 8,000 IOPs does not significantly improve performance.
```
laurapacilio, line 74:
```suggestion
We recommend using a disk with a minimum of 500 IOPS, but high load systems should consider increasing this significantly. For example, a production instance with a consistently high level of utilization and a concurrency of 10 should ideally have a disk with about 3,000 IOPS. Internal testing has shown performance increases with additional IOPs up to 8,000. Scaling the disk beyond 8,000 IOPs does not significantly improve performance.
```

content/enterprise/system-overview/capacity.mdx
```diff
@@ -66,12 +65,11 @@ A root volume with 200 GB of storage can sustain a capacity well over 100 concur
 Because of the amount of churn caused by container creation as well as Terraform state management,
 highly concurrent setups will begin pushing hard on disk I/O. In cloud environments like AWS that limit disk
 I/O to IOPS that are credited per disk, it's important to provision a minimum number to prevent I/O related
-stalls.
+stalls. Low disk I/O can create significant performance issues. 
 
 This resource is harder to predict than memory or CPU usage because it varies per Terraform module,
-but in general we suggest 35 IOPS per concurrent Terraform run. So if an instance is configured for
-20 concurrent runs, the disk should have 700 IOPS allocated. For reference, on AWS, an EBS volume
+but in general we recommend a minimum of 50 IOPS per concurrent Terraform run. So if an instance is configured for 10 concurrent runs, the disk should have 500 IOPS allocated. For reference, on AWS, an EBS volume
```
laurapacilio, line 71:
```suggestion
but we generally recommend a minimum of 50 IOPS per concurrent Terraform run. So if an instance is configured for 10 concurrent runs, the disk should have 500 IOPS allocated. For reference, on AWS, an EBS volume
```

content/enterprise/before-installing/os-specific/rhel-requirements.mdx
```diff
@@ -31,7 +31,9 @@ Afterwards, restart the Docker service and verify the newly installed version us
 
 ### Can I use the Docker version in the RHEL Extras repository?
```
laurapacilio, line 32:
```suggestion
### Can I use the Docker version in the RHEL Extras Repository?
```

content/enterprise/before-installing/os-specific/rhel-requirements.mdx
```diff
@@ -31,7 +31,9 @@ Afterwards, restart the Docker service and verify the newly installed version us
...
 
-Yes! Just be sure to [modify the default `libseccomp` profile](/enterprise/before-installing#option-3-docker-engine-using-a-modified-libseccomp-profile).
+Yes, if you are running Red Hat Enterprise Linux v7. If you are running v7 and choose to run Docker from the Extra Packages for Enterprise Linux (EPEL) repository, you must [modify the default `libseccomp` profile](/enterprise/before-installing#option-3-docker-engine-using-a-modified-libseccomp-profile).
+
+Red Hat Enterprise Linux v8 does not offer docker in the extras repo.
```
laurapacilio, line 36:
```suggestion
Red Hat Enterprise Linux v8 does not offer docker in the EPEL repository.
```

content/enterprise/before-installing/index.mdx
```diff
@@ -130,9 +130,11 @@ See [Network Requirements](/enterprise/before-installing/network-requirements) f
...
 Terraform Enterprise requires **at least one** of the following Docker Engine configurations, in order of preference:
 
-1. Docker Engine 17.06.2-ce, 18.09.x, 19.03.x, or 20.10.x with `runc` v1.0.0-rc93 or greater.
-1. Docker Engine 20.10.x with `libseccomp` 2.4.4 or greater.
-1. Docker Engine 1.13.1 (RHEL only), 17.06.2-ce, 18.09.x, 19.03.x, or 20.10.x using a modified `libseccomp` profile.
+1. 17.06.2-ce, 18.09.x, 19.03.x, or 20.10.x with `runc` v1.0.0-rc93 or greater.
+1. 20.10.x with `libseccomp` 2.4.4 or greater.
+1. 17.06.2-ce, 18.09.x, 19.03.x, or 20.10.x using a [modified `libseccomp` profile](https://www.terraform.io/enterprise/before-installing#option-2-docker-engine-with-a-compatible-libseccomp-version).
+
+If you are installing on RHEL7, you can use Docker Engine 1.13.1 from the Extras Repo, with a [modified `libseccomp` profile](https://www.terraform.io/enterprise/before-installing#option-2-docker-engine-with-a-compatible-libseccomp-version).
```
laurapacilio, line 137:
```suggestion
If you are installing on RHEL7, you can use Docker Engine 1.13.1 from the Extra Packages for Enterprise Linux (EPEL) repository, with a [modified `libseccomp` profile](https://www.terraform.io/enterprise/before-installing#option-2-docker-engine-with-a-compatible-libseccomp-version).
```

laurapacilio (approved):
A couple of additional nits for consistency. If you feel good about adding those in, I feel comfortable with you merging. Thank you! :-)

PR #2218, opened 2022-03-30
Update Gitlab VCS integration docs

slavrd:
In version `14.3` Gitlab added an option for their OAuth tokens to auto-expire 2 hours after they are issued. They also made that enabled by default for any new OAuth applications that are created in Gitlab.
Currently, TFC will refresh Gitlab OAuth tokens if at least 6 hours have passed since they were issued. 

So, If a user does not disable the auto-expire option in their Gitlab OAuth application this creates problems with the TFC VCS integration - there is a gap of 4 hours where TFC will attempt to use a token that has expired.

This should eventually be resolved in TFC but until it is I thought that we can update the docs. 
At this point, the user disabling the `Expire access tokens` option is basically a requirement and the fact that it is not mentioned has been creating problems for users.


slavrd:
Hi. I was wondering about a note myself but hesitated because the answer to that question is not really great :) Still, if you feel that we should put a note I can make one.

laurapacilio:
hmmm - I think we should add one if we think users will have questions about it. But if we think we're overcompensating, then we can leave as is. Thoughts? Sorry to be so nitpicky about this - I just want to make sure users will be clear about what's happening and why!

slavrd:
I think I ultimately agree with you. If users have questions about it (and they likely will) they are going to open a ticket with support at which point we would give them this exact information. So it seems better to have that note in the docs.

content/cloud-docs/vcs/gitlab-eece.mdx
```diff
@@ -83,8 +83,11 @@ Leave the page open in a browser tab. In the next step you will copy values from
    | Name                    | Terraform Cloud (`<YOUR ORGANIZATION NAME>`)   |
    | Redirect URI            | `https://app.terraform.io/<YOUR CALLBACK URL>` |
    | Confidential (checkbox) | ✔️ (enabled)                                   |
+   | Expire access tokens (checkbox) | ✘ (disabled)                           |
    | Scopes (all checkboxes) | api                                            |
 
+   -> **Note:** If the `Expire access tokens` checkbox is not disabled Giltab will mark OAuth tokens for the application as expired 2 hours after they are issued. Currently, Terraform Cloud will refresh tokens if they are older than 6 hours. Because of this a gap of 4 hours would exist where Teraform Cloud will attempt to use a token which is marked as expired on Gitlab's side.
```
laurapacilio, line 89:
```suggestion
   -> **Note:** Terraform Cloud refreshes tokens if they are older than 6 hours, but GitLab marks OAuth tokens as expired 2 hours after they are issued. Disabling the `Expire access tokens` checkbox prevents Terraform Cloud from attempting to use expired tokens. 
```

content/cloud-docs/vcs/gitlab-com.mdx
```diff
@@ -70,8 +70,11 @@ Leave the page open in a browser tab. In the next step you will copy values from
    | Name                    | Terraform Cloud (`<YOUR ORGANIZATION NAME>`)                                                     |
    | Redirect URI            | `https://app.terraform.io/<YOUR CALLBACK URL>`, the redirect URI you copied from Terraform Cloud |
    | Confidential (checkbox) | ✔️ (enabled)                                                                                     |
+   | Expire access tokens (checkbox) | ✘ (disabled)                                                                             |
    | Scopes (all checkboxes) | api                                                                                              |
 
+   -> **Note:** If the `Expire access tokens` checkbox is not disabled Giltab will mark OAuth tokens for the application as expired 2 hours after they are issued. Currently, Terraform Cloud will refresh tokens if they are older than 6 hours. Because of this a gap of 4 hours would exist where Teraform Cloud will attempt to use a token which is marked as expired on Gitlab's side.
```
laurapacilio, line 76:
```suggestion
   -> **Note:**    -> **Note:** Terraform Cloud refreshes tokens if they are older than 6 hours, but GitLab marks OAuth tokens as expired 2 hours after they are issued. Disabling the `Expire access tokens` checkbox prevents Terraform Cloud from attempting to use expired tokens. 
```

laurapacilio (approved):
I'm going to approve so I don't block, but I had a wording suggestion to make the note clearer. See what you think! If you feel comfortable merging those changes, you can go ahead and merge this PR :-) Thank you!

slavrd:
Definitely sounds better. Thank you!

PR #2225, opened 2022-03-31
Update permissions required to create configuration versions

tylerwolf:
The permissions required to create a configuration version have been relaxed slightly.  Whereas previously configuration version creation required the apply runs permission, they can now be created by users with the queue plans permission.  Resulting runs will still follow the previous model (runs created by users with the queue plans permission will still need to be confirmed by a user with the apply runs permission).

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201972896866661

### DIFF
--- a/content/cloud-docs/api-docs/configuration-versions.mdx
+++ b/content/cloud-docs/api-docs/configuration-versions.mdx
@@ -42,7 +42,7 @@ page_title: Configuration Versions - API Docs - Terraform Cloud and Terraform En
 
 A configuration version (`configuration-version`) is a resource used to reference the uploaded configuration files. It is associated with the run to use the uploaded configuration files for performing the plan and apply.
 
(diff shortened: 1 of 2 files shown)


radditude (approved):
Makes sense to me content-wise! 

I seem to recall that we were making an attempt to move away from the "queue plans" terminology and toward "start plans" in the application and our documentation - not sure if that's still a thing, but Laura will know 🎉

laurapacilio (approved):
It does say "start plan" in the UI, which makes me feel like we're moving in that direction, but when I got to the permissions page, I see the actual permission is called "queue plans".

I support keeping this to match what's on the permissions page. Because if users go there for more details, they'll be looking for whatever we call it here --> so we don't want to use a term that's not on that page. 

When you actually set this permission in the UI, it's also referred to as "queueing" Screen Shot 2022-04-01 at 11 15 34 AM. 

Just my two cents :) I am finding out more about whether this language change is rolling out/when so we can pay attention to in in future docs updates, but for now I think this is the right approach!

PR #2233, opened 2022-04-08
Some small variable sets documentation fixes.

jondavidjohn:
- Fix navigation ordering based on alphabetical sorting.
- Remove reference to `PUT` for consistency with other "update" endpoint documentation.
- Add request body parameters to UPDATE documentation

### DIFF
--- a/content/cloud-docs/api-docs/variable-sets.mdx
+++ b/content/cloud-docs/api-docs/variable-sets.mdx
@@ -134,6 +134,18 @@ curl \
 | ------------ | ------------------- |
 | `:varset_id` | The variable set ID |
 
+### Request Body
+
+Properties without a default value are required.
+
+| Key path                        | Type    | Default | Description                                                                                                                             |
+| ------------------------------- | ------- | ------- | --------------------------------------------------------------------------------------------------------------------------------------- |
+| `data.name`                     | string  |         | The name of the variable set.                                                                                                           |
+| `data.description`              | string  | `""`    | Text displayed in the UI to contextualize the variable set and its purpose.                                                             |
+| `data.global`                   | boolean | `false` | When true, Terraform Cloud automatically applies the variable set to all current and future workspaces in the organization.             |
+| `data.relationships.workspaces` | array   | \[]     | Array of references to workspaces that the variable set should be assigned to. Sending an empty array clears all workspace assignments. |
+| `data.relationships.vars`       | array   | \[]     | Array of complete variable definitions that comprise the variable set.                                                                  |
+
 Terraform Cloud does not allow global variable sets to contain conflicting variables with the same name and type. You will receive a 422 response if you try to create a global variable set that contains conflicting variables.
 
 | Status  | Response                  | Reason(s)                                                                      |
--- a/content/cloud-docs/api-docs/variable-sets.mdx
+++ b/content/cloud-docs/api-docs/variable-sets.mdx
@@ -128,7 +128,7 @@ curl \
 
 ## Update a Variable Set
 
-`PUT/PATCH varsets/:varset_id`
+`PATCH varsets/:varset_id`
 
 | Parameter    | Description         |
 | ------------ | ------------------- |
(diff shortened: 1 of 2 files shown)


content/cloud-docs/api-docs/variable-sets.mdx
```diff
@@ -128,12 +128,24 @@ curl \
...
+| ------------------------------- | ------- | ------- | --------------------------------------------------------------------------------------------------------------------------------------- |
+| `data.name`                     | string  |         | The name of the variable set.                                                                                                           |
+| `data.description`              | string  | `""`    | Text displayed in the UI to contextualize the variable set and its purpose.                                                             |
+| `data.global`                   | boolean | `false` | When true, Terraform Cloud automatically applies the variable set to all current and future workspaces in the organization.             |
+| `data.relationships.workspaces` | array   | \[]     | Array of references to workspaces that the variable set should be assigned to. Sending an empty array clears all workspace assignments. |
+| `data.relationships.vars`       | array   | \[]     | Array of complete variable definitions that comprise the variable set.                                                                  |
```
nfagerlund, line 147: - The paths are wrong for several of these properties -- the ones that aren't relationships should be `data.attributes.thing`, not `data.thing`, as per the structure of the example payload. This error affects some other tables on this page as well.
- Since this is a patch request, the `default` value for all of these properties should be `(previous value)` -- compare to the tables in the workspaces API page, for example. 
- "Array of complete variable definitions" is a very strange way to word this, and is likely confusing to readers. (I know that wording came from the table for the create endpoint; it's probably a good idea to change both.) What I can gather from going back up to the "create" endpoint is, this relationships array doesn't work like most of the relationships in the API and instead requires full filled-out `vars` objects, but we're not pointing to a useful definition of those objects. I think there's two possible approaches for making this nice: 
    - Add a section to this page that describes the vars objects, and refer to it from the various endpoints that interact with it. (This almost already exists, in the content for the add/update a variable endpoints further down the page.)
    - Add the sub-properties of the vars objects to these tables, using the `data.relationships.vars[].key` etc. notation that we've used in some other places to list the properties of objects in an array. (See for example [the payload for workspace remote state consumers](https://www.terraform.io/cloud-docs/api-docs/workspaces#request-body-7).)

content/cloud-docs/api-docs/variable-sets.mdx
```diff
@@ -128,12 +128,24 @@ curl \
...
 
+### Request Body
+
+Properties without a default value are required.
+
+| Key path                        | Type    | Default | Description                                                                                                                             |
+| ------------------------------- | ------- | ------- | --------------------------------------------------------------------------------------------------------------------------------------- |
+| `data.name`                     | string  |         | The name of the variable set.                                                                                                           |
+| `data.description`              | string  | `""`    | Text displayed in the UI to contextualize the variable set and its purpose.                                                             |
+| `data.global`                   | boolean | `false` | When true, Terraform Cloud automatically applies the variable set to all current and future workspaces in the organization.             |
```
nfagerlund, line 145: This could use an explanation of how `attributes.global` and `relationships.workspaces` interact. I knew this at one point when we were reviewing a go-tfe PR, but I can't remember and wouldn't be able to derive it from the current docs without doing an experiment.

nfagerlund (commented):
Thanks for taking this on! 🌻 

I'm leaving some notes since I'm the one who filed the ticket, but I don't want to own the approval on this PR, so I'm marking this review as a "comment" instead of a "request changes;" Laura or someone from your team can be in charge of final approval once y'all feel the usability issues are dealt with.

laurapacilio:
Hi @jondavidjohn - wanted to check in on this, since it's been open for a bit. Is this still something you're hoping to merge?

jondavidjohn:
Thanks for the ping @laurapacilio, got distracted by a few other things and totally forgot about this.

I hope to circle back on it later this week.  Thanks!

laurapacilio:
Hi! I wanted to let you know that we're migrating all of the content in the `cloud-docs` directory to a new repository. To support the migration, there will be a docs code freeze on **Wednesday May 25th** where we'll ask folks not to merge any PRs for the TFC docs. Please either merge or close this PR before the freeze. Otherwise, we'll need to re-open this PR and recreate these changes in the new repository. 

Let me know if there's any way I can help out here. As always, thank you for your contributions to our docs! 😄

jondavidjohn:
@laurapacilio No worries, I'll probably just have to revisit after the repo migration.  Thanks for the heads up!

PR #2239, opened 2022-04-13
Update API docs and Publishing Private Providers to make process clearer

laurapacilio:
Adding a private provider is a more complex process than most users would expect. In order to make the provider usable, users need to:
- Release their provider with certain release artifacts
- Create the provider
- Create a provider version
- Create a provider platform
- Upload all the required release files

While we have a documentation page on this topic, recent user feedback let us know that many users skip right to the API endpoints (which makes total sense). Once they get there, the endpoints don't have enough context to help them understand the other process steps they must complete to make a private provider usable, leading to confusion down the line. 

This PR updates the API documentation and our existing "Publishing a Private Provider" page to hopefully help make the process clearer and lead users who skip right to the API endpoints through the process a bit more seamlessly.


madaleines (approved):
LGTM! Appreciate the wording improvements and reformatting 🙂

laurenrother (approved):
It helps if I correctly read my options and choose "Approve" rather than just comment 🤦

PR #2246, opened 2022-04-20
Move PR template to right folder

laurapacilio:
I added it in the wrong place and now it doesn't show up. This PR should fix this.


BRKalow:
@laurapacilio This should go in https://github.com/hashicorp/terraform-website/tree/master/.github I think!

laurapacilio:
heh - thank you @BRKalow I _think_ I did it? Let me know if it needs to move again...if not, would you mind approving?

BRKalow (approved):
LGTM! 🤞

PR #2281, opened 2022-05-02
Add docs for no-operation (empty) applies.

sebasslash:
Docs for empty applies!

### DIFF
--- a/content/cloud-docs/api-docs/run.mdx
+++ b/content/cloud-docs/api-docs/run.mdx
@@ -109,6 +109,7 @@ Properties without a default value are required.
 
 | Key path                                           | Type                 | Default                                            | Description                                                                                                                                                                                                                                |
 | -------------------------------------------------- | -------------------- | -------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `data.attributes.allow-empty-apply`                | bool                 | false                                              | Whether this run should execute a [no operation apply](/cloud-docs/run/modes-and-options#no-operation-apply).                                                                                                                                             |
 | `data.attributes.auto-apply`                       | bool                 | The current Auto Apply setting in the workspace    | Whether to automatically apply changes when a Terraform plan is successful. Defaults to the [Auto Apply](/cloud-docs/workspaces/settings#auto-apply-and-manual-apply) workspace setting.                                                   |
 | `data.attributes.is-destroy`                       | bool                 | false                                              | Specifies whether this plan is a destroy plan that will destroy all provisioned resources. Mutually exclusive with `refresh-only`.                                                                                                         |
 | `data.attributes.message`                          | string               | "Queued manually via the Terraform Enterprise API" | Specifies the message to be associated with this run.                                                                                                                                                                                      |
@@ -182,6 +183,7 @@ curl \
       "created-at": "2021-05-24T07:38:04.171Z",
       "has-changes": false,
       "auto-apply": false,
+      "allow-empty-apply": false,
       "is-destroy": false,
       "message": "Custom message",
       "plan-only": false,
@@ -332,6 +334,7 @@ curl \
         "created-at": "2021-05-24T07:38:04.171Z",
         "has-changes": false,
         "auto-apply": false,
+        "allow-empty-apply": false,
         "is-destroy": false,
         "message": "Custom message",
         "plan-only": false,
@@ -420,6 +423,7 @@ curl \
       "created-at": "2021-05-24T07:38:04.171Z",
       "has-changes": false,
       "auto-apply": false,
+      "allow-empty-apply": false,
       "is-destroy": false,
       "message": "Custom message",
       "plan-only": false,
--- a/content/cloud-docs/run/ui.mdx
+++ b/content/cloud-docs/run/ui.mdx
@@ -53,6 +53,10 @@ Users can also leave comments if there's something unusual involved in a run.
 
 Note that once the plan stage is completed, until you apply or discard a plan, Terraform Cloud can't start another run in that workspace.
 
+### Empty apply
+
+By default, Terraform Cloud will automatically mark a run as completed if the plan contains no changes. However, plans that contain no changes can be applied if you create a run in [Allow Empty Apply Mode](/cloud-docs/run/modes-and-options#allow-empty-apply). This is particularly useful for upgrading between Terraform versions in your workspace, [read more on upgrading workspace state.](/cloud-docs/workspaces/state#upgrading-state).  
+
 ### Auto apply
 
 If you would rather automatically apply plans that don't have errors, you can [enable auto apply](/cloud-docs/workspaces/settings#auto-apply-and-manual-apply) on the workspace's "General Settings" page. Some plans can't be auto-applied, like plans queued by [run triggers](/cloud-docs/workspaces/settings/run-triggers) or by users without permission to apply runs. ([More about permissions.](/cloud-docs/users-teams-organizations/permissions))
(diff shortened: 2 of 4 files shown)


content/cloud-docs/run/ui.mdx
```diff
@@ -53,6 +53,10 @@ Users can also leave comments if there's something unusual involved in a run.
 
 Note that once the plan stage is completed, until you apply or discard a plan, Terraform Cloud can't start another run in that workspace.
 
+### Empty apply
+
+Plans that contain no changes can be applied if you create a run in [No-Operation Apply Mode](/cloud-docs/run/modes-and-options#no-operation-apply). This particularly useful for upgrading between Terraform versions in your workspace, [read more on upgrading workspace state.](/cloud-docs/workspaces/state#upgrading-state).
```
Uk1288, line 58: Should this be, `...This is particularly useful...`

content/cloud-docs/api-docs/run.mdx
```diff
@@ -109,6 +109,7 @@ Properties without a default value are required.
 
 | Key path                                           | Type                 | Default                                            | Description                                                                                                                                                                                                                                |
 | -------------------------------------------------- | -------------------- | -------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `data.attributes.allow-empty-apply`                | bool                 | false                                              | Whether this run should execute a no operation apply. ## TODO ADD NO OPERATION APPLY LINK HERE                                                                                                                                             |
```
laurapacilio, line 112: don't forget :)

content/cloud-docs/workspaces/state.mdx
```diff
@@ -41,6 +41,12 @@ Terraform's built-in [`terraform_remote_state` data source](/language/state/remo
 
 Terraform Cloud automatically manages API credentials for `terraform_remote_state` access during [runs managed by Terraform Cloud](/cloud-docs/run#remote-operations). This means you do not usually need to include an API token in a `terraform_remote_state` data source's configuration.
 
+## Upgrading State
+
+A workspace's state version can be upgraded to a new Terraform version using no-operation (empty) applies. Empty applies are a [type of run mode](/cloud-docs/run/modes-and-options) that allows Terraform Cloud to proceed with an apply when a plan produces no changes.
```
nfagerlund, line 46: Hmm, this is great but might benefit from a bit of expansion. 

I think the key points to try and fit into here somehow are: 

- Sometimes the state format changes between Terraform versions.
- Terraform will automatically perform any needed state upgrades when you apply a run. 
- If you choose a version of Terraform that isn't compatible with your existing state, it will fail the run and tell you which version you need to apply with first. You can solve the problem by switching to that version and running a full apply.
- You can use a no-op (empty) apply to perform these state upgrades without needing to make any config changes.

nfagerlund (changes requested):
I think this is heading in a good direction, but I have some requests around the wording! 

As best as my memory serves, I _think_ you managed to find all the most important places that need to be updated for this! But I've been out of the game for a while and my Total Docs Mind-Map has crumbled a bit, so I defer to @laurapacilio to think of any other places that might want a note. 😅

laurapacilio:
@sebasslash - I'm going to wait until you respond to @nfagerlund's comments and then I'd love to take a pass :-)

content/cloud-docs/api-docs/run.mdx
```diff
@@ -109,6 +109,7 @@ Properties without a default value are required.
 
 | Key path                                           | Type                 | Default                                            | Description                                                                                                                                                                                                                                |
 | -------------------------------------------------- | -------------------- | -------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `data.attributes.allow-empty-apply`                | bool                 | false                                              | Whether this run should execute a [no operation apply](/cloud-docs/run/modes-and-options#no-operation-apply).                                                                                                                                             |
```
laurapacilio, line 112: I like that we're using "no-operation" to help people understand what this means, but the actual type is called "empty apply" so I don't want to use "no-operation" as a link. What do you think of this?

```suggestion
| `data.attributes.allow-empty-apply`                | bool                 | false                                              | Whether this run should execute a no-operation apply from a plan with no changes. Refer to [empty apply](/cloud-docs/run/modes-and-options#no-operation-apply) for details.                                                                                                                                            |
```

content/cloud-docs/workspaces/state.mdx
```diff
@@ -41,6 +41,12 @@ Terraform's built-in [`terraform_remote_state` data source](/language/state/remo
 
 Terraform Cloud automatically manages API credentials for `terraform_remote_state` access during [runs managed by Terraform Cloud](/cloud-docs/run#remote-operations). This means you do not usually need to include an API token in a `terraform_remote_state` data source's configuration.
 
+## Upgrading State
+
+A workspace's state version can be upgraded to a new Terraform version without making any configuration changes using no-operation (empty) applies. Empty applies are a [type of run mode](/cloud-docs/run/modes-and-options) that instructs Terraform Cloud to perform an apply when a plan produces no changes.   
+
+-> **Note:** If the desired Terraform version is incompatible with a workspace's existing state version, the run will fail prompting to run an apply with a compatible version first. Please refer to the [Terraform upgrade guides](/language/upgrade-guides) for more information on upgrading between versions.
```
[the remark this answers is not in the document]
laurapacilio: TBD pending some discussions we're having :)

content/cloud-docs/run/ui.mdx
```diff
@@ -53,6 +53,10 @@ Users can also leave comments if there's something unusual involved in a run.
 
 Note that once the plan stage is completed, until you apply or discard a plan, Terraform Cloud can't start another run in that workspace.
 
+### Empty apply
+
+By default, Terraform Cloud will automatically mark a run as completed if the plan contains no changes. However, plans that contain no changes can be applied if you create a run in [Allow Empty Apply Mode](/cloud-docs/run/modes-and-options#allow-empty-apply). This is particularly useful for upgrading between Terraform versions in your workspace, [read more on upgrading workspace state.](/cloud-docs/workspaces/state#upgrading-state).
```
laurapacilio, line 58:
```suggestion
By default, Terraform Cloud automatically marks a run as completed if the plan contains no changes. Creating a run in [Allow Empty Apply Mode](/cloud-docs/run/modes-and-options#allow-empty-apply) allows Terraform to perform a no-operation apply from a plan containing no infrastructure changes. During an empty apply, Terraform upgrades the state file format to match the Terraform version, if required. This is particularly useful for upgrading between Terraform versions in your workspace. Refer to [upgrading state](/cloud-docs/workspaces/state#upgrading-state) for more details..  
```

laurapacilio (approved):
Approving so I don't block this, as I'm out tomorrow, but I have the first round of suggested edits. Some things that I think we may want to add still (pending ongoing discussions)
- Should we explain to users that they can't just upgrade more than one version at a time? My understanding is that for example, if you want to upgrade from v0.11 to v0.13, you need to upgrade to v0.12 first, run the empty apply and then upgrade to v0.13 and then run empty apply again to upgrade state.
- When do you actually get an error? I'm confused a tiny bit by this concept of "when you try to run with a versio that's not compatible with your state, you get an error" but then we say that you need to do a full apply to upgrade.

brandonc:
@sebasslash I'm sorry to do this, but there was so much overlap between plan-only and allow-empty-apply that I decided to snapshot this PR and incorporate it into [#2313](https://github.com/hashicorp/terraform-website/pull/2313) -- including Laura's latest suggestion but removing the `workspaces/state.mdx` changes and instead going with more of an illustrated upgrade guide in `run/ui.mdx`

PR #2270, opened 2022-04-28
Update creating workspaces and managing variables docs

ker-an:
### What

Team Terraform Webex has added a new Variable Editor step to the Create Workspace Wizard for VCS workflows. The new step extracts variables from the Terraform configuration and prompts users to add missing variable values. If a user skips the Variable Editor step, the variables will still be added to the workspace Variables page and a user can fill them out there.

The pages updated:

- [Creating Workspaces](https://www.terraform.io/cloud-docs/workspaces/creating)
- [Managing Variables](https://www.terraform.io/cloud-docs/workspaces/variables/managing-variables)

### Why
This change makes it easier for a user to add required variables to a workspace. Previously, a user had to add each variable manually - now they're only responsible for entering the correct value for undefined variables in a Terraform configuration.

----------

#### Pull Request
- [x] One or more labels describe the type of change (e.g. clarification) and associated product (e.g. tfc).
- [x] Description links to related pull requests or issues, if any.

#### Content
- [x] Redirects have been added for moved, renamed, or deleted pages. (N/A)
- [x] Sidebar navigation files have been updated for added, deleted, reordered, or renamed pages. (N/A)
- [x] Links to related content where appropriate (e.g., API endpoints, permissions, etc.).
- [x] Pages with related content are updated and link to this content when appropriate.
- [x] New pages have metadata (page name and description) at the top. (N/A)
- [x] New images are 2048 px wide. They have HashiCorp standard annotation color (#F92672) and format (rectangle with rounded corners), blurred sensitive details (e.g. credentials, usernames, user icons), and descriptive alt text in the markdown for accessibility. (N/A)
- [x] New code blocks have the correct syntax and line breaks to eliminate horizontal scroll bars. (N/A)
- [x] UI elements (button names, page names, etc.) are bolded.
- [x] The Vercel website preview successfully deployed.

#### Reviews
- [x] I or someone else reviewed the content for technical accuracy.
- [x] I or someone else reviewed the content for typos, punctuation, spelling, and grammar.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201817828939976
  - https://app.asana.com/0/0/1202194029061777

PR #2293, opened 2022-05-04
Revert "Update creating workspaces and managing variables docs"

ker-an:
We are reverting these changes for now while we investigate and fix an issue with the variable editor. The variable editor is back behind a feature flag and unavailable to users.

Reverts hashicorp/terraform-website#2270


laurapacilio (approved):
To save for when the feature is finished :-)

PR #2383, opened 2022-08-01
feat(core): versioned-docs

thiskevinwang:
# Description

This enables versioned-docs for Terraform "core" pages


pages/docs/[[...page]].tsx
```diff
@@ -62,6 +62,7 @@ const { getStaticPaths, getStaticProps } = getStaticGenerationFunctions(
         product: SOURCE_REPO,
         remarkPlugins,
         rehypePlugins,
+        enabledVersionedDocs: true,
```
Comment (BRKalow) [inline pages/docs/[[...page]].tsx:65]: 🤔 does this need to be here?

Comment (thiskevinwang) [reply pages/docs/[[...page]].tsx:65]: Oopsies — nope

PR #2384, opened 2022-08-02
chore: redirects from /configuration/*

thiskevinwang:
# Description

This adds redirects for 3 pages that have nearly no content and serve only as landing pages that direct to other pages

Traffic for past 30 days (from https://usefathom.com)

```
                           Entries     Visitors    Views
/configuration             2           11          15
/configuration/expressions 1.2K        3.7K        5.3K
/configuration/resources   1.4K        2.9K        4.2K
/configuration/modules     687         1.6K        2K
```

### DIFF
--- a/redirects.next.js
+++ b/redirects.next.js
@@ -218,8 +218,8 @@ module.exports = (async () => {
     '/cdktf/concepts/providers-and-resources' : '/cdktf/concepts/providers',
     // The /configuration pages have been deleted.
     '/configuration/expressions': '/language/expressions',
-    '/configuration/modules': '/language/modules/syntax',
-    '/configuration/resources': '/language/resources/syntax',
+    '/configuration/modules': '/language/modules',
+    '/configuration/resources': '/language/resources',
   }
   const miscRedirects = Object.entries(miscRedirectsMap).map(
     ([source, destination]) => {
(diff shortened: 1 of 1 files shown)


redirects.next.js
```diff
@@ -215,7 +215,11 @@ module.exports = (async () => {
     '/cloud-docs/api-docs/run-tasks': '/cloud-docs/api-docs/run-tasks/run-tasks',
     '/cloud-docs/api-docs/run-tasks-integration': '/cloud-docs/api-docs/run-tasks/run-tasks-integration',
     '/language/provider-checksum-verification': '/language/files/dependency-lock#checksum-verification', // Used by the Terraform CLI to short-link to documentation.
-    '/cdktf/concepts/providers-and-resources' : '/cdktf/concepts/providers'
+    '/cdktf/concepts/providers-and-resources' : '/cdktf/concepts/providers',
+    // The /configuration pages have been deleted.
```
thiskevinwang, line 219: _will be deleted_

thiskevinwang: I didn't add a redirect for `/configuration` since it gets close to 0 traffic

BRKalow (approved):

redirects.next.js
laurapacilio: tbh this makes complete sense to me. Idk where we would even send folks anyway for that one

laurapacilio (approved):
Per Judith's direction these are all to the high level pages - makes sense to me!

PR #2386, opened 2022-08-02
chore: `/guides` redirects

thiskevinwang:
# Description

The two `/guides` pages will be removed. This adds redirects for those


redirects.next.js
```diff
@@ -220,6 +220,9 @@ module.exports = (async () => {
     '/configuration/expressions': '/language/expressions',
     '/configuration/modules': '/language/modules',
     '/configuration/resources': '/language/resources',
+    // The /guides pages have been deleted.
+    '/guides': '/intro',
+    '/guides/terraform-provider-development-program': '/docs/partnerships',
```
thiskevinwang, line 225: This appears to be a stray page.

We have a redirect for the `.html` page, to `/docs/partnerships`, here:
https://github.com/hashicorp/terraform-website/pull/2386/files#diff-47e6b345a70afa7d557486ce2bfac0d2b35e454b53705d2159b57a84e2762ba7R159

BRKalow (approved):

PR #2441, opened 2022-11-10
fix: redirect downloads to dev-dot

zchsh:
This PR adds `downloads` as a route that we redirect to Dev Dot via middleware.


ashleemboyer (approved):
Testing by going to `/downloads` on the preview URL and it works! ✅

PR #2493, opened 2023-05-24
Remove editions from nav

mwickett:
### What
Removes the "Editions" menu item. CLI was pointing to the homepage, and Cloud & Enterprise both now go to https://www.hashicorp.com/products/terraform so the whole entry is redundant. 

### Why
Simplifies the nav for users

### DIFF
--- a/data/primary-navigation.js
+++ b/data/primary-navigation.js
@@ -4,21 +4,6 @@ export default [
     url: '/',
   },
   {
-    text: 'Editions',
-    type: 'anchor',
-    submenu: [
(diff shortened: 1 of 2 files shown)


data/primary-navigation.js
mwickett, line 1: I _think_ that this file is not used any longer since the nav data was hardcoded into `Subnav` props (see other changed file), but I left this in just for the sake of completeness.

nandereck (approved):
LGTM! 🔪

